PR #1, opened 2023-06-19
Use Sync in isolate

StaxoLotl:
- Use Sync in isolate
- format code with dart format


mocodesmo:
Hi thanks for the PR.

The issue we are experiencing is - **when the wallet is syncing and the user clicks the receive button then loading an address needs to wait for sync to complete** - the .getAddress(addressIndex: const AddressIndex.new()) waits for the wallet to complete syncing - for now, i have added a loadingindicator - getAddress() is an offline operation and shouldn't wait for .sync() to complete before completing

Issue #20, opened 2023-07-01
Bug when creating unsigned PSBTs


BullishNode:
When creating an unsigned PSBT from a watch-only wallet, the tx builder seems to be trying to add the recipient address as one of the inputs. The end result is that a Coldcard is not able to sign the unsigned PSBT with the error message "Signing failed late. Path (m/84'/0'/0'/0/3) let to the wrong pubkey for input#0). I used a brand new coldcard wallet for this so the path should probably have been (m/84'/0'/0'/0/0). Not exactly sure what is going on, but definitely a bug here in the way the wallet is crafting the unsigned transaction.

i5hi:
This can be closed.

i5hi:
Broadcast in the psbt flow is still buggy, made a new issue for this which should be resolved by eod.

Following that please review the entire psbt flow.

i5hi:
Fixed in 0.1.5

PR #26, opened 2023-07-06
Fixed coldcard descriptor bug

i5hi:
The source of this issue was in the `_pkg/wallet/create/loadBdkWallet` function, in which cold card wallets were following the same case as xpub. 

- Added a new case for cold card wallets which do not need to further derive keys.

There were few other issues I noticed along the way, which we can consider fixing as part of this PR: 

- [ ] In some derivation paths the hardened paths are being removed because of errors using the apostrophe `'` character. It will be better to use the letter `h` character consistently across the codebase which is less prone to stringification errors and it also works reliably with bdk.
- [ ] Broadcast page shows scan xpub; instead it must show scan signed psbt or tx hex.
- [ ] Broadcast does not work for a signed psbt

### DIFF
--- a/lib/_model/wallet.dart
+++ b/lib/_model/wallet.dart
@@ -92,7 +92,8 @@ class Wallet with _$Wallet {
       if (err3 != null) errs.add(err3.message);
       wallets.add(wallet44!);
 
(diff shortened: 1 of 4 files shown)


i5hi:
This PR only resolves https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/20

PR #83, opened 2023-09-17
Two digits issue on recover wallet and receive tx page disabled bump fee

yshwanth:
Receive tx page cant bump fee now, fixed a two digit number size on recover wallet


i5hi:
Thanks for the PR @yshwanth !

Taking a look!

i5hi:
LGTM!

PR #96, opened 2023-09-23
Shrink vertical padding on import words

yshwanth:


i5hi:
Thanks @yshwanth 

Checking this out now!

PR #98, opened 2023-09-25
Transaction sent animation

yshwanth:
Need reviews,will update UI accordingly


yshwanth:
Updated the UI and fixed small bugs

i5hi:
Thanks @yshwanth !

PR #114, opened 2023-10-09
Disabled coin selection if selecting sweep balance

yshwanth:
Fixed issues

- coin selection should not be allowed if send full wallet balance is selected.

- In recieve - create invoice, if my chosen unit is btc i should be able to set the amount in BTC, right now it only allows setting amount in sats.


i5hi:
Thanks @yshwanth

PR #120, opened 2023-10-19
Price API update

BullishNode:
changing price API to the production endpoint

http://api.bullbitcoin.com/price


i5hi:
yoo! cheers!

Issue #136, opened 2023-10-30
[Recover backup] Jump to next word on Keyboard return


saiy2k:
App version: v0.1.91

When entering the seed words,
If I select the world from suggestion popup, the focus goes to next word automatically, which is good.

Same can happen when I press return key in the keyboard. Right now it closes the keyboard.

![Recover backup - 12 words 001](https://github.com/SatoshiPortal/bullbitcoin-mobile/assets/701985/015db3b5-5b5c-4ef8-85ab-2ed6b0fe2a5b)

i5hi:
@mocodesmo The enter button now cancels the word, rather than selecting the word in the list. 

Enter should select the highlighted word.

If three words are left and they press enter, it will select the first word.

If one work is left, it will select the only word left. 


It is intuitive to press the enter button after typing the whole word and assume your entry will be selected. 

If you no words in the list exist when you press enter, then it can remove the entry.

i5hi:
@mocodesmo currently it selects the word but the drop down remains.

Drop down should dissapear and focus should be on the next input field.

mocodesmo:
fixed on main yesterday @i5hi

PR #176, opened 2023-11-03
Error handler dialog & Fix for #136

saiy2k:
1. Added a generic ErrorAlert component and converted all error message code to make use of this.
 - The red bg shade is coming from the theme.
 - RIght now, we got error in the states as strings. Ideally error variable in the states should of some Error class, with fields like type, message, code, stacktrace, etc.,
 - I thought of writing some logic to convert the technical error messages to user friendly error messages. But don't want to do that with the current structure.

2. Fix for [ #136](https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/136)

### DIFF
--- a/lib/_ui/components/alert.dart
+++ b/lib/_ui/components/alert.dart
@@ -1,0 +1,67 @@
+import 'package:bb_mobile/styles.dart';
+import 'package:flutter/material.dart';
+
+class AlertData {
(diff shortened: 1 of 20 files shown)


i5hi:
Closing this, since your last PR has been merged.

Thanks @saiy2k !

PR #196, opened 2023-11-14
Alert dialog: Integrated in Setting :: Broadcast page

saiy2k:
Old error message is still displayed for testing. Can remove it later.
Have to look for the red bg color, that's coming from theme.


mocodesmo:
@saiy2k pls clean up utils.dart showAlert method 
- no 'dynamic' types allowed
- looks like a js arrow function

PR #238, opened 2024-02-20
No screenshot

i5hi:


mocodesmo:
is this tested ? @saiy2k  @i5hi

saiy2k:
Yes. Tested in my mobile.

Added `routes` of sensitive screens, which are not be screenshotted/recordede in main.dartL126.
If any more screen is missing, route of that screen can be added in that array.

i5hi:
Just tested. LGTM.

PR #269, opened 2024-04-26
Dynamic fee calculation 1/2

saiy2k:
Few Notes:
* newBoltzApi() iniaitlization url needs to be done proper.
* `claimFeesEstimate` and `lockupFees` values needs to be verified in `boltz-dart` if they are right.

### DIFF
--- a/lib/receive/receive_page.dart
+++ b/lib/receive/receive_page.dart
@@ -59,7 +59,7 @@ class _ReceivePageState extends State<ReceivePage> {
       walletSensitiveRepository: locator<WalletSensitiveStorageRepository>(),
       swapBoltz: locator<SwapBoltz>(),
       walletTx: locator<WalletTx>(),
-    );
+    )..fetchFees(context.read<NetworkCubit>().state.testnet);
 
     _currencyCubit = CurrencyCubit(
       hiveStorage: locator<HiveStorage>(),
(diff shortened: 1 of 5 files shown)


lib/swap/bloc/swap_cubit.dart
```diff
@@ -185,7 +197,7 @@ class SwapCubit extends Cubit<SwapState> {
     }
 
     final updatedSwap = swap!.copyWith(
-      boltzFees: (fees.btcSubmarine.boltzFeesRate * amount) as int,
+      boltzFees: (fees.btcSubmarine.boltzFeesRate * amount / 100) as int,
```
i5hi: we are using submarine here. should be reverse.

lib/receive/receive_page.dart
```diff
@@ -59,7 +59,7 @@ class _ReceivePageState extends State<ReceivePage> {
       walletSensitiveRepository: locator<WalletSensitiveStorageRepository>(),
       swapBoltz: locator<SwapBoltz>(),
       walletTx: locator<WalletTx>(),
-    );
+    )..fetchFees(context.read<NetworkCubit>().state.testnet);
```
i5hi, line 62: why are we using testnet here? what if its mainnet?

saiy2k: So `context.read<NetworkCubit>().state.testnet` is a bool. It returns true/false, based on current network selection in app.

i5hi: ah okay. thanks!

PR #270, opened 2024-04-26
App migration

saiy2k:
Logic summarized @ https://docs.google.com/document/d/16iH7PlA6IqnNUNd0NE0rYH4XOVkCLmmhkTrDwjUtWtg/edit

### DIFF
--- a/lib/_pkg/storage/migration.dart
+++ b/lib/_pkg/storage/migration.dart
@@ -78,6 +78,7 @@ Future<void> doMigration01to02(SecureStorage secureStorage, HiveStorage hiveStor
         if (mainWalletIndex == 0) {
           walletObj['type'] = 'secure';
           walletObj['name'] = 'Secure Bitcoin Wallet / ' + (walletObj['name'] as String);
+          walletObj['mainWallet'] = true;
           mainWalletIndex++;
 
           final mnemonicFingerprint = walletObj['mnemonicFingerprint'] as String;
@@ -94,6 +95,7 @@ Future<void> doMigration01to02(SecureStorage secureStorage, HiveStorage hiveStor
         if (testWalletIndex == 0) {
           walletObj['type'] = 'secure';
           walletObj['name'] = 'Secure Bitcoin Wallet / ' + (walletObj['name'] as String);
+          walletObj['mainWallet'] = true;
           testWalletIndex++;
 
           final mnemonicFingerprint = walletObj['mnemonicFingerprint'] as String;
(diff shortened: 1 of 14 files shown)


lib/_pkg/storage/migration.dart
```diff
@@ -0,0 +1,185 @@
...
+      network: BBNetwork.Testnet,
+      walletCreate: walletCreate,
+    );
+    final liquidWallet = lw?.copyWith(name: lw.creationName(), mainWallet: true);
+    print(liquidWallet?.id);
+    await walletsStorageRepository.newWallet(liquidWallet!);
+  }
+
+  // Finally update version number to next version
+  await secureStorage.saveValue(key: StorageKeys.version, value: '0.2');
```
i5hi, line 176: this should be BBVersion, not hard coded as 0.2

pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: bb_mobile
 description: Bull Bitcoin Mobile Wallet
 publish_to: 'none' 
-version: 0.2.0-1
+version: 0.2.0+18
```
i5hi, line 4: we are currently on +14

lib/_pkg/storage/migration.dart
saiy2k: Actually this line is inside `doMigration01to02()`, once this function is done, the app is now at version 0.2.
So between 0.2 to 0.3, 0.3 to 0.4 migrations, if any error happens, another attempt at migration can skip this `doMigration01to02()` function since it's already done.

But in `storage.dart`, we have these lines:
```
    await doMigration(version!, bbVersion, secureStorage, hiveStorage);
    await secureStorage.saveValue(key: StorageKeys.version, value: bbVersion);
```

Once all migrations are done, finally current appversion is saved in storage, with that line.

saiy2k: Adding pseudo code for better clarity:

```
//storage.dart
if (currentVersion != bbVersion) {
  doMigration()
  save bbVersion in storage
}

//migration.dart
doMigration() {
  if (toVersion == 0.2 && fromVersion == 0.1) {
  
    migrate_0.1to0.2();
    
  } else if (toVersion == 0.3)
  
    if (fromVersion == 0.1) {
    
      migrate_0.1to0.2();
      migrate_0.2to0.3();
      
    } else if (fromVersion == 0.2) {
    
      migrate_0.2to0.3();
      
    } 
    
  }
}

migrate_0.1to0.2() {
  // 0.1 to 0.2 migration logic
  // save current version as 0.2 in storage
}

migrate_0.2to0.3() {
  // 0.2 to 0.3 migration logic
  // save current version as 0.3 in storage
}
```

Issue #322, opened 2024-09-01
Error message when adding custom node is unclear


Bayernatoor:
**ISSUE:** 

When trying to add my own electrum server the error returned in not visible to the user. After pressing "Save", a menu appears with what seems to be an error message but quickly disappears. The error string at the top overflows and is not visible. 

Can you not connect to a Tor node?

**Screenshots:** 

<img src="https://github.com/user-attachments/assets/74a39f5a-afc8-466e-ae2b-54e5d4751ab6" alt="Image" width="300" height="500">

After attempting to "sync transactions" I got the correct error message: 

<img src="https://github.com/user-attachments/assets/ddf2d73f-8435-4a54-b6bb-def0204a6ad6" alt="Image" width="300" height="500">

**Additional Info:**

My electrum server connects fine to Green Wallet using:

tcp://hiddenaddress.onion:50001

When revisiting the connection page, Bull wallet displays a green icon next to my tor address, when tapping save again it turns to red and I get the error message.

<img src="https://github.com/user-attachments/assets/77c23b25-e1a1-423e-b8e1-9ea1d3f880a2" alt="Image" width="300" height="500">

i5hi:
Thanks for the report.

We will reproduce and address the issues stated.

Tor isnt supported just yet.

raehat:
@i5hi Is this issue up for grabs? If yes, I can try addressing this issue, I'm new to this repository and wanted to contribute if possible, I've good experience in both bitcoin and app development.

i5hi:
Hey @raehat, Yes it is! Feel free to DM me on telegram if you need help. TBH the way we handle electrum needs some rework - especially errors.

raehat:
@i5hi can you please share your telegram account?

i5hi:
@i5hi1

raehat:
@Bayernatoor can you please share address of electrum server you're using

raehat:
@i5hi I noticed that some code was pushed after this issue was created, which somewhat addresses the problem by adding validation rules to check the URL format. As a result, the user will no longer be able to proceed with an onion URL due to these rules.

However, one issue still remains: the error message is not descriptive enough. Ideally, we should inform the user that Tor nodes are not supported. Additionally, there could be other cases like this, where a more detailed error message would be helpful.

If you'd like, I can raise a PR that provides a clear error indicating that Tor nodes are not supported in this case.

i5hi:
hey, yes that would be great!

[...]

PR #330, opened 2024-10-20
feat(network): add error handling for Tor node connection attempt

raehat:
Fix for #322 
Returns a descriptive error message when user tries to add a custom tor node


i5hi:
Hey!

Thanks for the PR @raehat

Issue #340, opened 2024-12-12
Wallet will lists balance, outgoing transactions but not received transactions


DanGould:
I'm finding this issue in `syncWalletOld`. Only seeing "No Transactions yet" despite a balance being displayed. In `BDKTransactions.getTransactions()`, `storedTx` is `null`, so the following line `updatedTx = Transaction(...` returns without being handled and sync never completes.

![image](https://github.com/user-attachments/assets/835be9fb-1be6-4a7d-9ad2-72fbddd68abe)

HLeach-Tech:
> I'm finding this issue in `syncWalletOld`. Only seeing "No Transactions yet" despite a balance being displayed. In `BDKTransactions.getTransactions()`, `storedTx` is `null`, so the following line `updatedTx = Transaction(...` returns without being handled and sync never completes.
> 
[...]

PR #342, opened 2024-12-13
fix: BigInt to int conversions

kumulynja:
This PR fixes some conversion issues between `BigInt` and `int` types in bdk's `getTransactions` function causing it to throw an error and not show the transaction history.


DanGould (approved):
tACK b39de1d57655f6511404f3d6e8a0be2b66f884af

This fixes my history problem and closes #340 

Would be good to be able to test this to prevent it from regressing, but the fix is better then the broken version so I'm comfortable merging as-is.

image

PR #347, opened 2024-12-16
chore: track pubspec.lock

ethicnology:
[Dart doc on pubspec.lock](https://dart.dev/tools/pub/private-files#pubspeclock)

>_For application packages, we recommend that you commit the pubspec.lock file. Versioning the pubspec.lock file ensures changes to transitive dependencies are explicit. Each time the dependencies change due to dart pub upgrade or a change in pubspec.yaml the difference will be apparent in the lock file._

In depth, if a dev change a dependency version, `flutter clean` + `flutter pub get` will not be enough. Since `flutter clean` doesn't delete the `pubspec.lock` you will have to delete it manually if you expect the `flutter pub get` command to fetch the correct version


DanGould (commented):
ACK [fe5f932](https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/347/commits/fe5f93235ab79221072a4b7dda485516c729e258)

it builds on my machine

PR #348, opened 2024-12-16
fix: custom manual fee not used in Send

kumulynja:
Two different NetworkFeesCubit instances where used between the SendPage and the SendCubit: the SendPage components used a local NetworkFeesCubit to store the fee state, but the global default NetworkFeesCubit was injected in the SendCubit and was used to get the fees on confirmation from, not having the locally set state.


kumulynja:
I see that I accidentally closed this when forking the repo from Dan's github and deleting the original fork.
Will restore this.

kumulynja:
I see that I accidentally closed this when forking the repo from Dan's github and deleting the original fork.
Will restore this.

PR #349, opened 2024-12-16
Fix untrack-generated-files

DanGould:
lwkWallet.sync no longer takes a validateDomain parameter

I was unable to build 0acf3e4669ce8a11b6d5a712c397a7bd611951d5 without this change

Unclear to me whether or not this was a lock file or freeze change that overlooked a breaking change in that library, but this fixes it.

### DIFF
--- a/macos/Flutter/ephemeral/Flutter-Generated.xcconfig
+++ b/macos/Flutter/ephemeral/Flutter-Generated.xcconfig
@@ -1,12 +1,11 @@
 // This is a generated file; do not edit or check into version control.
-FLUTTER_ROOT=/Users/ishi/Software/flutter
-FLUTTER_APPLICATION_PATH=/Users/ishi/Code/wallet/bullbitcoin-mobile
+FLUTTER_ROOT=/opt/homebrew/Caskroom/flutter/3.24.4/flutter
+FLUTTER_APPLICATION_PATH=/Users/dan/f/dev/bullbitcoin-mobile
 COCOAPODS_PARALLEL_CODE_SIGN=true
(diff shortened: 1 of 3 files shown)


macos/Flutter/ephemeral/Flutter-Generated.xcconfig
```diff
@@ -1,12 +1,11 @@
 // This is a generated file; do not edit or check into version control.
-FLUTTER_ROOT=/Users/ishi/Software/flutter
-FLUTTER_APPLICATION_PATH=/Users/ishi/Code/wallet/bullbitcoin-mobile
+FLUTTER_ROOT=/opt/homebrew/Caskroom/flutter/3.24.4/flutter
+FLUTTER_APPLICATION_PATH=/Users/dan/f/dev/bullbitcoin-mobile
```
DanGould, line 3: /ephemeral

and 

.idea

files should almost definitely be ignored in the shared repo by using `.git/info/exclude` (or `skip-worktree`)

ethicnology: I will do a PR to exclude these files in few minutes

ethicnology: https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/347

I got the same issue earlier. Ishi updated an existing dependency, since the pubspec.lock is not tracked, your local pubspec.lock keep the same old version of the package each time you do a flutter pub get

DanGould:
This is a nonissue due to ignored pubspec.lock

irrelevant and fixed by #347

Issue #334, opened 2024-11-18
[Send] bip49 addresses in QR are being converted into lowercase when scanned 


i5hi:
Example: 3Fz8Lj3pjGprg2nnZByXrGMhUGd1yPpn6E

kumulynja:
@i5hi https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/355 fixes this.

i5hi:
Cheers!

kumulynja:
@i5hi The fix didn't got merged to main yet, I accidentally closed the PR, I opened a new one here now: https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/362
Sorry for the confusion.

PR #355, opened 2024-12-16
fix: remove address toLowerCase

kumulynja:
For bip49 it is important to retain letters in lower and uppercase. Removing the `toLowerCase()` in the correct place makes sure it is and bip49 addresses are correctly scanned and passed to the Send input.


kumulynja:
This fixes https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/334.

kumulynja:
Accidentally closed this, will reopen.

PR #362, opened 2024-12-19
fix: remove address toLowerCase

kumulynja:
For bip49 it is important to retain letters in lower and uppercase. Removing the toLowerCase() in the correct place makes sure it is and bip49 addresses are correctly scanned and passed to the Send input.

This fixes https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/334.


kumulynja:
@i5hi my previous PR for this got closed without being merged because I accidentally deleted my fork.
Here is a new PR for it.

i5hi:
Cheers!

Issue #331, opened 2024-10-28
[Custom Fee Selection] Custom value not being caputured


i5hi:
When a custom fee value is typed in fee selection, it is not actually used in building the transaction.

kumulynja:
@i5hi This should be fixed by https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/348.

i5hi:
Cheers!

kumulynja:
@i5hi I accidentally closed my previous PR and it wasn't merged to main yet, I made a new one here: https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/363
Sorry for the confusion.

PR #363, opened 2024-12-19
fix: custom manual fee not used in Send

kumulynja:
Two different NetworkFeesCubit instances where used between the SendPage and the SendCubit: the SendPage components used a local NetworkFeesCubit to store the fee state, but the global default NetworkFeesCubit was injected in the SendCubit and was used to get the fees on confirmation from, not having the locally set state.

This fixes https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/331.


kumulynja:
@i5hi I accidentally closed the previous PR without it being merged and am unable to reopen it, so this isn't merged to main yet. Can you do that with this PR, please?
Sorry for the confusion.

i5hi:
Nice catch! Cheers @kumulynja !

PR #365, opened 2024-12-19
Send Payjoin

DanGould:
This is a basic sender implementation that may be merged so we can build off of it. It does not include any isolate behavior or do accounting properly. I would like help getting the accounting right by taking data from a signed payjoin and pushing it into the Transaction model @ethicnology.

This separates the signing and "confirmed" stages and marshals successful payjoins to the final screen. I don't think a new payjoin handshake screen is merited until the isolates can allow that behavior to happen outside of the send cubit.

Please see the commit log for hacks taken to get this working. 

For a payjoin send:

- [x] when a payjoin address is pasted the UI should mention that the transaction is a payjoin on the build tx page (3rd commit)
- [x] the build process should be separated from the confirm process and we should have a screen to confirm the details (2nd commit)
- [ ] after confirm there should be a status page for the payjoin (similar to swaps) - there is an option to exit and go to the transaction detail page or go back home
- [ ] there should be a toast that notifies the user when a payjoin broadcasts
- [ ] uri & address copy / paste should be separate

I ask reviewers to confirm errors are handled appropriately, since this flutter paradigm is new to me.

### DIFF
--- a/lib/_pkg/payjoin/manager.dart
+++ b/lib/_pkg/payjoin/manager.dart
@@ -1,0 +1,122 @@
+import 'dart:async';
+import 'dart:math';
+
+import 'package:dio/dio.dart';
+import 'package:payjoin_flutter/common.dart';
+import 'package:payjoin_flutter/send.dart';
+import 'package:payjoin_flutter/uri.dart' as pj_uri;
+
+const List<String> _ohttpRelayUrls = [
+  'https://pj.bobspacebkk.com',
+  'https://ohttp.achow101.com',
+];
+
+class PayjoinManager {
+  Future<Sender> initSender(
+    String pjUriString,
+    int networkFeesSatPerVb,
+    String originalPsbt,
+  ) async {
+    try {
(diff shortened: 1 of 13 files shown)


ethicnology:
rebased on upstream/main

DanGould:
There is a bug in here where after multiple sends the sender will throw: flutter: BdkError.electrum(field0: Electrum server error: "sendrawtransaction RPC error -26: {\"code\":-26,\"message\":\"mandatory-script-verify-flag-failed (Witness program hash mismatch)\"}") which must be from the broadcast call. My hunch is that the sender is trying to spend a UTXO that's already been spent but the wallet history hasn't  updated, so the spend check during broadcast fails.

i5hi:
The remaining UI related items can be left out of this PR

i5hi:
ACK!
Cheers @DanGould !

i5hi:
I'll patch the linting issues after merge so we can move forward.

PR #367, opened 2024-12-21
Receive payjoin

DanGould:
Spawn a receiver in an cubit, then use the Isolate to give it its own memory.

Commits with ✅ have been smoke tested


i5hi:
ACK.


Cheers Dan!

PR #369, opened 2024-12-21
Trigger toast with PayjoinEvent on pj send broadcast

DanGould:
This is just the trigger logic from which we can add more details and perhaps an app link into the history. This same logic can be used to display errors or other status updates.


i5hi:
Cheers Dan!

PR #372, opened 2024-12-22
DRAFT push payjoin details to transaction model

DanGould:
This is a draft attempt (IDEK WHAT I'M DOING IN FLUTTER) to move data from some of these isolated processes into the main wallet history data structures.

I don't it make sense to push completed transactions to the success screen, because payjoin is async. A sender should have the originalPsbt to pass to the success screen before entering spawnSender, and it will await a response from the sender with a payjoinPsbt to tupdate the transaction with.

A receiver should have an *unsigned* payjoinPsbt that it can listen for in mempool.

the transaction from an isolate is going to have to be passed as an event. That should propagate into the Transaction model and history. Then the network fees of each contributor can be calculated by taking the difference between originalPsbt and payjoinPsbt

### DIFF
--- a/lib/_model/transaction.dart
+++ b/lib/_model/transaction.dart
@@ -46,6 +46,7 @@ class Transaction with _$Transaction {
     bdk.TransactionDetails? bdkTx,
     // Wallet? wallet,
     @Default(false) bool isSwap,
+    @Default(false) bool isPayjoin,
     SwapTx? swapTx,
     @Default(false) bool isLiquid,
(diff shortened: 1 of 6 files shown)


ethicnology:
@DanGould I'm closing this one since we have refactored the whole codebase it's unlikely any work is still compatible. The work remain available on `trigger-event` branch

DanGould:
Yes, the underlying payjoin-flutter lib has changed too. This issue still remains however.

PR #373, opened 2024-12-23
Async Payjoin send and receives

DanGould:
The whole point of Payoin V2 is to save the state of a payjoin and resume it whenever you have internet. Our first implementation doesn't do that since the logic is limited to the Send or Receive UI context where it's created and it disappears when the app goes out of memory.

This change persists that information in HiveStorage, pauses sessions on app background, and resumes sessions on app foregroudn, allowing async payjoin to happen.

There was a bug where Sender's didn't persist information sent in their first request, so once they were persisted and resumed they were out of sync with receivers. This has been fixed in an upcoming release of payjoin-flutter and so requires that update in order to function.

In order to test, start a receiver, background that app, then start a sender to that receiver, then background that app, then resume the receiver, then stop it, then resume the sender. The payjoin should flow good as ever as long as there's enough time to send the first payjoin message before stopping at each stage.

### DIFF
--- a/lib/home/listeners.dart
+++ b/lib/home/listeners.dart
@@ -1,5 +1,7 @@
 import 'package:bb_mobile/_model/wallet.dart';
 import 'package:bb_mobile/_pkg/logger.dart';
+import 'package:bb_mobile/_pkg/payjoin/listeners.dart';
+import 'package:bb_mobile/_pkg/payjoin/manager.dart';
 import 'package:bb_mobile/_pkg/wallet/address.dart';
 import 'package:bb_mobile/_pkg/wallet/balance.dart';
 import 'package:bb_mobile/_pkg/wallet/create.dart';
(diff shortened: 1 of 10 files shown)


i5hi:
ACK

lib/home/listeners.dart
```diff
@@ -73,6 +75,34 @@ class WalletBlocListeners extends StatelessWidget {
         context.select((HomeCubit cubit) => cubit.state.walletBlocs);
     if (wallets == null || wallets.isEmpty) return child;
     // .read<HomeCubit>().state.walletBlocs ?? [];
+
+    // print each wallet id
+    for (final wallet in wallets) {
+      print('wallet id: ${wallet.state.wallet!.id}');
+    }
```
DanGould, line 82: This should most probably be removed, as should the print on line 97

Issue #370, opened 2024-12-22
[Payjoin Receive] initReceiver Error while creating receive URI


i5hi:
initReceiver throws the following error when creating a Receive uri

```
_Exception (Exception: Error initializing payjoin Receiver: PayjoinException( Error while parsing the address: legacy address base58 string ))
```

i5hi:
Note: This does not prevent the address from being created and displayed.

i5hi:
`legacy address base58 string`

The address being used is bech32 segwit

i5hi:
Seems like its using the liquid address.

We should not use the `Instant Wallet` for payjoins.

kumulynja:
@i5hi as mentioned in the PR that is already merged, maybe adding an extra check on the wallet type is also needed still instead of only the payment network, what do you think? Shall I add it before closing this issue completely?

ethicnology:
Closed by #381

PR #374, opened 2024-12-23
fix: only init payjoin when payment network is bitcoin

kumulynja:
The first receive tab and network that seems to be loaded initially is Lightning. For this an amount has to be entered and no default invoice should be generated, so returning immediately in the updateWalletBloc function is added for Lightning.
And then this PR also adds a check to make sure to only init payjoin for the bitcoin network, this way we avoid the liquid address being used to init the payjoin.

### DIFF
--- a/lib/receive/bloc/receive_cubit.dart
+++ b/lib/receive/bloc/receive_cubit.dart
@@ -53,6 +53,7 @@ class ReceiveCubit extends Cubit<ReceiveState> {
 
     if (state.paymentNetwork == PaymentNetwork.lightning) {
       emit(state.copyWith(defaultAddress: null));
+      return;
     }
 
     if (!walletBloc.state.wallet!.mainWallet) {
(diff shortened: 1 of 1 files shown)


kumulynja:
This should fix https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/370

kumulynja:
@i5hi This seems to fix it, but I have a doubt though. I am checking on the payment network now to init payjoin or not.
Do you think it is maybe better to also base it on the wallet that is selected being the secure onchain wallet instead of just the network?

kumulynja:
> @i5hi This seems to fix it, but I have a doubt though. I am checking on the payment network now to init payjoin or not. Do you think it is maybe better to also base it on the wallet that is selected, making sure the secure onchain wallet is selected too, instead of just the network?

As clarified by @i5hi and @ethicnology out-of-band, we should add a check for `BaseWalletType.Bitcoin`.

ethicnology:
> > @i5hi This seems to fix it, but I have a doubt though. I am checking on the payment network now to init payjoin or not. Do you think it is maybe better to also base it on the wallet that is selected, making sure the secure onchain wallet is selected too, instead of just the network?
> 
> As clarified by @i5hi and @ethicnology out-of-band, we should add a check for `BaseWalletType.Bitcoin`.

#381

Issue #371, opened 2024-12-22
[Payjoin Receive] After creating a payjoin address for Bitcoin, the same address is used under the Liquid tab


i5hi:
<img width="345" alt="Screenshot 2024-12-22 at 1 23 23 PM" src="https://github.com/user-attachments/assets/0e005dab-f3ca-4f3c-ae78-dc8f5747a040" />
<img width="343" alt="Screenshot 2024-12-22 at 1 23 30 PM" src="https://github.com/user-attachments/assets/849113cf-0440-41bd-b27f-1b88021704d6" />

kumulynja:
This should be fixed by: https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/375

PR #375, opened 2024-12-23
fix: remove duplicate payjoin code in getAddressWithAmountAndLabel

kumulynja:
This PR removes some code that is executed when payjoinReceiver != null that was present two times in the same function, one time before checking `isLiquid` and one time after checking `isLiquid`. The code before the Liquid check was causing the issue and so I removed it, since if not Liquid, it will still be executed after this check.

A more elaborate fix would also make sure the payjoinReceiver gets reset when isLiquid, but I think a bigger refactoring can eliminate this problem completely, so for now just removing the duplicate code before the check works as a quick fix.

### DIFF
--- a/lib/receive/bloc/state.dart
+++ b/lib/receive/bloc/state.dart
@@ -49,16 +49,6 @@ class ReceiveState with _$ReceiveState {
     if (paymentNetwork == PaymentNetwork.lightning ||
         (amount == 0 && description.isEmpty && payjoinReceiver == null)) {
       finalAddress = address;
-    } else if (payjoinReceiver != null) {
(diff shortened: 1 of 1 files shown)


kumulynja:
This should fix https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/371

ethicnology (approved):
Remove duplicate starting at [line 70](https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/375/files#diff-24485bf9b5e964bcb0d51a05fbb0118203b0f922efab1bf73d5c39fa742b1048L70)

PR #381, opened 2024-12-23
refactor: check of BaseWalletType

ethicnology:
this check may be superfluous, what do you think @i5hi ?


ethicnology:
once merged should Close #370

PR #382, opened 2024-12-23
feat: send flow pj success screen

kumulynja:
This PR adds and propogate the necessary events in the sender isolate and `PayjoinEventBus` to know when a payjoin request is successfully posted.
It also uses that event to progress the send flow and show a payjoin success screen.

In this version the cubit doesn't check if the event stems from the send in progress or from another payjoin isolate, since the event doesn't have any data fields yet. So it assumes the event is from the transaction the user is sending at that moment.
In a later version this can be handled by adding some data to the event to check against the send state.

### DIFF
--- a/lib/_pkg/payjoin/manager.dart
+++ b/lib/_pkg/payjoin/manager.dart
@@ -97,6 +97,11 @@ class PayjoinManager {
       receivePort.listen((message) async {
         print('Sender isolate: $message');
         if (message is Map<String, dynamic>) {
+          if (message['type'] == 'request_posted') {
+            PayjoinEventBus().emit(
+              PayjoinSenderPostMessageASuccessEvent(),
+            );
+          }
           if (message['type'] == 'psbt_to_sign') {
             final proposalPsbt = message['psbt'] as String;
             final (wtxid, err) = await _walletTx.signAndBroadcastPsbt(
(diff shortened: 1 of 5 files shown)


lib/_pkg/payjoin/manager.dart
```diff
@@ -440,6 +445,8 @@ Future<String?> _runSender(Sender sender) async {
 
     // Attempt V2
     final postRes = await _postRequest(dio, postReq);
+    // TODO: make an abstract class for these send port messages and subclasses for each type
+    sendPort.send({'type': 'request_posted'});
```
DanGould, line 449: I Think this call should be made just a few lines later, after getCtx is successfully parsed from response. Only then can the caller confirm that their request was received by the directory and ought to be available to the payjoin receiver. Otherwise, the request could have received a 500 response from the directory or similar if my understanding of dio is correct.

DanGould:
> In this version the cubit doesn't check if the event stems from the send in progress or from another payjoin isolate, since the event doesn't have any data fields yet. So it assumes the event is from the transaction the user is sending at that moment.

This is low risk because the trigger for this event happens outside of the loop, only when the isolate starts. So it would only trigger UI change if a the app was brought to the foreground on this particular "confirming payjoin" pending state OR if you somehow managed to open the app and input to get to get to that state from startup (likely impossible even with scripted input)

lib/_pkg/payjoin/manager.dart
kumulynja: Moved it a line lower here: https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/385

Issue #389, opened 2024-12-24
[Testnet] During an On-chain/Liquid send transaction when On-chain/Liquid address QR is scanned first then followed by entering the 'Amount to send' observing 'Send' button is not getting enabled.


kiranmetri:
https://github.com/user-attachments/assets/049e3142-c9db-4b41-ae09-9f737ccdf73f

kumulynja:
This should be fixed once https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/390 is merged.

kiranmetri:
Issue is resolved on the latest APK (v0.3.8), so closing this issue.

PR #390, opened 2024-12-24
fix: check balance on amount change

kumulynja:
This PR exposes the checkBalance function of the SendCubit so it can be called again on amount change with a CurrencyCubit listener, which was previously done by calling the updateAddress function with null and false parameters but caused bugs when entering bip21 URI's.
This fix makes sure the checkBalance is called, triggering enabling or disabling the send button, without the other side effects of calling the updateAddress again.


kumulynja:
This should fix https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/389

PR #392, opened 2024-12-24
Select first utxo for payjoin contribution

DanGould:
The tryPreservingPrivacy function avoids some varieties of Unnecessary Input Heuristic but sometimes finds no available selection. We always want to contribute an input for the sake of demonstration, and the resulting transaction can't be worse for privacy than a naive transaction, it just might be identifiable as a payjoin.

In the future, we definitely want more sophisticated coin selection, but for now, this makes more payjoins so we get more feedback from testing to make a more reliable, legible product.


i5hi:
Thanks @DanGould

PR #395, opened 2024-12-26
fix: core.init() for startup resume + remove await + complete and clean up sessions

kumulynja:
This PR fixes some of the problems with resuming sessions at startup and with the sessions and isolates in general, namely:

- core.init() was needed in main to be able to reinitiate persisted sessions fromJson
- the resumeSessions now spawns all in parallel since otherwise another stored session couldn't be spawned before the previous was completed. Since it is unsure when a session will complete, it was not good to let all other sessions wait for the previous to complete.
- The completer of the spawnReceiver was never completed and the sessions were never cleaned, so isolates just kept piling up over time.
- The spawnSender didn't complete everywhere I think is necessary either, neither did the clean up work, this is now fixed by using the pjUrl as an identifier to store the isolates and clean them up.
- A parameter is added to the PostMessageASuccessEvent to be able to only progress the send flow for a post of that particular send and not because of another session emitting this event at the same time.
- Removed the setting of the state to `sent` in payjoinSend, as well as syncing and the delay, since they are not useful here. A payjoin shouldn't go to the success screen, so shouldn't come on `sent: true`, neither should it have a delay and then sync, since at this stage the tx is not broadcasted yet, only the original psbt is posted to the pj directory.

### DIFF
--- a/lib/_pkg/payjoin/event.dart
+++ b/lib/_pkg/payjoin/event.dart
@@ -30,8 +30,9 @@ class PayjoinBroadcastEvent extends PayjoinEvent {
 }
 
 class PayjoinSenderPostMessageASuccessEvent extends PayjoinEvent {
-  // TODO: add to relate the event to a specific send transaction/payjoin session
-  PayjoinSenderPostMessageASuccessEvent();
+  final String pjUri;
+
+  PayjoinSenderPostMessageASuccessEvent({required this.pjUri});
 }
 
 class PayjoinFailureEvent extends PayjoinEvent {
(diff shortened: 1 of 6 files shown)


DanGould (commented):
I understand why so many changes were made in one commit for the sake of expedience, and still I hope in the future we can practice commit hygiene that separates them.

Because the commit message doesn't yet include the "why" in the body I'd like to see a merge commit that includes the body of this PR so that someone down the line finding this change knows why its changes made were made.

DanGould:
tACK [57dc788](https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/395/commits/57dc7889a279f7d457541f59446bff1995a96073) closed & resumed receiver

DanGould (approved):
tACK [57dc788](https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/395/commits/57dc7889a279f7d457541f59446bff1995a96073) closed & resumed receiver

Just a smoke test but a smoke test nonetheless

PR #400, opened 2024-12-27
Mark successful payjoin sessions in storage.

DanGould:
Payjoin sessions marked 'complete' do not get spawned. This reduces the number of isolates spawned on resume or restart.

Mark rather than delete so that the historic status may eventually be shown in transaction history.


i5hi:
Thanks @DanGould

Issue #399, opened 2024-12-27
[Payjoin] Receive address should show a loader while creating a new address


i5hi:
Some feedback is required while a PJ address is being created.

ethicnology:
#409 merged

PR #409, opened 2024-12-30
show spinner while loading payjoin address

kumulynja:
This PR makes sure that loadingAddress is still true when payjoin is enabled and the payjoin receiver is still being initialized.
It then uses the loadingAddress state to show a spinner instead of the QR and addresses as long as it is true, effectively showing the address only when the payjoin uri is also constructed.


kumulynja:
This should fix https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/399.

ethicnology:
Close #399

PR #412, opened 2025-01-02
Mark unrecoverable errors so they don't spawn

DanGould:
Mark unrecoverable errors in storage so that they no longer spawn isolates.

Spawning isolates and failing them immediately uses more resources than necessary. This is therefore a performance improvement for the stability of the app.

In the medium term we should switch on errors from payjoin-flutter, but those errors are not yet available to switch on. This will reduce the amount of error definition and handling we have to manually implement, but since this is a potential immediate performance issue I'm addressing it straight away.

### DIFF
--- a/lib/_pkg/payjoin/manager.dart
+++ b/lib/_pkg/payjoin/manager.dart
@@ -26,6 +26,23 @@ const List<String> _ohttpRelayUrls = [
 
 const payjoinDirectoryUrl = 'https://payjo.in';
 
+sealed class SendError {
+  final String message;
+
+  const SendError.SessionError(this.message);
+
+  factory SendError.recoverable(String message) = RecoverableError;
+  factory SendError.unrecoverable(String message) = UnrecoverableError;
+}
+
+class RecoverableError extends SendError {
+  const RecoverableError(super.message) : super.SessionError();
+}
+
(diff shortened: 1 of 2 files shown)


DanGould:
> unrecoverabl

lib/_pkg/payjoin/manager.dart
```diff
@@ -281,6 +301,9 @@ class PayjoinManager {
 
       return completer.future;
     } catch (e) {
+      if (e is UnrecoverableError) {
```
kumulynja: @DanGould I think this never gets here and should be moved to the receive port listener where the `UnrecoverableError` is send to from the isolate. So I guess instead of the `message is Err` on line 283, check for and handle the `UnrecoverableError` there.

DanGould: I made this change as well as one where if errors in the switch statement crop up we mark the receiver payjoin session as unrecoverable which I detail in the commit message. Let me know what you think.

bitcoin-b (approved):
LGTM!

kumulynja (approved):
This works now 👍

i5hi:
Thanks @DanGould!

Thanks @kumulynja and @bitcoin-b  for the review!

Issue #403, opened 2024-12-27
[Payjoin] App occasionally freezes on startup


i5hi:
After closing the app and reopening the app occasionally freezes (sync loader is stuck)

I assume this is due to the multiple isolates being spawned.

DanGould:
did #416 or #417 fix this?

i5hi:
Not sure which of those two fixed it; but its been fixed. Possibly the combination of both.

DanGould:
I do not have permission to close the issue

PR #417, opened 2025-01-07
Unnecessary killing and starting of isolates when app is backgrounded

kumulynja:
As far as I know, isolates are backgrounded and resumed together with the main app thread, so I think it is not necessary to kill and resume them explicitly ourselves. Normally they will be paused when the app is backgrounded and resumed when it is foregrounded again.

### DIFF
--- a/lib/_pkg/payjoin/listeners.dart
+++ b/lib/_pkg/payjoin/listeners.dart
@@ -37,24 +37,5 @@ class _PayjoinLifecycleManagerState extends State<PayjoinLifecycleManager>
   }
 
   @override
-  void didChangeAppLifecycleState(AppLifecycleState state) {
-    final inBg = state == AppLifecycleState.inactive ||
-        state == AppLifecycleState.paused ||
-        state == AppLifecycleState.hidden ||
-        state == AppLifecycleState.detached;
-
-    if (!inBackground && inBg) {
-      // App going to background
-      widget.payjoinManager.pauseAllSessions();
-    } else if (inBackground && !inBg) {
-      // App coming to foreground
-      widget.payjoinManager.resumeSessions(widget.wallet);
-    }
-
-    inBackground = inBg;
-    super.didChangeAppLifecycleState(state);
-  }
-
-  @override
   Widget build(BuildContext context) => widget.child;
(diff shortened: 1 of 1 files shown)


kumulynja:
@DanGould I might be wrong here, but also tested it and isolates seem to resume automatically after the app is backgrouned. Did you notice it was needed to resume them explicitly? Or was there any reason you killed and resumed them explicitly on app lifecycle state changes?

kumulynja:
@i5hi might be needed to test this on a real device still. I can only test on emulator for the moment.

DanGould:
My understanding is that processing on the main isolate can freeze the UI, (like #403 which might be caused by the main isolate being blocked by the process starting a bunch of isolates) and since those sessions are all waiting for responses on web requests, my concern was that either many current requests or response handling they could block the main isolate and cause jank in the UI. This remains my primary concern.

If this design is not necessary I am by no means tied to it. The isolates are way more complicated than having everything on main.

The other reason spawning isolates per session made sense to me was because I assumed (wrongly?) that in order to have concurrent requests in the background they would need to be spawned in isolates. My intuition was that we want them to run concurrently and if we wait for the OS to stop them, we might get penalized for not handling background task termination properly. I am informed from studying [iOS background processing tasks](https://developer.apple.com/documentation/backgroundtasks/bgprocessingtask), which I believe carries over in large part to Android, but I may be a few years out of date or misunderstand how Flutter abstractions translate to OS specifics. Simple searches are telling me that `Dio` HTTP operations are generally automatically cancelled when the app goes to the background, which may make my rationale for isolates insufficient. But I'm not sure how concurrent long-polling requests would behave.

The background processing concern I have is from iOS, where if you didn't explicitly clean up the main process the OS will force kill it after some time. However, every time the OS kills your process for you you get penalized in your ability to get background process time / send push notifications, so I beleive the hygienic thing to do is to leave the isolates. Eventually, we want to use the specific OS background processing APIs to allow it to happen in the background for more liveliness. We probably want to use the cross-platform equivalent to [BGAppRefreshTask](https://developer.apple.com/documentation/backgroundtasks/bgapprefreshtask) for this.

PR #429, opened 2025-01-13
Stax recoverbull contributions

ethicnology:
- [x] Gdrive integration during backup
- [ ] Gdrive integration during recovery
- [x] Selectable backup
- [x] Recovery with conflicts (not enough information to recover a specific wallet)
- [x] Partial recovery (recover only some information of a existing wallet)
- [ ] BIP85 derivation path management


lib/_pkg/gdrive.dart
```diff
@@ -1,18 +1,16 @@
-import 'dart:convert';
-
 import 'package:extension_google_sign_in_as_googleapis_auth/extension_google_sign_in_as_googleapis_auth.dart';
 import 'package:google_sign_in/google_sign_in.dart';
 import 'package:googleapis/drive/v3.dart';
 
-class Gdrive {
+class GoogleDriveApi {
```
ethicnology, line 5: if we keep gdrive.dart as filename, to enforce consistency we should keep GDrive as classname
if we go for the GoogleDriveApi let's rename the file too

lib/_pkg/gdrive.dart
```diff
@@ -26,34 +24,34 @@ class Gdrive {
...
-      await api.files.update(file, createdFile.id!, uploadMedia: media);
-
-      print("File created");
-      return true;
-    } catch (e) {
-      print("Error: $e");
-      return false;
-    }
-  }
+  // Future<bool> write({required String filename, required Map content}) async {
```
ethicnology, line 32: before merging we should remove the comments

lib/_pkg/gdrive.dart
ethicnology, line 1: ~~It seems this class is not needed anymore in favour of https://github.com/StaxoLotl/recoverbull-dart~~

After internal discussions we decided to make cloud integration part of the wallet codebase and remove it from the recoverbull-client

lib/backup/bloc/cloud_state.dart
ethicnology, line 1: as mentioned earlier cloud code should be extracted from recoverbull and implemented directly into the wallet

ethicnology (commented):
I have to review and rework recoverbull client

We will move out the cloud integration related code from the client as a wallet pkg.

ethicnology:
Included in a bigger PR #434

PR #443, opened 2025-01-27
Refactor/gradle update

StaxoLotl:
### PR: Refactor/Gradle Update
- **Gradle Wrapper**: Upgraded to version 8.7.
- **Build Config**: Updated to Kotlin and Gradle's latest versions; now using Java 21 and SDK 34.
- **Package Name**: Changed to `com.bullbitcoin.mobile`.
- **Dynamic Library**: Fixed loading issue by updating Boltz package source.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,11 +1,11 @@
 buildscript {
-    ext.kotlin_version = '1.9.0'
+    ext.kotlin_version = '1.9.20'
     repositories {
         google()
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.2.1'
+        classpath 'com.android.tools.build:gradle:8.5.0'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
@@ -20,8 +20,8 @@ subprojects {
         if (project.plugins.hasPlugin("com.android.application") ||
             project.plugins.hasPlugin("com.android.library")) {
             project.android {
-                compileSdkVersion 35
-                buildToolsVersion "34.0.0" 
+                compileSdk = 34
+                buildToolsVersion "34.0.0"
             }
         }
         if (project.hasProperty('android')) {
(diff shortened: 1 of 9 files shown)


android/build.gradle
```diff
@@ -1,11 +1,11 @@
 buildscript {
-    ext.kotlin_version = '1.9.0'
+    ext.kotlin_version = '1.9.20'
     repositories {
         google()
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:8.2.1'
+        classpath 'com.android.tools.build:gradle:8.5.0'
```
ethicnology, line 8: is it normal the `classpath` gradle version (8.5) doesn't match the `distributionUrl` version (8.7)

pubspec.yaml
```diff
@@ -75,8 +75,10 @@ dependencies:
   auto_size_text: ^3.0.0
   no_screenshot: ^0.0.1+6
   flutter_svg: ^2.0.10+1
-  boltz: ^0.1.8
-    # git:
+  boltz:
+    git:
+      url: https://github.com/StaxoLotl/boltz-dart.git
```
ethicnology, line 80: Since ishi committed https://github.com/SatoshiPortal/boltz-dart/commit/0a5444a4fcca46f9fb43d4a3cfd4b3f1e3662711 you can set back the original repo targetting develop

ios/Podfile.lock
```diff
@@ -138,29 +138,29 @@ EXTERNAL SOURCES:
     :path: ".symlinks/plugins/url_launcher_ios/ios"
 
 SPEC CHECKSUMS:
-  bdk_flutter: 08274535b7bd58a09bfa9c3733474fb44133484c
+  bdk_flutter: fb57a7400a7f3f181c5977bcdc2a5ef347ae4e7f
   boltz: 6388ec2412f3753b63a9e65c97f87ea26f43bddc
```
ethicnology, line 142: Isn't this should be updated since boltz is not the same version?

ethicnology (changes requested):

android/build.gradle
saambaby: That's because AGP 8.5 requires minimum 8.7

pubspec.yaml
StaxoLotl: > [0a5444a](https://github.com/SatoshiPortal/boltz-dart/commit/0a5444a4fcca46f9fb43d4a3cfd4b3f1e3662711) won't be enough us, since cargokit gives the toolchain mentioned cargokit.yaml, the first priority when building the binaries. so we need to include [a182a58](https://github.com/SatoshiPortal/boltz-dart/commit/a182a5854da98301df740b24ddda4cf543ef5558)

android/build.gradle
StaxoLotl: That's because AGP 8.5 requires minimum 8.7

PR #493, opened 2025-02-27
fix: issues-476, Physical backup not switch to tested if the wallet has an Encrypted Backup tested

StaxoLotl:
…kupTested


ethicnology:
This PR is a mistake

PR #502, opened 2025-03-03
Refactor/recoverbull

StaxoLotl:


lib/_pkg/recoverbull/_interface.dart
```diff
@@ -12,6 +12,12 @@ import 'package:hex/hex.dart';
 import 'package:recoverbull/recoverbull.dart';
 
 abstract class IRecoverbullManager {
+  /// Initialize connection and verify access
+  Future<(dynamic, Err?)> connect();
```
ethicnology, line 16: Why adding these functions to the abstract class if they shouldn't exists in some of the implementation such as Local storage

lib/_pkg/recoverbull/local.dart
```diff
@@ -8,10 +8,19 @@ import 'package:bb_mobile/locator.dart';
 import 'package:recoverbull/recoverbull.dart';
 
 class FileSystemBackupManager extends IRecoverbullManager {
-  final FileStorage fileStorage = locator<FileStorage>();
-
   FileSystemBackupManager();
 
+  @override
+  Future<(String?, Err?)> connect() async {
```
ethicnology, line 14: related to unnecessary functions in abstract class

lib/recoverbull/bloc/backup_settings_cubit.dart
```diff
@@ -130,182 +142,147 @@ class BackupSettingsCubit extends Cubit<BackupSettingsState> {
...
+
+      if (decodeErr != null || backup == null) {
+        debugPrint('Error decoding backup: ${decodeErr?.message}');
+        _handleLoadError("Corrupted backup file");
+        return;
       }
+
+      final backupMap = backup.toMap();
+      backupMap.addAll({
+        'source': 'drive',
```
ethicnology, line 272: I was expecting that we get rid of this

ethicnology (changes requested):

PR #506, opened 2025-03-05
Added Swap Watcher Service

i5hi:
- [x] Refactor repository and services to ensure domain does not depend on data layer
- [x] Add swap watcher & swap processing
- [x] Add basic swap interaction with wallets 
- [x] Implement Fees & Limits logic


lib/_core/data/services/swap_watcher_service_impl.dart
```diff
@@ -0,0 +1,345 @@
...
+import 'package:bb_mobile/_core/domain/entities/swap.dart';
+import 'package:bb_mobile/_core/domain/services/swap_watcher_service.dart';
+import 'package:bb_mobile/_core/domain/services/wallet_manager_service.dart';
+import 'package:boltz/boltz.dart' as boltz_type;
+
+class SwapWatcherServiceImpl implements SwapWatcherService {
+  final WalletManagerService _walletManager;
+  final BoltzSwapRepositoryImpl _boltzRepo;
+
+  StreamSubscription<boltz_type.SwapStreamStatus>? _swapSubscription;
```
i5hi: We have to depend on SwapStreamStatus here because this type is tightly coupled to the BoltzWebSocket and I am not sure how we can instead return a custom type. 

This is the type for the message returned by the websocket onEvent

lib/_core/data/services/swap_watcher_service_impl.dart
```diff
@@ -0,0 +1,345 @@
...
+
+  SwapWatcherServiceImpl({
+    required WalletManagerService walletManager,
+    required BoltzSwapRepositoryImpl boltzRepo,
+  })  : _walletManager = walletManager,
+        _boltzRepo = boltzRepo;
+
+  @override
+  Future<void> startWatching() async {
```
i5hi: Removed this from the constructor, so that it can be called somewhere else, maybe the bloc? 

Or should I still have it in the constructor?

lib/_core/data/services/wallet_manager_service_impl.dart
i5hi, line 1: I will revert these changes as they will cause conflicts to you @kumuly.

i5hi: reverted.

lib/_core/data/datasources/boltz_storage_data_source.dart
```diff
@@ -0,0 +1,121 @@
...
+  @override
+  Future<ChainSwap> getChainSwap(String swapId) async {
+    final key = '${SecureStorageKeyPrefixConstants.swap}$swapId';
+    final jsonSwap = await _secureSwapStorage.getValue(key) as String;
+    return ChainSwap.fromJson(jsonStr: jsonSwap);
+  }
+}
+
+extension EnvironmentToChain on Environment {
+  Chain toBtcChain() {
```
kumulynja: Since this extension makes the `toBtcChain` and `toLbtcChain` on the `Environment` type also accesible from outside of this boltz related datasource, it would be better to make it clear that they are related to Boltz and not to Btc or Lbtc chains in the rest of the app. So I think it is better to add that to the name like `toBoltzBtcChain` and `toBoltzLbtcChain` so it is a bit clearer. Similar as the extension function called `bdkNetwork` on `Network`, making it clear it is for the bdk dependency.

lib/_core/data/datasources/boltz_storage_data_source.dart
```diff
@@ -0,0 +1,121 @@
...
+import 'package:boltz/boltz.dart';
+
+abstract class BoltzStorageDataSource {
+  // Local Storage
+  Future<void> store(SwapModel swap);
+  Future<SwapModel?> get(String swapId);
+  Future<List<SwapModel>> getAll();
+  Future<void> delete(String swapId);
+  // Secure Storage
+  Future<void> storeBtcLnSwap(BtcLnSwap swap);
```
kumulynja: Is it possible to make every concrete Boltz Swap type (BtcLnSwap, LbtcLnSwap and ChainSwap) have a common base type? If so, it would not be required to have a separate method that does exactly the same thing for every type. This will reduce the amount of code needed.

lib/_core/data/models/swap_model.dart
```diff
@@ -58,17 +59,17 @@ class SwapModel with _$SwapModel {
       keyIndex: keyIndex,
       chainSwapDetails: chainSwapJson == null
```
kumulynja: I think the model and it's field definitions are good, I would just rename it to BoltzSwapModel to be more specific that the fields json fields are related to the boltz data source specifically.
And, instead of having these three optional fields (`chainSwapJson`, `lnReceiveSwapJson`, `lnSendSwapJson`), it is better to  make `BoltzSwapModel` a sealed class and have a factory method for every type separately. See `pdk_payjoin_model.dart` that also has different json String fields for sender and receiver types.

lib/_core/data/services/swap_watcher_service_impl.dart
kumulynja: Since the `SwapStreamStatus` messages are related to a specific swap, we should be able to reconstruct a `BoltzSwapModel` from it in the `boltz_data_source`, change the status of this `BoltzSwapModel` with the new status as received from the `SwapStreamStatus`, and then probably also update the `BoltzSwapModel` with this new status in local storage for later resume in case the app is closed. The boltz_data_source can then expose a stream of models instead of the dependency type. Like this:

```
@override
  @override
  Stream<BoltzSwapModel> get stream => _boltzLib.stream.map((event) {
    // Get model by id (event.id)
    // Update model with event status
    // (possibly store the updated model in the local storage)
    // Return the model so it is streamed to listeners
  });
  ```

lib/_core/data/services/swap_watcher_service_impl.dart
kumulynja: If it can be self-contained, without another part of the code having to worry about calling it or where/when to call it, it would be better, so I would put it back in the constructor so it is never possible that the service is in a "not-listening" state.

lib/_core/data/services/swap_watcher_service_impl.dart
i5hi: fixed

kumulynja (changes requested):
I think the main things to change are combining storage and api again to just a boltz_data_source class, or at least inject the storage one in the api one, so that creating a swap contains both creating the swap with the api as persisting it, and returning one `BoltzSwapModel` type instead of library types.

Then the second main thing is also using the model in the boltz SwapStatusStream processing so the types of the stream the datasource exposes also is `BoltzSwapModel` and not a library type, keeping all dependencies on boltz-dart abstracted in the datasource.

i5hi:
@kumulynja This is ready for final review and merge.

lib/_core/data/datasources/boltz_storage_data_source.dart
i5hi: Its possible to make a common base type for the *LnSwap but ChainSwap has a different interface. It will require some additional code in the library in dart beyond what is generated automatically by flutter_rust_bridge. 

Although, I think the explicit methods are clear and safe ? We won't need to check if what we got back is a Btc or Lbtc swap because the method will always return exactly the one the caller wants.

lib/_core/data/datasources/boltz_storage_data_source.dart
i5hi: Removed all this and just using `isTestnet` instead

lib/_core/data/services/swap_watcher_service_impl.dart
i5hi: Done.

lib/_core/data/models/swap_model.dart
i5hi: Done.

PR #508, opened 2025-03-07
Payjoin receive and send

kumulynja:
This PR adds the data layer using PDK with session persistence, continuous fetching of the payjoin directory in a separate isolate and automatic resume at startup. The data layer streams events when a payjoin request or proposal is received and the higher layers (payjoin repository and payjoin service) are implemented to listen and process these events. The payjoin service uses other services and repositories like the Wallet service, Electrum Server repository and others to verify, sign and broadcast the payjoins. The payjoin service exposes a stream the rest of the application can listen to to receive updates of payjoin sessions while they get processed.
Also the use cases to create a payjoin, send a payjoin and listen for updates of a specific payjoin were added.

To make sure the code works, an end-2-end integration test of a wallet of the app receiving with a payjoin from another wallet of the app is added as well. For now it is just a happy case test. Negative and more specific test cases should be implemented still.

This should complete the full refactor of payjoin for now though as it has a working happy case path with the new architecture.


i5hi:
Thanks @kumulynja !

PR #535, opened 2025-03-17
fix: MnemonicSeed missing passphrase

ethicnology:


ethicnology:
External review @kumulynja

kumulynja:
Good catch! Thanks @ethicnology

PR #536, opened 2025-03-17
refactor: remove basic warnings/infos related to unused imports, trailing comma…

ethicnology:


ethicnology:
I guess we will do a big cleanup before merging into develop

PR #554, opened 2025-03-20
Receive feature

kumulynja:
This is WIP on the receive feature that has the way to generate Bitcoin and Liquid addresses and bip21 uri's with amount and for Bitcoin also payjoin parameters. It also makes it possible to generate lightning invoices with a swap. It supports entering amounts in different currencies, both in bitcoin (sats/btc) amounts as in fiat amounts and it uses the exchange rate datasource to show the correct equivalents in fiat or bitcoin respectively.

Known issues to solve/add still:

- The animation of the Segmented Buttons to switch between Bitcoin, Lightning and Liquid doesn't seem to work, the colors do not change accordingly when pressing a network. Two times have to be pressed to have it reflect correctly.
- The currency dropdown on the amount screen should be styled and positioned better still
- Not listening to swap stream to know if a (lightning) payment was received yet
- Success screen not implemented yet
- Bottomsheet to add a note should be added
- Note/Description can not be passed to lightning swap
- Hide the 'New address' button for Lightning
- The 'New address' button for Liquid only gives one new address, then it keeps returning the 'last unused index + 1'. We should keep track of the generated indexes for lwk wallets ourselves as it doesn't have the way to increase the index persistently as with BDK. I would do this with a simple key/int value storage in the lwk datasource once we refactor lwk stuff from repository to datasource.
- If the user has BTC as his preffered denomination in settings, but he enters the amount in sats denomination, we should maybe show it in sats as well on the QR screen instead of in BTC. (Similar as is done when the amount is entered as a fiat currency different than the preferred one in settings)
- Error messages in the UI and handling of specific errors like no UTXO's to payjoin/ Boltz swaps not available etc.
- Swap fee info in QR screen and check for min and max amount for swaps.


i5hi:
Priority issues would be:

- [ ] Not listening to swap stream to know if a (lightning) payment was received yet
- [ ] Success screen not implemented yet
- [ ] Hide the 'New address' button for Lightning
- [ ] Swap fee info in QR screen and check for min and max amount for swaps.
- [ ] If the user has BTC as his preffered denomination in settings, but he enters the amount in sats denomination, we should maybe show it in sats as well on the QR screen instead of in BTC. (Similar as is done when the amount is entered as a fiat currency different than the preferred one in settings)

Issue #673, opened 2025-04-23
For a lightning send, the confirm screen need not show the entire invoice


thibistaken:
Confirm screen can just say `Lightning Invoice` and on click display the full invoice in a small overlay popup like an info popup (the from wallet didnt show up because the app needed to be deleted and reinstalled because of migration - so that part can be ignored). 

For context: https://satoshi-portal.slack.com/archives/C08LDGZ8KED/p1744814605537789

kumulynja:
Solved by https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/677.

Issue #672, opened 2025-04-23
Address display in confirm Send screen with line wrapping issue


thibistaken:
For context: https://satoshi-portal.slack.com/archives/C08LDGZ8KED/p1744814152677989

kumulynja:
Solved by https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/677.

PR #677, opened 2025-04-24
show shorthand address or invoice on send confirmation

kumulynja:


kumulynja:
Should solve https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/672 and https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/673

PR #806, opened 2025-05-13
Update to payjoin 0.23.0

DanGould:
via payjoin-flutter @ DanGould/deps-for-bb

The persistence API changed but is incomplete, and so not worth moving to on this commit. Instead, I use an in-memory persister as a hack to bridge this release to the next one when the real API is available.

I am had some issues running the test, but I guess it's from my environment needing java21 (where an emulator is running) and not bb-mobile's setup. Copying the .env.template verbatim into .env also worked (Though for some reason mere `touch .env` did not)

### DIFF
--- a/lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
+++ b/lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
@@ -49,7 +49,7 @@ class PdkPayjoinDatasource {
   Stream<PayjoinModel> get expiredPayjoins => _expiredController.stream;
 
   Future<(OhttpKeys?, Url?)> fetchOhttpKeyAndRelay({
-    required Url payjoinDirectory,
+    required String payjoinDirectory,
   }) async {
     Url? ohttpRelay;
     OhttpKeys? ohttpKeys;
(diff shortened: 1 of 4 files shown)


DanGould:
This seems to build but I don't have all of the `.env` keys which seem to be required to make the integration test run. Would be nice if API keys were stored in github CI tokens so that outside contributors could test without needing to ask for the keys

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -77,25 +77,43 @@ class PdkPayjoinDatasource {
...
+          return receiver;
+        },
+      );
+      final token = await newReceiver.persist(persister: noOpPersister);
+
+      final receiver = await Receiver.load(
+        token: token,
+        persister: noOpPersister,
+      );
+
```
spacebear21, line 116: Is the completer necessary here? Can we use this more succinct approach like with the Sender below? (it compiles)

```suggestion
      final imp = InMemoryReceiverPersister();
      final persister = DartReceiverPersister(
        save: (receiver) async {
          return await imp.save(receiver: receiver);
        },
        load: (token) async {
          return await imp.load(token: token);
        },
      );
      final token = await newReceiver.persist(persister: persister);
      final receiver = await Receiver.load(
        token: token,
        persister: persister,
      );

```

PR #810, opened 2025-05-16
feat: BBQR codes are collected, joined, decompressed and deserialized as bdk PSBT. The PSBT is not processed. (Available for SuperUser in Debug only)

ethicnology:
- [x] iOS (missing files in the flutter_rust_bridge package)
- [ ] Refactor the next screen to make work the whole workflow until a payment (fix the send types etc…)
- [ ] Automatically switch scan mode instead of manually


ethicnology:
rebased on `develop`

PR #977, opened 2025-07-06
feat: `make feature <NameYourFeature>` 

ethicnology:


i5hi:
Some more context would help. 

I was expecting it to be minimal boilerplate but seems like this is an example implementation of a feature that queries an IP but its not very clear exactly what it does.

lib/features/template/presentation/template_cubit.dart
```diff
@@ -0,0 +1,62 @@
...
+  final GetCachedIpUsecase _getCachedIpUsecase;
+
+  TemplateCubit({
+    required CollectAndCacheIpUsecase collectAndCacheIpUsecase,
+    required GetCachedIpUsecase getCachedIpUsecase,
+  }) : _collectAndCacheIpUsecase = collectAndCacheIpUsecase,
+       _getCachedIpUsecase = getCachedIpUsecase,
+       super(const TemplateState());
+
+  void clearError() => emit(state.copyWith(error: ''));
```
i5hi: clearErrors should make the error null so we can do a null check for existence of an error

lib/features/template/ui/template_page.dart
```diff
@@ -0,0 +1,160 @@
...
+
+class TemplatePage extends StatelessWidget {
+  const TemplatePage({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: const Text('IP Address Info'),
+        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
```
i5hi, line 21: need the front end specialists opinion on this but im generally in favor of creating a 

final theme = Theme.of(context); and using theme variable

lib/features/template/ui/template_page.dart
```diff
@@ -0,0 +1,160 @@
...
+      ),
+      body: BlocConsumer<TemplateCubit, TemplateState>(
+        listener: (context, state) {
+          if (state.error != null && state.error!.isNotEmpty) {
+            ScaffoldMessenger.of(context).showSnackBar(
+              SnackBar(
+                content: Text(state.error!),
```
i5hi: We should use BBText

lib/features/template/ui/template_page.dart
```diff
@@ -0,0 +1,160 @@
...
+                backgroundColor: Colors.red,
```
i5hi: We should only use colors from Theme.of(context)

lib/features/template/ui/template_page.dart
```diff
@@ -0,0 +1,160 @@
...
+              ],
+            ),
+            const SizedBox(height: 12),
+            _infoRow('User Agent', ip.userAgent),
+            _infoRow('Compression', ip.isCompressionSupported ? 'Yes' : 'No'),
+            _infoRow('Timestamp', ip.timestamp.toString()),
+            const SizedBox(height: 8),
+            const Text(
+              'Supported Encodings:',
+              style: TextStyle(fontWeight: FontWeight.w600),
```
i5hi: TextStyle should also come from Theme.of(context)

lib/features/template/presentation/template_state.dart
```diff
@@ -0,0 +1,16 @@
...
+part 'template_state.freezed.dart';
+
+enum Redirection { none, toSomewhereElse }
+
+@freezed
+sealed class TemplateState with _$TemplateState {
+  const factory TemplateState({
+    @Default(Redirection.none) Redirection redirection,
+    @Default(false) bool isLoading,
+    String? error,
```
i5hi: Use a class for specific Exception type. Part of a larger discussion on where to place these errors. In some cases we have it in the state file itself; in other cases it is in core/errors

i5hi: i am in favor of standardizing all errors going into core/errors/$feature_name.dart

lib/features/template/presentation/template_cubit.dart
ethicnology: making a String variable nullable force us to check both `!= null` and `.isNotEmpty`

In case of multiple error type it would make sense thought

lib/features/template/presentation/template_state.dart
ethicnology: Why putting them in `core` instead of the feature folder `feature/feature_name_errors.dart`?

lib/features/template/ui/template_page.dart
ethicnology: fb8d44adafaa7196b0ec453a98732a1103a6f9cd

lib/features/template/ui/template_page.dart
ethicnology: fb8d44adafaa7196b0ec453a98732a1103a6f9cd

lib/features/template/ui/template_page.dart
ethicnology: fb8d44adafaa7196b0ec453a98732a1103a6f9cd

lib/features/template/ui/template_page.dart
ethicnology: fb8d44adafaa7196b0ec453a98732a1103a6f9cd

lib/features/template/presentation/template_state.dart
ethicnology: 0247297adeb353b0b6c2922da40a8af904dd5798

lib/features/template/presentation/template_cubit.dart
ethicnology: 0247297adeb353b0b6c2922da40a8af904dd5798

PR #978, opened 2025-07-07
refactor: send

i5hi:
This PR is not ready for merge at the time of creation. 

Udates made to send cubit:

Currently old behavious is not changed. New methods are created to replace old ones.

1. Created copies of the following methods:

- onAddress/onAmountConfirmed -> onAddress/onAmountConfirmed`Refactor`

2. SendType -> SendAddressType. Move to state getter rrather than variable. 

Potential additions: separate cubit methods called by the UI and itternal methods (like bloc). Internal methods should not update the state and should return a Ok or Error value. Currenrtly internal methods update the state and if they are called the caller has to explicitly check if the state has any errors. 

- [ ] Tested.

### DIFF
--- a/integration_test/_values.dart
+++ b/integration_test/_values.dart
@@ -1,0 +1,191 @@
+// https://jlopp.github.io/xpub-converter
+
+class TestValues {
+  // XPUB
+  static String get xpub =>
+      'xpub6DJwRncrB8eNrzUq8XxgjwCZsEeWP8FeqBJbJQZ8JfuDwLdAzyjhHiHJieNuar1wjQTyihhMWtaKGE4DUd8uBgtyrNJqF5drwbNVUqb83b7';
+  static String get xpubFingerprint => '2bcd33b0';
+  // conversions to other formats
+  static String get xpubToYpub =>
+      'ypub6Y9CjTHmKpBriHfwxtkJx2J53CnxKkF9kHpp5oT1ggH6zSSQFduFumwSjrLVakfs93anUBHuyYvs9WfnCKYuyvaaii1FpzTMDKS8sVJo82W';
+  static String get xpubToZpub =>
+      'zpub6ryU37xgUVjLZas4oFXwA7PaDAwQGNEefQM2sCLu4gez3YFdWJ4pXqbam4J5afKnYghbDetUSDHR2oHLv1xvnAGBb3hgQuGqV3VnG2rvas6';
+  static String get xpubToTpub =>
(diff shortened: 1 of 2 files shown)


ethicnology:
We should use `lib/features/send/request_identifier` as the entry point of send.

Alone, this sub-route take care of identifying/validating the user input from either scan or paste

lib/features/send/presentation/bloc/send_state.dart
```diff
@@ -13,50 +13,59 @@ import 'package:freezed_annotation/freezed_annotation.dart';
...
+}
+
 @freezed
 abstract class SendState with _$SendState {
   const factory SendState({
     @Default(SendStep.address) SendStep step,
-    @Default(SendType.lightning) SendType sendType,
     @Default('') String scannedRawPaymentRequest,
     @Default('') String copiedRawPaymentRequest,
     PaymentRequest? paymentRequest,
```
ethicnology: We should make paymentRequest non-nullable passed by the `request_identifier`

lib/features/send/presentation/bloc/send_state.dart
```diff
@@ -99,12 +108,16 @@ abstract class SendState with _$SendState {
...
+    // exceptions
     SwapCreationException? swapCreationException,
     InsufficientBalanceException? insufficientBalanceException,
     InvalidBitcoinStringException? invalidBitcoinStringException,
     SwapLimitsException? swapLimitsException,
     BuildTransactionException? buildTransactionException,
     ConfirmTransactionException? confirmTransactionException,
+    ExchangeApiException? exchangeApiException,
+    FeesException? feesException,
+    LoadWalletException? loadWalletException,
```
ethicnology, line 118: Instead of explicitly listing each type of exception in the state, I suggest we use a single `Exception? error` field. This way, it can represent any kind of error, ensuring that we handle all cases consistently and don’t accidentally miss any.

ethicnology (dismissed):
I will try to plug the `request_identifier`

ethicnology:
@i5hi Can you review my changes ?

- `RequestIdentifier` can take an optional `Wallet?`
- `RequestIdentifier` pass `Wallet?` and `PaymentRequest` named tuple as Extra to the `SendAmountScreen`
- Integration of the `RequestIdentifier` with the send feature which allow a big cleanup of all code related to scan/paste parsing

ethicnology:
rebased on `develop`

ethicnology:
rebased on `develop` with conflicts resolved

PR #979, opened 2025-07-08
Feat/download backup

StaxoLotl:


ethicnology:
> On the backup settings page, can we add "download backup file" and "download backup key" buttons? The download backup key option should first have a warning about not saving it on the same device or same place.

> I think we need to do that, in case key server goes offline. As long as we out proper disclaimers in place and we make it scary for newbies

To download the backup key we need either:
- Recoverbull Backup + KeyServeer
- Recoverbull Backup + Mnemonic loaded in the wallet (from which we will derive.

The straightforward way to solve this need is:
1. The user HAS to select the backup, because as pointed out by Stax we don't know where the user decided to store it.
2. Once the recoverbull backup is selected, we can attempt to derive the key and decipher it with the current loaded wallet. If it works, we display the key.
3. If step 2 fails, we need to require the user PIN to fetch it from the Key Server.
4. If 2 and 3 fails, we can't display the key.

lib/core/recoverbull/data/datasources/google_drive_datasource.dart
```diff
@@ -33,6 +33,20 @@ class GoogleDriveAppDatasource {
     }
   }
 
+  Future<void> connectSilently() async {
+    try {
+      final account = await _google.signInSilently();
```
ethicnology: I recall that we didn't want to `signInSilently` anymore because it was introducing some unexpected behaviors

lib/features/backup_settings/presentation/cubit/backup_settings_cubit.dart
```diff
@@ -1,47 +1,255 @@
...
+            (e) => e.isEncryptedVaultTested,
+          ),
+          lastPhysicalBackup:
+              defaultWallets
+                  .firstWhere((e) => e.network == Network.bitcoinMainnet)
+                  .latestPhysicalBackup,
+          lastEncryptedBackup:
+              defaultWallets
+                  .firstWhere((e) => e.network == Network.bitcoinMainnet)
+                  .latestEncryptedBackup,
```
ethicnology: For the sake of readability can we assign these values into a variable first and then emit using these variables

lib/features/backup_settings/presentation/cubit/backup_settings_cubit.dart
```diff
@@ -1,47 +1,255 @@
...
+      }
+      final (content, fileName) =
+          await _fetchLatestGoogleDriveBackupUsecase.execute();
+      final folderPath = await _selectFolderPathUsecase.execute();
+      if (folderPath == null) {
+        emit(state.copyWith(status: BackupSettingsStatus.initial));
+        return;
+      }
+      final filePath = '$folderPath/$fileName';
+      await _saveToFileSystemUsecase.execute(filePath, content);
```
ethicnology, line 108: Should we use and pass a File directly ?

lib/features/onboarding/presentation/bloc/onboarding_bloc.dart
```diff
@@ -156,7 +156,9 @@ class OnboardingBloc extends Bloc<OnboardingEvent, OnboardingState> {
       final encryptedBackup = await switch (state.vaultProvider) {
         FileSystem(:final fileAsString) => _fetchBackupFromFileSystemUsecase
             .execute(fileAsString),
-        GoogleDrive() => _fetchLatestGoogleDriveBackupUsecase.execute(),
+        GoogleDrive() => _fetchLatestGoogleDriveBackupUsecase.execute().then(
+          (result) => result.$1,
```
ethicnology: Please use named [records](https://dart.dev/language/records) because we have no idea of what $1 is.

ethicnology (changes requested):
Few comments to improve readability

The only blocking is related to sign-in silently, we need to be sure that we want to re-introduce it in this feature because I recall when we removed it in the past due to unexpected behaviors.

@i5hi do you recall it too ?

lib/features/backup_settings/presentation/cubit/backup_settings_cubit.dart
StaxoLotl: we are not actually fetching a file from the datasource. Google drive's fetchBackupContent returns bytes. So I am doing the filename formatting in the usecase and file creation in fileSystemRepository.saveFile .

lib/features/backup_settings/presentation/cubit/backup_settings_cubit.dart
StaxoLotl: 54e22147a9c75a988f77a07b393fdc01dab8ef55

lib/core/recoverbull/data/datasources/google_drive_datasource.dart
StaxoLotl: a0d43b980d93ca925d5a63c4e598d4b710a5f48a

ethicnology:
LGTM @i5hi  only a doubting on sign-in silently if it was necessary or not

i5hi:
IIRC we didnt want sign-in silently for the case where a user wants to login with a different email id.

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
```diff
@@ -40,9 +44,31 @@ class _Screen extends StatelessWidget {
     final isSuperuser = context.select(
       (SettingsCubit cubit) => cubit.state.isSuperuser ?? false,
     );
+
     return BlocConsumer<BackupSettingsCubit, BackupSettingsState>(
       listener: (context, state) {
-        // Add global state handling if needed
+        if (state.downloadedBackupFile != null) {
```
kumulynja: Better to use the `listenWhen` parameter and use the previous value of the state, instead of conditions in the `listener`, because then you can easier and more clearly control that it only gets triggered if the `downloadedBackupFile` value changes, now if `downloadedBackupFile` is not null and another state value changes, this will still be triggered and executed again, since the state changed and `downloadedBackupFile` is still not null.
In the `listenWhen` you can check that the previous value of `downloadedBackupFile` was null.

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
```diff
@@ -40,9 +44,31 @@ class _Screen extends StatelessWidget {
...
+          Clipboard.setData(ClipboardData(text: state.downloadedBackupFile!));
+          log.info('Vault exported and copied to clipboard');
+          context.read<BackupSettingsCubit>().clearDownloadedData();
+        }
+        if (state.error != null) {
```
kumulynja: Same comment as above, better to use `listenWhen` and make sure this only gets triggered if state.error was previously null and is currently not null.

Also, use MultiBlocListener to have a separate `listener` and `listenWhen` for the error listener.

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
```diff
@@ -40,9 +44,31 @@ class _Screen extends StatelessWidget {
...
+          log.severe('Export failed: ${state.error}');
+          if (state.error == 'Local backup key derivation failed.') {
```
kumulynja: Better to use concrete error types instead of an hardcoded String.

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
```diff
@@ -225,3 +258,64 @@ class _StartBackupButton extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    return BlocBuilder<BackupSettingsCubit, BackupSettingsState>(
+      builder: (context, state) {
+        return BBButton.big(
+          label:
+              state.status == BackupSettingsStatus.exporting
+                  ? 'Exporting...'
+                  : 'Export Vault',
+          onPressed:
+              state.status == BackupSettingsStatus.exporting
```
kumulynja, line 289: Better to just disable the button instead of adding an empty callback I think, to give feedback to the user.

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
```diff
@@ -225,3 +258,64 @@ class _StartBackupButton extends StatelessWidget {
...
+    return BlocBuilder<BackupSettingsCubit, BackupSettingsState>(
+      builder: (context, state) {
+        return BBButton.big(
+          label:
+              state.status == BackupSettingsStatus.viewingKey
+                  ? 'Revealing...'
+                  : 'View Backup Key',
+          onPressed:
+              state.status == BackupSettingsStatus.viewingKey
+                  ? () {}
```
kumulynja, line 315: Same comment as above, better to disable the button instead of keeping it clickable with an empty callback function.

lib/features/backup_settings/ui/screens/choose_vault_provider_screen.dart
```diff
@@ -0,0 +1,107 @@
...
+      case BackupProviderType.iCloud:
+        log.info('iCloud, not supported yet');
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return BlocListener<BackupSettingsCubit, BackupSettingsState>(
+      listener: (context, state) {
+        if (state.selectedBackupFile != null) {
```
kumulynja: Same comment as above, use `listenWhen` instead of conditions in the `listener`.

lib/features/backup_settings/ui/screens/choose_vault_provider_screen.dart
```diff
@@ -0,0 +1,107 @@
...
+          // Navigate to backup key display screen
+          context.pushNamed(
+            BackupSettingsSubroute.viewBackupKey.name,
+            extra: state.selectedBackupFile,
+          );
+          // Clear the selectedBackupFile to prevent double navigation
+          context.read<BackupSettingsCubit>().clearDownloadedData();
+        }
+        if (state.error != null) {
```
kumulynja: Also, use MultiBlocListener to have a separate `listener` and `listenWhen` for the error listener.

lib/features/backup_settings/ui/screens/view_backup_key_screen.dart
```diff
@@ -0,0 +1,131 @@
...
+}
+
+class _Screen extends StatelessWidget {
+  const _Screen();
+
+  @override
+  Widget build(BuildContext context) {
+    return BlocConsumer<BackupSettingsCubit, BackupSettingsState>(
+      listener: (context, state) {
+        if (state.error != null) {
```
kumulynja: Same as above, better to use the `listenWhen` parameter to check the condition to trigger showing the error snackbar.

lib/features/backup_settings/ui/screens/choose_vault_provider_screen.dart
```diff
@@ -0,0 +1,107 @@
...
+
+          return _buildScaffold(context);
+        },
+      ),
+    );
+  }
+
+  Widget _buildScaffold(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
```
kumulynja: I think you can just set the title instead of using `flexibleSpace` and remove the automaticallyImplyLeading: false.
No need to overwrite default `pop` behavior.

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
StaxoLotl: @kumulynja, instead of reusing the existing exceptions or creating a new error type just for this specific case, I did this instead: state.derivedBackupKey == null && state.status == BackupSettingsStatus.viewingKey. Basically, I kept the BackupSettingsStatus as viewingKey and updated the error object only if an exception occurs.

i5hi: We should still use a structured error - we are following this pattern everywhere now; even if just a single case. the state.error can be a genericRecoverBull error also but it should not just be a plain string

i5hi: and checking against a string is bad because if someone changes the message they wont know to change it here too

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
```diff
@@ -40,43 +44,86 @@ class _Screen extends StatelessWidget {
...
+    return MultiBlocListener(
+      listeners: [
+        BlocListener<BackupSettingsCubit, BackupSettingsState>(
+          listenWhen:
+              (previous, current) =>
+                  previous.downloadedBackupFile !=
+                      current.downloadedBackupFile &&
+                  current.downloadedBackupFile != null,
+          listener: (context, state) {
+            if (state.downloadedBackupFile != null) {
```
kumulynja: Since the listenWhen condition already checks that current.downloadedBackupFile != null, no need to check it again in the listener.

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
```diff
@@ -40,43 +44,86 @@ class _Screen extends StatelessWidget {
...
+              );
+              log.info('Vault exported and copied to clipboard');
+              context.read<BackupSettingsCubit>().clearDownloadedData();
+            }
+          },
+        ),
+        BlocListener<BackupSettingsCubit, BackupSettingsState>(
+          listenWhen: (previous, current) => previous.error != current.error,
+          listener: (context, state) {
+            if (state.error != null &&
```
kumulynja: Why still have this condition in the listener and not in the `listenWhen`?
And feels like this can be simplified. It is not really clear why an error and 2 other fields need to have a specific value for it to navigate, maybe this could be solved if just specific error for this case was used?

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
kumulynja: Yeah, I agree with @i5hi , safer to not just use local String comparison, better to use concrete Error types.

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
StaxoLotl: My idea was not to make more classes, but the concrete error types makes more sense.

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
```diff
@@ -54,21 +55,18 @@ class _Screen extends StatelessWidget {
...
         BlocListener<BackupSettingsCubit, BackupSettingsState>(
-          listenWhen: (previous, current) => previous.error != current.error,
+          listenWhen:
+              (previous, current) =>
+                  previous.error != current.error && current.error != null,
           listener: (context, state) {
-            if (state.error != null &&
-                state.derivedBackupKey == null &&
-                state.status == BackupSettingsStatus.viewingKey) {
+            if (state.derivedBackupKey == null &&
```
kumulynja, line 68: Move this if condition to the `listenWhen` so the listener is not always triggered if the condition isn't met:

```
listenWhen:
              (previous, current) =>
                  previous.error != current.error && 
                  current.error is BackupKeyDerivationFailedError && 
                  current.derivedBackupKey == null,
```

kumulynja (changes requested):

lib/features/backup_settings/ui/screens/backup_settings_screen.dart
i5hi: imo this is fine because listenWhen triggers if there is any type of error and then in the listener we check for the specific error type. If we use the whole condition in listenWhen for BackupKeyDerivationFailedError only then we don't listen to other potential errors.

PR #982, opened 2025-07-09
BREAKING CHANGE: replace `WalletSource` to `Signer` with SQLite migration and tests

ethicnology:
`WalletSource` [`mnemonic`, `descriptors`, `xpub`, `coldcard`] was ambiguous, so we decided to replace this enum with `Signer` [`local`, `remote`, `none`].

Since `WalletSource` was stored in the `wallet_metadatas` table under the `source` column, this change requires a data migration.

This commit includes:
- replacing `source` with `signer`
- SQLite migration from `source` to `signer` in the `wallet_metadatas` table
- schema version bump from 3 to 4
- migration test from schema v3 to v4


lib/core/storage/migrations/schema_v4.dart
```diff
@@ -0,0 +1,48 @@
...
+          signer = Signer.remote.name;
+        default:
+          signer = Signer.none.name;
+      }
+
+      // Update the new signer value
+      await db.customUpdate(
+        'UPDATE wallet_metadatas SET signer = ? WHERE id = ?',
+        variables: [Variable.withString(signer), Variable.withString(id)],
+      );
```
kumulynja: Could this be optimised with the following custom sql statement?

```
await db.customStatement('''
    UPDATE wallet_metadatas
    SET signer = CASE signer
      WHEN 'mnemonic' THEN 'local'
      WHEN 'descriptors' THEN 'remote'
      ELSE 'none'
    END
''');
```

kumulynja (approved):
Just a small optimisation comment above, but lgtm!

lib/core/storage/migrations/schema_v4.dart
ethicnology: Good idea I will change this

ethicnology: updated

lib/core/storage/migrations/schema_v4.dart
```diff
@@ -0,0 +1,31 @@
...
+    await db.customStatement(
+      'ALTER TABLE wallet_metadatas RENAME COLUMN source TO signer',
+    );
+
+    // Map old wallet_source values to new signer enum values
+    await db.customStatement('''
+    UPDATE wallet_metadatas
+    SET signer = CASE signer
+      WHEN 'mnemonic' THEN 'local'
+      WHEN 'descriptors' THEN 'remote'
```
i5hi, line 27: descriptors should also be local; although it doesnt matter because no one has this type

ethicnology: What kind of descriptors were imported back then ?
I assumed that was external signers descriptors like we do now

i5hi: it was intended to import private or public descriptors but we didnt actually expose it to users

i5hi: so technically it could be either

i5hi: what about handling the coldcard case?

i5hi: should be remote, no?

PR #983, opened 2025-07-12
WIP: Address list view

kumulynja:
This PR adds the Address List View to Wallet Details.

A refactoring of the wallets list and details screens was needed, especially to remove WalletBloc being passed in WalletDetailsCubit and in the end remove the WalletDetailsCubit all together since the WalletBloc could be used directly for what is needed, also remove unnecessary BlocConsumers and Builders and move navigation Listeners to the route definitions. Also the wallet deletion functionality is moved to the WalletBloc and optimised.

WIP: The address list functionality itself is still WIP, the above refactoring had to be done first.


lib/core/swaps/data/repository/boltz_swap_repository_impl.dart
```diff
@@ -507,23 +507,26 @@ class BoltzSwapRepositoryImpl implements SwapRepository {
...
+                  swap is ChainSwap && swap.receiveWalletId == walletId) &&
+              ((swap.status == SwapStatus.pending) ||
+                  (swap.status == SwapStatus.paid) ||
+                  (swap.status == SwapStatus.canCoop) ||
+                  (swap.status == SwapStatus.claimable) ||
+                  (swap.status == SwapStatus.refundable) ||
+                  (swap is ChainSwap &&
+                      swap.status == SwapStatus.completed &&
+                      swap.receiveTxid == null &&
+                      swap.refundTxid == null)),
```
ethicnology: This condition is hard to read. Let's put all `SwapStatus` into a list and check if `swap.status` is contained in it?

ethicnology (approved):
I didn't test/review presentation layer

I've made a single suggestion to improve readability of a condition using `list.contains` instead of multiple SwapStatus comparisons

lib/core/swaps/data/repository/boltz_swap_repository_impl.dart
kumulynja: Yeah, I agree, but I was thinking that it would be better to add some filters to the `fetchAll` function and so do the filtering in the datasource level with drift/sql, instead of fetching all and then looping through all from the repository. But I didn't want to change the datasource in this PR now to not expand the scope and possible regressions with swaps.

What do you think of adding more possibilities of filtering to the datasource in another PR instead of still refactoring this now with contains?

ethicnology: - If you do it in another PR you will have to rebase one of your work on top of the other.
- If you do the change in the current one, this PR may be longer to be merged.

I would say make a new PR and we will see which one will be merged first, but I guess it's fine if you do it here too

ethicnology:
rebased on `develop`

PR #984, opened 2025-07-15
fix: testnet unconfirmed balance and swaps showing up on mainnet and vice versa

i5hi:
- Ensures getAllSwaps and related queries are filtered by environment
- Updates BoltzStorageDatasource.fetchAll() to support isTestnet filtering
- Modifies SwapRepository and usecases to pass environment filtering
- Prevents cross-network swap and balance display issues

### DIFF
--- a/lib/core/swaps/data/repository/boltz_swap_repository_impl.dart
+++ b/lib/core/swaps/data/repository/boltz_swap_repository_impl.dart
@@ -507,8 +507,8 @@ class BoltzSwapRepositoryImpl implements SwapRepository {
   }
 
   @override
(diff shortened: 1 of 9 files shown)


i5hi:
yah; i had this feeling too; will make the change.

lib/core/swaps/data/repository/boltz_swap_repository_impl.dart
ethicnology, line 1: We may take profit of this fix to make `BoltzSwapRepository` concrete

i5hi: done

PR #990, opened 2025-07-16
add BuyError and use in buy input screen for min/max amounts

kumulynja:
The first buy order screen didn't show any errors in case the min or max amount were exceeded yet, keeping the user guessing of what the problem might be. This PR fixes this buy adding a BuyError with subtypes for min and max amount errors (and others that can be used in the future).

It also moves specific Buy feature use cases to the Buy feature, which makes more sense.

And it replaces the SingleChildScrollView of the `BuyInputScreen`with a newly added, reusable widget `ScrollableColumn`, as recommended here: https://api.flutter.dev/flutter/widgets/SingleChildScrollView-class.html, which we should use for screens that should be scrollable while still wanting to position Buttons at the bottom of the screen without the button being hidden when the keyboard is open, which was the case before with having the button in the BottomNavigationBar field of the scaffold.
The use of `BBText` had to be eliminated for the widgets inside the `ScrollableColumn`, because `BBText` can not be used inside of a `LayoutBuilder` because of the use of the dependency autosize... which we should get rid off as well imo.


lib/features/buy/domain/buy_error.dart
i5hi, line 1: nice!

i5hi (approved):

lib/features/bitcoin_price/ui/currency_text.dart
```diff
@@ -62,6 +61,10 @@ class CurrencyText extends StatelessWidget {
       text = '** ${text.split(' ').last}';
     }
 
-    return BBText(text, style: style, color: color, textAlign: textAlign);
+    return Text(
```
ethicnology, line 64: I thought we should always use `BBText`, should we patch `BBText` or something?

ethicnology (approved):
Internal discussion regarding the usage of `BBText`

lib/features/bitcoin_price/ui/currency_text.dart
kumulynja: imo we should stop using it, it doesn't add any value over the standard Text widget and it can not be used in LayoutBuilders, so we can't even use it anywhere. Best would be to replace all BBText widgets with normal Text and then delete the BBText widget.

ethicnology:
rebased on `develop`

ethicnology:
Approved twice merging

PR #991, opened 2025-07-17
feat: ColdCard Q integration PSBT / BBQR / PushTx (NFC)

ethicnology:
- Repaired the BBQR scanner since the replacement by `flutter_zxing`
- Reduced scan delay for BBQR codes
- Moved transaction mapper functions to entity directly
- Unified QR and NFC scanning infrastructure
- **File reorganization**: Scanner widgets moved to `core/widgets/`
- Custom error handling 

### PushTx
- Added `NfcScannerWidget` with `flutter_nfc_kit` integration
- NDEF record parsing for PushTx URIs
- **SDK requirements**: Android SDK `26+` and iOS `13.0+`
- Dual broadcasting: PushTx or Broadcast within the app the decoded transaction from PushTx Uri
- Browser integration for PushTx URLs

### Database
- **Schema v5**: Added `signerDevice` **nullable** column to `wallet_metadatas`
- New `SignerDevice.coldcardQ` enum


lib/core/transaction/domain/entities/tx.dart
```diff
@@ -1,21 +1,113 @@
...
+import 'package:bdk_flutter/bdk_flutter.dart' as bdk;
 import 'package:freezed_annotation/freezed_annotation.dart';
 
 part 'tx.freezed.dart';
 part 'tx.g.dart';
 
 @freezed
-abstract class Tx with _$Tx {
-  const factory Tx({
+abstract class TxEntity with _$TxEntity {
```
kumulynja: I feel like the transaction folder name and also here TxEntity can be a bit confusing to understand it is different from Wallet Transactions or other types of transactions. What about `RawBitcoinTx` or something more telling?

Also, we only use this for the Coldcard/hardware wallet feature now, right? In that case, it would be better to not have a separate folder `transaction` either, but move these models and entities to the coldcard/hardware wallet feature, and then make an entity specifically for the transaction data needed to sign with a hardware wallet instead of trying to use this general entity for all features that need Bitcoin transaction data.

lib/core/transaction/domain/entities/tx.dart
```diff
@@ -1,21 +1,113 @@
...
     required BigInt vsize,
     required int locktime,
     required List<TxVin> vin,
     required List<TxVout> vout,
-  }) = _Tx;
+  }) = _TxEntity;
+
+  const TxEntity._();
+
+  factory TxEntity.fromJson(Map<String, dynamic> json) =>
```
kumulynja: Does the domain or UI needs this entity in JSON format?

If not, then this shouldn't be in the entity.

lib/core/transaction/domain/entities/tx.dart
```diff
@@ -1,21 +1,113 @@
...
+      _$TxEntityFromJson(json);
+
+  static TransactionModel toModel(TxEntity tx) {
```
kumulynja: The model, which is from the data layer shouldn't be in the domain layer.

Move this to the model if needed.

lib/core/transaction/domain/entities/tx.dart
```diff
@@ -1,21 +1,113 @@
...
+      version: tx.version,
+      size: tx.size.toString(),
+      vsize: tx.vsize.toString(),
+      locktime: tx.locktime,
+      vin: json.encode(tx.vin),
+      vout: json.encode(tx.vout),
+    );
+  }
+
+  factory TxEntity.fromSqlite(TransactionModel row) {
```
kumulynja: This also mixes things from the data layer, like the model and sqlite into the domain layer.

Move this to the model.

lib/features/experimental/broadcast_signed_tx/locator.dart
```diff
@@ -0,0 +1,23 @@
...
+    registerUsecases();
+    registerBlocs();
+  }
+
+  static void registerUsecases() {
+    // Add any use cases here if needed in the future
+  }
+
+  static void registerBlocs() {
+    locator.registerFactoryParam<BroadcastSignedTxCubit, void, void>(
```
kumulynja: Why does it use `registerFactoryParam` and not just `registerFactory` if no parameters are passed?
Both are void...
Is this because you expect it to need runtime params in the future?

lib/features/experimental/broadcast_signed_tx/router.dart
```diff
@@ -0,0 +1,69 @@
...
+import 'package:bb_mobile/features/experimental/broadcast_signed_tx/presentation/pages/broadcast_signed_tx_page.dart';
+import 'package:bb_mobile/features/experimental/broadcast_signed_tx/presentation/pages/scan_nfc_page.dart';
+import 'package:bb_mobile/features/experimental/broadcast_signed_tx/presentation/pages/scan_qr_page.dart';
+import 'package:bb_mobile/locator.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:go_router/go_router.dart';
+
+enum BroadcastSignedTxRoute {
+  broadcastHome('/broadcast-signed-tx'),
+  broadcastScanQr('scan-qr'),
```
i5hi, line 12: some routes have / some dont...is there a reason for that?

i5hi: just checked other features; we don't need the "/" it seems for routes

lib/features/experimental/broadcast_signed_tx/presentation/broadcast_signed_tx_state.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:bb_mobile/core/bbqr/bbqr_service.dart';
+import 'package:bb_mobile/features/experimental/broadcast_signed_tx/type.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+part 'broadcast_signed_tx_state.freezed.dart';
+
+@freezed
+abstract class BroadcastSignedTxState with _$BroadcastSignedTxState {
+  const factory BroadcastSignedTxState({
+    required BbqrService bbqr,
```
kumulynja: This feels a bit weird to store a service in the UI state.

lib/features/experimental/experimental_router.dart
```diff
@@ -9,7 +9,7 @@ class ExperimentalRouterConfig {
     routes: [
       ImportWatchOnlyRouterConfig.route,
       PsbtRouterConfig.route,
-      ScanSignedTxRouterConfig.route,
+      BroadcastSignedTxRouter.route,
```
kumulynja, line 12: It feels like PsbtRouter and BroadcastSignedTxRouter are part of a bigger functional end-user feature and not a feature in itself that should have their own separate router, especially not seeing that PsbtRouterConfig has only one route. I think we should think how to structure this better in a clear feature where it fits and just have one router for that feature.

But this is something we can define outside of this PR and is not only related to this feature, just mentioning it here.

lib/features/experimental/psbt_flow/show_psbt/show_psbt_screen.dart
```diff
@@ -28,10 +29,10 @@ class _ShowPsbtView extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return Scaffold(
-      backgroundColor: context.colour.surface,
       appBar: AppBar(
-        title: const Text('PSBT'),
-        backgroundColor: context.colour.surface,
+        forceMaterialTransparency: true,
+        automaticallyImplyLeading: false,
+        flexibleSpace: TopBar(title: 'PSBT', onBack: () => context.pop()),
```
kumulynja: Since this is the only one route in the PsbtRouter and there is not even a shared Bloc/cubit in the ShellRoute, I think the ShellRoute can be removed from the PsbtRouter and then the default leading widget can be used, so no need to overwrite the onBack or even use flexibleSpace or TopBar.

lib/features/send/presentation/bloc/send_cubit.dart
```diff
@@ -9,6 +9,7 @@ import 'package:bb_mobile/core/fees/domain/get_network_fees_usecase.dart';
 import 'package:bb_mobile/core/payjoin/domain/usecases/send_with_payjoin_usecase.dart';
 import 'package:bb_mobile/core/settings/domain/get_settings_usecase.dart';
 import 'package:bb_mobile/core/settings/domain/settings_entity.dart';
+import 'package:bb_mobile/core/storage/tables/wallet_metadata_table.dart';
```
kumulynja: This is something to review and avoid as it is mixing data layer into presentation layer, that should not be done.

lib/features/experimental/broadcast_signed_tx/router.dart
kumulynja: This depends on if it is a subroute or a top level route.

lib/features/experimental/broadcast_signed_tx/locator.dart
ethicnology: AI…

Does the cubit even need to be declared?

lib/core/transaction/domain/entities/tx.dart
ethicnology: 7434b9f2988fa932d1de5569cd5e4c2ce9e7a8e5

lib/core/transaction/domain/entities/tx.dart
ethicnology: 7434b9f2988fa932d1de5569cd5e4c2ce9e7a8e5

lib/core/transaction/domain/entities/tx.dart
ethicnology: 7434b9f2988fa932d1de5569cd5e4c2ce9e7a8e5

lib/core/transaction/domain/entities/tx.dart
ethicnology: 7434b9f2988fa932d1de5569cd5e4c2ce9e7a8e5

lib/features/experimental/broadcast_signed_tx/presentation/broadcast_signed_tx_state.dart
ethicnology: 0bf8562456d5b3a402f5cf2282f92471918c0489

lib/features/experimental/broadcast_signed_tx/locator.dart
ethicnology: 643f790f6d4ba2af99e894543186d73dd8acd710

lib/features/experimental/psbt_flow/show_psbt/show_psbt_screen.dart
ethicnology: 0f1ae34ba00105e971814ded4cde9505fffca783

lib/features/send/presentation/bloc/send_cubit.dart
ethicnology: The enum `Signer` is related to the `wallets_metadatas` table.

Usually I make table related enum into the table file to make sure that others devs understand it's a sensitive data structure, some changes to this enum may need a migration or will break the app.

The enum is imported to check : 
`if (state.selectedWallet!.signer == Signer.remote)`

Signer is the replacement of previously used `WalletSource`.

To separate the layers, I should wrap `Signer` into a `SignerEntity` used in `Wallet`is it correct ?

ethicnology: 18604ee842cd8bc26413cc02e80d97719b712f67

kumulynja: Yes, and it is exactly because it's sensitive that it is better to have a separate enum in the domain, to avoid that someone changes and breaks the table enum accidentally.
I understand it feels like unnecessary duplication, but after a while someone might think: "We need another name for the domain in this enum, so let's just change it in the enum."
If the domain has its own enum for it, even if it's duplication now, changing it will not break the db, because it means the converter (`toModel()`) will need to be changed consciously as well. So the separation between the two is actually kind of a safeguard against accidentally changing the table enum because of a change only needed in the domain. And, if needed at some point, it also makes it possible for the domain enum to evolve independently without needing db migrations.

ethicnology: I agree, unfortunately this pattern appears in various place, for example `ScriptType` in `Wallet` and `WalletModel` 

We should enforce suffixing all entities by `Entity` to avoid the mistake

PR #995, opened 2025-07-17
uncomment and fix payjoin integration tests

kumulynja:


ethicnology (approved):
👍 I will work on test CI to ensure we avoid breaking tests in the future

PR #1000, opened 2025-07-21
Updated settings

i5hi:
Updated settings to new UI and added placeholders for exchange settings.


ethicnology:
`squash` the two first commits to be able to rebase on `develop`

lib/features/settings/ui/screens/all_settings_screen.dart
```diff
@@ -0,0 +1,177 @@
...
+                    );
+                    launchUrl(url, mode: LaunchMode.inAppBrowserView);
+                  },
+                ),
+              ],
+            ),
+          ),
+        ),
+      ),
+      bottomNavigationBar: BottomAppBar(
```
kumulynja, line 98: I think we should change this bottomNavigatonBar's to just show something at the bottom, with the ScrollableColumn. So replace the SingleChildScrollView and Column above with the ScrollableColumn widget, and then use a Spacer to push these bottom widgets down.

kumulynja: But, we can always do that in another PR, this works for now.

lib/features/settings/ui/screens/exchange/logout_screen.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:bb_mobile/core/widgets/navbar/top_bar.dart';
+import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
+
+class ExchangeLogoutScreen extends StatelessWidget {
```
kumulynja, line 5: Does this need a separate screen or just a dialog that asks if the user is sure of wanting to log out?

kumulynja (approved):

lib/features/settings/ui/screens/exchange/logout_screen.dart
i5hi: just a dialog should be fine; we will implement it in another PR

lib/features/settings/ui/screens/all_settings_screen.dart
i5hi: ok; will do this in another PR. We need to return to this again with a dedicated PR for exchange settings; we will make sure we address this there.

PR #1003, opened 2025-07-21
fix: iphone SE reported not seeing `Open the camera` button

ethicnology:
Temporary work around is tapping anywhere on `OpenTheCameraWidget` open the scanner


ethicnology:
@i5hi commit amended

simulator_screenshot_395E0570-BADB-40A8-A1A3-DA151D5101FD

PR #1005, opened 2025-07-21
fix: 958 migrate liquid wallet origin

ethicnology:
- fix the typo
- include the migration to ~~v5~~ v4 migration (there is no schema breaking)
- test ~~v5~~ v4 schema migration (signer_device + id)
- externalize directory `path_provider` instanciation from `Logger.init` to make it works in migration tests

⚠️ ~~I've added it to the existing v5 migration since the latest release `main` is at schema v3.~~ **Devs and TestFlight testers will have to re-install the app**

⚠️⚠️ The last commit is merging schema 5 and migration to v5 into schema4 and migration to v4 since main is still on schema3


i5hi (approved):
LGTM

PR #1006, opened 2025-07-22
Feat/view cloud backups

StaxoLotl:


lib/core/recoverbull/domain/entity/drive_file.dart
```diff
@@ -1,11 +1,32 @@
...
   final String id;
   final String name;
   final DateTime createdTime;
 
-  DriveFile({
-    required this.id,
-    required this.name,
-    required this.createdTime,
-  });
+  DriveFile({required this.id, required this.name, required this.createdTime});
```
ethicnology, line 8: I'm now strongly in favor of using the `Entity` suffix for class names and `_entity.dart` for file names, consistent with how we name other architectural components.

Would you agree to rename it `DriveFileEntity` and `drive_file_entity.dart`

lib/core/recoverbull/domain/entity/drive_file.dart
```diff
@@ -1,11 +1,32 @@
...
+      if (parts.length < 2) {
+        return name.replaceAll('.json', '');
+      }
+
+      final lastPart = parts.last;
+      final cleaned = lastPart
+          .replaceAll('.json', '')
+          .replaceAll(RegExp(r'[\[\]\s]'), '');
+
+      if (!RegExp(r'^[\d,]+$').hasMatch(cleaned)) {
```
ethicnology, line 23: We need some comments explanation there about all these regex used

lib/core/recoverbull/domain/usecases/complete_cloud_backup_verification_usecase.dart
```diff
@@ -0,0 +1,41 @@
...
+  }) : _walletRepository = walletRepository;
+
+  Future<void> execute() async {
+    try {
+      final defaultWallets = await _walletRepository.getWallets(
+        onlyDefaults: true,
+        environment: Environment.mainnet,
+      );
+      if (defaultWallets.isEmpty) {
+        throw Exception('No default wallet found');
```
ethicnology: Let's add it to your custom errors

lib/core/recoverbull/domain/usecases/restore_encrypted_vault_from_backup_key_usecase.dart
```diff
@@ -37,47 +33,40 @@ class RestoreEncryptedVaultFromBackupKeyUsecase {
...
+      if (defaultWallets.isNotEmpty) {
+        if (isVerifying &&
+            !defaultWallets.every(
+              (w) =>
+                  w.masterFingerprint ==
+                  decodedRecoverbullWallet.masterFingerprint,
+            )) {
+          throw const WalletMismatchError();
         }
+        throw DefaultWalletExistsError(decodedRecoverbullWallet);
```
ethicnology: ⚠️ If `decodedRecoverbullWallet` may contain sensitive information like the mnemonic it shouldn't be thrown in an Exception that may be logged and the log shared with the support team.

lib/core/recoverbull/domain/usecases/restore_encrypted_vault_from_backup_key_usecase.dart
```diff
@@ -37,47 +33,40 @@ class RestoreEncryptedVaultFromBackupKeyUsecase {
...
+
+      if (duplicateWallet) {
+        throw const WalletAlreadyExistsError();
       }
 
-      log.info('Default wallets updated');
+      // No default wallet exists and no duplicate - return for default wallet creation
+      return decodedRecoverbullWallet;
     } catch (e) {
       log.severe('$RestoreEncryptedVaultFromBackupKeyUsecase: $e');
```
ethicnology, line 62: :warning: `throw DefaultWalletExistsError(decodedRecoverbullWallet);` will be logged there!

lib/core/wallet/domain/usecases/import_wallet_usecase.dart
```diff
@@ -0,0 +1,76 @@
...
+          scriptType: scriptType,
+          label: seed.masterFingerprint,
+          isDefault: false,
+        ),
+        _wallet.createWallet(
+          seed: seed,
+          network: liquidNetwork,
+          label: seed.masterFingerprint,
+          scriptType: scriptType,
+          isDefault: false,
```
ethicnology: @i5hi do we want to create 1 bitcoin + 1 liquid wallet for each imported ?

lib/features/onboarding/ui/screens/choose_encrypted_vault_provider_screen.dart
```diff
@@ -57,25 +57,23 @@ class _ScreenState extends State<_Screen> {
...
+                  OnboardingRoute.availableGoogleBackupsForRecovery.name,
+                  extra: state.availableCloudBackups,
+                )
+                .then((_) {
+                  if (mounted) {
+                    bloc.add(const EndTransitioning());
+                  }
+                });
+          } else {
+            bloc.add(const EndTransitioning());
```
ethicnology, line 70: Can we simplify this statement by removing `then` and `else` and performing `bloc.add(const EndTransitioning());` after the if once we `awaited` for the `context.pushNamed`

lib/features/onboarding/ui/screens/onboarding_recovery_success.dart
ethicnology, line 1: suffix the file and class name by screen/page

lib/features/onboarding/ui/widgets/recovered_wallet_cards.dart
ethicnology, line 1: suffix the file and class by widget

lib/features/test_wallet_backup/ui/screens/choose_encrypted_vault_provider_screen.dart
```diff
@@ -31,9 +31,14 @@ class _ChooseVaultProviderScreenState extends State<ChooseVaultProviderScreen> {
   }
 }
 
-class _Screen extends StatelessWidget {
+class _Screen extends StatefulWidget {
```
ethicnology, line 34: Does it need to be Stateful?

ethicnology (changes requested):
⚠️ `decodedRecoverbullWallet` may contain sensitive information like the mnemonic, it shouldn't be thrown in an Exception that may be logged and shared with the support team.

ethicnology:
rebased on `develop` to resolve conflicts

lib/core/recoverbull/domain/entity/drive_file.dart
StaxoLotl: Well, I’ve kept it that way intentionally — so far, most entities don’t have an Entity suffix. Instead, we use a Model suffix for models, like Swap and SwapModel. I suggest we either rename all our class names to follow this pattern or revert the ones that currently use the Entity suffix. @i5hi, what do you think?

ethicnology: You are right so far most entities don't have it and that's why I suggest change this pattern. This behavior on entities is inconsistent regarding what we do for usecases, datasources, models and so on.

I don't think we need to either rename them all or accept it this way. 

Now that we have this review process we can simply pay attention to it and slowly refactoring the codebase without troubles. Actually it's what we are doing to convert abstract repositories into concrete implementation. We don't convert them all in a single PR, we refactor when we catch them.

Refactoring everything in a single PR is not a funny job, I did it many times, you end up with conflicts for all other opened PR that need to be rebased.

i5hi: I personally dont like the idea of having Entity in its name; especially for Entity because its the core type.

lib/core/wallet/domain/usecases/import_wallet_usecase.dart
i5hi: yes; and recoverbull is only for default wallets so its always 1 bitcoin + 1 liquid. if you have default wallets you cannot recover.

i5hi: this PR is only for viewing all existing backups; recovering all the backups is not part of the scope for this; that requires some thought and spec.

lib/features/test_wallet_backup/ui/screens/choose_encrypted_vault_provider_screen.dart
StaxoLotl: No it's not needed anymore

i5hi (commented):
Simulator Screenshot - iPhone 16 Pro - 2025-07-31 at 19 37 52

When a wallet is loaded there are a few problems:

- There is no indicator showing the wallet is syncing so we see a 0 balance which takes a while ot update.
- If we chose try another; we see all the previous wallets too and they all get recovered

i5hi:
In the multiWallet PR https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/1028
We have a `DirtyUsecase` which needs to be cleaned up - which basically syncs the wallet without importing it. We should use that OR we should delete the wallet when the user says try another. 

I would prefer the first approach.

i5hi:
Simulator Screenshot - iPhone 16 Pro - 2025-07-31 at 19 43 22

Simulator Screenshot - iPhone 16 Pro - 2025-07-31 at 19 43 35


No loading indicator to show that wallets are syncing

i5hi:
Simulator Screenshot - iPhone 16 Pro - 2025-07-31 at 19 37 23

PR #1016, opened 2025-07-23
add default liquid wallet filter for selected wallet to start buy flow

kumulynja:
As requested: Always select instant wallet as default when buying Bitcoin, regardless of amount .


i5hi (approved):
LGTM

PR #1017, opened 2025-07-24
Access wallet settings from wallet page 

i5hi:
Access wallet settings from wallet page. Also Updated network and address type in address details

### DIFF
--- a/lib/core/wallet/domain/entities/wallet.dart
+++ b/lib/core/wallet/domain/entities/wallet.dart
@@ -117,19 +117,19 @@ abstract class Wallet with _$Wallet {
   const Wallet._();
 
   String get id => origin;
-  String getAddressType() {
+  String get addressType {
     if (isLiquid) {
       return 'Confidential Segwit';
     }
 
     return switch (scriptType) {
       ScriptType.bip84 => 'Native Segwit',
-      ScriptType.bip49 => 'Compatible Script',
+      ScriptType.bip49 => 'Nested Segwit',
       ScriptType.bip44 => 'Legacy',
     };
   }
 
-  String getWalletTypeString() {
+  String get walletTypeString {
     String name = switch (network) {
       Network.bitcoinMainnet || Network.bitcoinTestnet => 'Bitcoin network',
       Network.liquidMainnet ||
@@ -139,7 +139,7 @@ abstract class Wallet with _$Wallet {
     return name;
   }
 
-  String getNetworkString() {
+  String get networkString {
     return switch (network) {
       Network.bitcoinMainnet => 'Bitcoin Network',
       Network.bitcoinTestnet => 'Bitcoin Testnet',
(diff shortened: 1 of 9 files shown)


lib/core/wallet/domain/entities/wallet.dart
```diff
@@ -117,6 +117,17 @@ abstract class Wallet with _$Wallet {
...
 
   String get id => origin;
+  String getAddressType() {
+    if (isLiquid) {
+      return 'Confidential Segwit';
+    }
+
+    return switch (scriptType) {
+      ScriptType.bip84 => 'Native Segwit',
+      ScriptType.bip49 => 'Compatible Script',
```
ethicnology: It's supposed to be `Native Segwit` and `Nested Segwit` but if we want to avoid mixing the two terms and simplify we could do:
- `Native Segwit` into `Segwit`
- `Compatible Script` into `Nested Segwit`

lib/features/settings/ui/screens/bitcoin/wallet_options_screen.dart
```diff
@@ -40,16 +40,17 @@ class WalletOptionsScreen extends StatelessWidget {
...
-                            icon: Icons.currency_bitcoin,
-                            title: 'Addresses',
-                            onTap: () {
-                              context.pushNamed(
-                                SettingsRoute.walletAddresses.name,
-                                pathParameters: {'walletId': walletId},
-                              );
-                            },
-                          ),
+                          if (!wallet.isLiquid)
```
kumulynja: Why no addresses for Liquid?

lib/core/wallet/domain/entities/wallet.dart
ethicnology: Better to make it a `getter` if there is no params

lib/core/wallet/domain/entities/wallet.dart
```diff
@@ -128,6 +139,13 @@ abstract class Wallet with _$Wallet {
     return name;
   }
 
+  String getNetworkString() {
```
ethicnology: Better to make it a `getter` if there is no params

ethicnology (approved):
I would convert `Compatible Script` into `Nested Segwit`, the official name and convert `getSomething()` function into getters

lib/core/wallet/domain/entities/wallet.dart
i5hi: done

lib/core/wallet/domain/entities/wallet.dart
i5hi: done

lib/features/settings/ui/screens/bitcoin/wallet_options_screen.dart
i5hi: reintroduced; i thought we didnt implement addresses for liquid.

PR #1018, opened 2025-07-24
fix: Order Status causing listOrders to fail. Use order status in list.

i5hi:
We were not handling an order status correctly. This fixes that.

Also, instead of showing Pending we now show order status. For example an expired order was showing as Pending. So now only if it is completed we show a tick and time; else we show the order status.

### DIFF
--- a/lib/features/exchange/presentation/exchange_cubit.dart
+++ b/lib/features/exchange/presentation/exchange_cubit.dart
@@ -38,6 +38,8 @@ class ExchangeCubit extends Cubit<ExchangeState> {
       final userSummary = await _getExchangeUserSummaryUsecase.execute();
 
       emit(state.copyWith(userSummary: userSummary));
+      emit(state.copyWith(selectedLanguage: userSummary.language));
+      emit(state.copyWith(selectedCurrency: userSummary.currency));
     } catch (e) {
       log.severe('Error during init: $e');
       if (e is ApiKeyException) {


lib/core/exchange/data/models/user_preference_payload_model.dart
```diff
@@ -0,0 +1,29 @@
+class UserPreferencePayload {
```
kumulynja: Just for consistency, call it `UserPreferencePayloadModel`, since there might be an entity as well.

kumulynja: Any reason to not use freezed and have the toMap function out-of-the-box? Less code is better in this case imo, and freezed is already a dependency anyways.

lib/core/widgets/buttons/multi_tap_trigger.dart
```diff
@@ -59,7 +60,7 @@ class _MultiTapTriggerState extends State<MultiTapTrigger> {
     messenger.clearSnackBars();
     messenger.showSnackBar(
       SnackBar(
-        content: Text(
+        content: BBText(
```
kumulynja, line 63: What is the added value of BBText here?
We should deprecate and remove the use of it, as it uses a package that will make us not catch certain overflows and it now made the MultiTapTrigger widget unusable in places where a LayoutBuilder is needed.

lib/core/widgets/inputs/copy_input.dart
```diff
@@ -100,7 +100,7 @@ class CopyInput extends StatelessWidget {
             backgroundColor: Colors.white,
             title:
                 modalTitle != null
-                    ? Text(
+                    ? BBText(
```
kumulynja, line 103: Same as above, why use BBText still?

lib/core/widgets/inputs/copy_input.dart
```diff
@@ -128,15 +128,15 @@ class CopyInput extends StatelessWidget {
                       ClipboardData(text: clipboardText ?? text),
                     );
                   },
-                  child: const Text('Copy'),
+                  child: BBText('Copy', style: theme.textTheme.bodyLarge),
```
kumulynja, line 131: Same as above, why use BBText still?

lib/core/widgets/inputs/copy_input.dart
```diff
@@ -128,15 +128,15 @@ class CopyInput extends StatelessWidget {
...
                 ),
               TextButton(
                 style: TextButton.styleFrom(
                   foregroundColor: theme.colorScheme.primary,
                   textStyle: theme.textTheme.bodyLarge,
                 ),
                 onPressed: () => Navigator.of(context).pop(),
-                child: const Text('Close'),
+                child: BBText('Close', style: theme.textTheme.bodyLarge),
```
kumulynja, line 139: Same as above, why use BBText still?

lib/core/widgets/logout_confirmation_bottom_sheet.dart
```diff
@@ -0,0 +1,96 @@
...
+                      onPressed: () => Navigator.of(context).pop(),
+                      bgColor: context.colour.secondaryFixed,
+                      textColor: context.colour.secondary,
+                    ),
+                  ),
+                  const SizedBox(width: 12),
+                  Expanded(
+                    child: BBButton.small(
+                      label: 'Logout',
+                      onPressed: () async {
```
kumulynja, line 80: Maybe better to first try to logout and only pop when successful.
And catch any errors that can be shown in the dialog if not.

lib/features/exchange/presentation/exchange_cubit.dart
```diff
@@ -63,21 +73,64 @@ class ExchangeCubit extends Cubit<ExchangeState> {
...
+
+    try {
+      emit(state.copyWith(isSaving: true));
+
+      await _saveUserPreferencesUsecase.execute(
+        language: state.selectedLanguage,
+        currency: state.selectedCurrency,
+      );
+
+      emit(state.copyWith(isSaving: false));
```
kumulynja, line 95: Since both the try as the catch should set the isSaving to false at the end of it, it is better to not repeat this code in both, but just one time add finally after the catch clause:
```
finally {
    emit(state.copyWith(isSaving: false));
}
```

lib/features/exchange/presentation/exchange_cubit.dart
```diff
@@ -63,21 +73,64 @@ class ExchangeCubit extends Cubit<ExchangeState> {
...
 
       // Clear any previous exceptions
       emit(state.copyWith(deleteApiKeyException: null));
-
       await _deleteExchangeApiKeyUsecase.execute();
-      log.fine('Successfully logged out from exchange');
 
-      // Now that the user should be logged out, try to fetch the user summary
-      // again, which should result in an ApiKeyException now.
-      await fetchUserSummary();
```
kumulynja, line 78: Are you sure removing this doesn't break any navigation behavior?
Since some things, like redirects are done based on the ApiKeyException to make sure the user is redirected to the login screen.
I should test it myself to be sure though. But from the comment I put, it seems it was there for a reason and if this is removed we should probably account for it in another way.

kumulynja: If the flag `manualLogout` in some way takes care of this, I don't really like that approach then. Now, the ApiKeyException is a common flag to know the user isn't logged in, with `manualLogout` it become two things to check and to reset when needed. It makes it more error prone and it also feels like `manual logout` is an action, not a state that should be stored in the state. Not being logged in is a state and that is flagged by the ApiKeyException for now, which is maybe not the most clear or optimal and maybe we should indeed use a specific field instead of this exception, but adding another flag and certainly one only used in a specific case doesn't make it simpler or clearer imo.

kumulynja: So for now, I prefer to remove the `manualLogout` field and keep the fetchUserSummary to make sure the user is properly logged out.

lib/features/exchange/ui/exchange_router.dart
```diff
@@ -36,7 +35,9 @@ class ExchangeRouter {
           child: BlocListener<ExchangeCubit, ExchangeState>(
             listenWhen:
                 (previous, current) =>
-                    !previous.isApiKeyInvalid && current.isApiKeyInvalid,
+                    !previous.notLoggedIn &&
+                    current.notLoggedIn &&
+                    current.manualLogout,
```
kumulynja: This is an example of what I mean by needing more checks now, instead of just checking `notLoggedIn`, we also have to not forget to check the `manualLogout` flag... I think that's not an improvement.

The `notLoggedIn` is good though, but I would get rid of the `manualLogout` field which can reduce bugs if forgotten.

lib/features/settings/ui/screens/exchange/account_info_screen.dart
```diff
@@ -1,14 +1,56 @@
...
+          ),
+        ),
+        body: const Center(child: CircularProgressIndicator()),
+      );
+    }
+
+    if (userSummary == null) {
+      return Scaffold(
+        backgroundColor: context.colour.secondaryFixed,
+        appBar: AppBar(
```
kumulynja: This can be replaced with just:

```
AppBar(title: 'Account information',),
```

No need to overwrite the leading icon if it should pop anyways, which is the default behavior. This `TopBar`widget also should be deprecated, it's another custom widget that doesn't add any value and does more harm than good since we don't get default behavior out of the box if we use it.

lib/features/settings/ui/screens/exchange/app_settings_screen.dart
```diff
@@ -1,100 +1,190 @@
...
-                selectedLanguage,
-                ['English', 'French', 'Spanish', 'German'],
-                (value) {
-                  setState(() {
-                    selectedLanguage = value!;
-                  });
-                },
+          return Scaffold(
+            backgroundColor: context.colour.secondaryFixed,
+            appBar: AppBar(
```
kumulynja: Same comment as above, verbose AppBar with non-standard approach, the use of `TopBar`should be deprecated.

lib/features/settings/ui/screens/exchange/app_settings_screen.dart
```diff
@@ -1,100 +1,190 @@
...
+            elevation: 4,
+            margin: const EdgeInsets.only(bottom: 100, left: 40, right: 40),
+            padding: const EdgeInsets.symmetric(horizontal: 20, vertical: 12),
+            shape: RoundedRectangleBorder(
+              borderRadius: BorderRadius.circular(20),
+            ),
+          ),
+        );
+      },
+      child: Builder(
```
kumulynja: Why is a builder needed here?
Why not just move the `context.select` statements to above the return statement in the `build` function?

kumulynja (changes requested):

lib/core/exchange/data/models/user_preference_payload_model.dart
i5hi: because the json data needs to have specific ALL_CAPS keys

i5hi: done

lib/features/settings/ui/screens/exchange/app_settings_screen.dart
i5hi: done

lib/features/settings/ui/screens/exchange/account_info_screen.dart
i5hi: done

lib/features/exchange/ui/exchange_router.dart
i5hi: removed

lib/features/settings/ui/screens/exchange/app_settings_screen.dart
i5hi: done

lib/features/exchange/presentation/exchange_cubit.dart
i5hi: done

PR #1019, opened 2025-07-24
Buy amount UI improvements

kumulynja:
This PR mainly updates the amount input component to be visually more equal to the web exchange and to be able to see the exchange rate and input an amount in bitcoin/sats as well.

It does some other minor refactors for the exchange and amount formatting like:

- Changing the currency in user summary to nullable, since a user that just signed up doesn't seem to have a currency yet and this was throwing an error.
- Show a 0 balance for a newly created user without balances and without a set currency in the set currency of the wallet
- Adding a parameter to the fiat amount formatting to show the amount with the symbol instead of the code
- Adding more currencies to the Order fiat currencies, since balances are returned for them too.

### DIFF
--- a/lib/core/exchange/data/models/user_summary_model.dart
+++ b/lib/core/exchange/data/models/user_summary_model.dart
@@ -5,7 +5,7 @@ class UserSummaryModel {
   final String email;
   final List<UserBalanceModel> balances;
   final String language;
-  final String currency;
+  final String? currency;
   final UserDcaModel dca;
   final UserAutoBuyModel autoBuy;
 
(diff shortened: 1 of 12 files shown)


i5hi:
Show a 0 balance for a newly created user without balances and without a set currency in the set currency of the wallet

I have addressed this in this PR: 
https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/1018

`displayBalances` in UserSummary

i5hi:
I think this PR is okay; I will rebase mine on top of this.

i5hi (approved):

PR #1020, opened 2025-07-24
Refactor/view backupkey

StaxoLotl:


lib/core/recoverbull/domain/entity/backup_info.dart
```diff
@@ -19,6 +19,10 @@ class BackupInfo {
...
   String get id => HEX.encode(BullBackup.fromJson(backupFile).id);
   DateTime get createdAt => DateTime.fromMillisecondsSinceEpoch(
-        BullBackup.fromJson(backupFile).createdAt,
-      );
+    BullBackup.fromJson(backupFile).createdAt,
+  );
+}
+
+extension ToBackupInfo on String {
+  BackupInfo get backupInfo => BackupInfo(backupFile: this);
```
ethicnology, line 27: I think this extension on string is confusing since you can already do `BackupInfo(backupFile: yourString)`

lib/features/backup_settings/ui/screens/view_backup_key_screen.dart
```diff
@@ -39,11 +56,13 @@ class _Screen extends StatelessWidget {
...
-              previous.error != current.error && current.error != null,
+              previous.error != current.error && current.error != null ||
+              previous.derivedBackupKey != current.derivedBackupKey ||
+              previous.status != current.status,
       listener: (context, state) {
-        ScaffoldMessenger.of(
-          context,
-        ).showSnackBar(SnackBar(content: Text(state.error.toString())));
+        if (state.error != null) {
+          log.severe('Failed to derive backup key: ${state.error}');
```
ethicnology, line 64: Does all the conditions stated in `listenWhen` are related to backupKey derivation ? 

Are we sure they do not log any sensitive information ?

lib/features/backup_settings/ui/widgets/backup_key_warning.dart
```diff
@@ -69,14 +69,14 @@ class BackupKeyWarningBottomSheet extends StatelessWidget {
                     ),
                     const Gap(24),
                     BBText(
-                      'Never save the backup key on the same device or in the same location as your backup file. Store them separately and securely.',
+                      'It is critically important that you do not save the backup key at the same place where you save your backup file. Always store them on separate devices or separate cloud providers.',
```
ethicnology, line 72: The backup file is encrypted so it can be stored in the cloud, but the backup key is not, it should be stored only in a password manager.

lib/features/backup_settings/ui/widgets/backup_key_warning.dart
```diff
@@ -69,14 +69,14 @@ class BackupKeyWarningBottomSheet extends StatelessWidget {
...
                       maxLines: 4,
                       style: context.font.bodyMedium,
                     ),
                     const Gap(16),
                     BBText(
-                      'Only download the backup key if you understand the security implications and have a secure storage plan.',
-                      maxLines: 4,
+                      'For example, if you used Google Drive for your backup file, do not use Google Drive for your backup key.',
```
ethicnology, line 78: The backup file is encrypted so it can be stored in the cloud, but the backup key is not, it should be stored only in a password manager.

lib/features/backup_settings/ui/widgets/backup_key_warning.dart
ethicnology: @BullishNode what do you think?

lib/core/recoverbull/domain/entity/backup_info.dart
StaxoLotl: we are actually converting it back and forth in multiple places that's why I have made such an extension

lib/features/backup_settings/ui/screens/view_backup_key_screen.dart
StaxoLotl: Yeah, all the state variables gets updated in case of an error, in the backupKey derivation

lib/features/backup_settings/ui/widgets/backup_key_warning.dart
i5hi: Any reason why we changed this? The original message seems fine. Also feels more to the point.

lib/features/backup_settings/ui/widgets/backup_key_warning.dart
i5hi: even this; i feel the original message is better

ethicnology: Me too

lib/features/backup_settings/ui/widgets/backup_key_warning.dart
ethicnology: Agreed original message is better.

The minimum recommendation we can provide is a password manager to avoid plain text backup key to be stored in a cloud or as a file.

PR #1023, opened 2025-07-28
refactor: increase top buttons size to ease tapping

ethicnology:
The size of icons are too smalls, not easy to tap them on real device
This PR slightly increase there size

before
simulator_screenshot_5AFE825D-9ABD-48E2-BCA6-AC4B58DD6B4A

after
simulator_screenshot_BD4C42F7-D4FA-4EF2-B4D8-5B0B631A1DBA


i5hi:
ACK

PR #1024, opened 2025-07-28
fix: duplicates addresses in `WalletAddressHistory` 

ethicnology:
- Replace `WalletAddressHistory` by `WalletAddresses`
- Update Schema v4
- Update migration v3 to v4
- Add unit test to check the existence and emptiness of `WalletAddresses` table


ethicnology:
The migration may be improved if we can replace the `customStatement` by a `columnTransformer`

tracked: https://github.com/simolus3/drift/issues/3620

i5hi (commented):
A test to try make a duplicate and see the error it gives will be great. this will help us make sure we handle that case.

ethicnology:
Also replaced hardcoded SQL with Drift statements in the migration, which can serve as an example for future migrations.

PR #1025, opened 2025-07-29
WIP: Sell feature

kumulynja:
Progress:
- [x] Routes enum, shellroute and routes with navigation listeners
- [x] BloC setup with a different SellState subtype for every 'step', so that it is more clear which data is expected in every step and avoid bugs from missing or old values
- [x] Added State transition extensions for convenience and less verbose code in the BLoC
- [x] Used a Form for better field validation in the Amount screen, while keeping the input field widgets Stateless. (When time, we should do this for Buy as well and be able to reuse the amount input widgets between both)
- [x] Lightning Network support for external wallet (data/domain layer adaption needed @i5hi)
- [x] Fee and Advanced Settings to send funds to sell from internal wallet
- [x] Show errors like insufficient balance or other Send failures on the send payment screen
- [x] The screen to receive the payment from the external wallet is not in the Figma and the current screen for it should be adapted and use real data still
- [x] Polling the order on the receive payment from external wallet screen to navigate to in-progress when received
- [x] Polling the order in the in-progress screen to navigate to success when it’s completed
- [x] Make sure Sell orders are properly shown in the tx overview and the details are correct as well

### DIFF
--- a/lib/features/sell/presentation/bloc/sell_bloc.dart
+++ b/lib/features/sell/presentation/bloc/sell_bloc.dart
@@ -534,7 +534,7 @@ class SellBloc extends Bloc<SellEvent, SellState> {
 
   void _startPolling() {
     _pollingTimer?.cancel();
-    _pollingTimer = Timer.periodic(const Duration(seconds: 10), (timer) {
+    _pollingTimer = Timer.periodic(const Duration(seconds: 5), (timer) {
       add(const SellEvent.pollOrderStatus());
     });
   }


lib/core/fees/domain/get_network_fees_usecase.dart
```diff
@@ -24,7 +25,8 @@ class GetNetworkFeesUsecase {
 
       return await _feesRepository.getNetworkFees(network: network);
     } catch (e) {
-      throw GetNetworkFeesException('$e');
+      log.severe('Network fees: $e');
```
ethicnology, line 28: This will be written in log files, is there any sensitive information ?

lib/core/wallet/domain/usecases/get_address_at_index_usecase.dart
```diff
@@ -0,0 +1,32 @@
...
+      throw GetAddressAtIndexException(e.toString());
+    }
+  }
+}
+
+class GetAddressAtIndexException implements Exception {
+  final String? message;
+
+  GetAddressAtIndexException(this.message);
+}
```
ethicnology, line 32: We are likely to display these errors using `.toString` and I've noticed accross the codebase that we will end up with something like `Instance of GetAddressAtIndexException` if we want to avoid that we should override the `toString` method to return the message.

Actually we may have a BullException that does this once and we would all extends it with no need to override it each time

lib/features/sell/presentation/bloc/sell_bloc.dart
ethicnology, line 1: Better using a cubit

lib/features/sell/presentation/bloc/sell_bloc.dart
```diff
@@ -0,0 +1,664 @@
...
+
+    OrderAmount orderAmount;
+    if (event.isFiatCurrencyInput) {
+      orderAmount = FiatAmount(double.parse(event.amountInput));
+    } else {
+      final amountBtc =
+          amountInputState.bitcoinUnit == BitcoinUnit.sats
+              ? double.parse(event.amountInput)
+              : int.parse(event.amountInput) / 1e8;
+      orderAmount = BitcoinAmount(amountBtc);
```
ethicnology, line 152: How sensitive is that ?

Shouldn't we have an utils that does the conversion from bitcoin to sats in a single place ?

lib/features/sell/presentation/bloc/sell_state.dart
```diff
@@ -0,0 +1,331 @@
...
+sealed class SellState with _$SellState {
+  const factory SellState.initial({
+    ApiKeyException? apiKeyException,
+    GetExchangeUserSummaryException? getUserSummaryException,
+  }) = SellInitialState;
+  const factory SellState.amountInput({
+    required UserSummary userSummary,
+    required BitcoinUnit bitcoinUnit,
+  }) = SellAmountInputState;
+  const factory SellState.walletSelection({
```
ethicnology, line 13: A blank line between each factory/function make it more readable (for the future)

lib/features/sell/presentation/bloc/sell_state.dart
```diff
@@ -0,0 +1,331 @@
...
+    final order = sellOrder;
+    String invoiceString = '';
+
+    switch (order.payinMethod) {
+      case OrderPaymentMethod.bitcoin:
+        if (order.bitcoinAddress != null) {
+          final amountBtc = order.payinAmount;
+          final address = order.bitcoinAddress!;
+
+          final bip21Uri = Uri(
```
ethicnology: This should use `Bip21Uri` constructor from `bip21_uri` package

lib/features/sell/presentation/bloc/sell_state.dart
```diff
@@ -0,0 +1,331 @@
...
+            queryParameters: {'amount': amountBtc.toString()},
+          );
+          invoiceString = bip21Uri.toString();
+        }
+      case OrderPaymentMethod.liquid:
+        if (order.liquidAddress != null) {
+          final amountBtc = order.payinAmount;
+          final address = order.liquidAddress!;
+
+          final bip21Uri = Uri(
```
ethicnology: This should use `Bip21Uri` constructor from `bip21_uri` package

lib/core/fees/domain/get_network_fees_usecase.dart
i5hi: nothing sensitive

lib/features/sell/presentation/bloc/sell_bloc.dart
i5hi: yes we have a util; i will use that instead

lib/features/sell/presentation/bloc/sell_state.dart
i5hi: done

lib/features/sell/presentation/bloc/sell_state.dart
ethicnology: https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/1025/commits/6bcb4946a7498fc352f49b436793977d5ac6fbc2

PR #1028, opened 2025-07-29
feat: Support multiple wallets from mnemonics of various lengths and derivation purposes

ethicnology:
This PR introduces a new `MnemonicWidget` that **replaces**, **simplifies**, and **enhances** the old `SeedWordsGrid`.

---

### 🆕 Features

- ✅ Word-level validation  
- ✅ Full mnemonic + passphrase validation  
- 💪 Supports all mnemonic lengths specified in BIP-39  
- 🧼  Clean mnemonic errors for end-users
- 🚀  Faster onboarding removing the `recover-success` route/screen which is not necessary since the mnemonic is now validated
- 🤑 Displays wallet status (`balance.confirmed`) and transactions for each derivation path

In the onboarding flow, `MnemonicWidget` is configured to **restrict passphrase, label, and lengths >12** for the default wallet recovery. These restrictions are toggleable, so enabling advanced recovery is just a matter of flipping parameters.

---

### ⚠️ Temporary Code

- This branch includes a temporary `TheDirtyUsecase`, which is a quick-and-dirty implementation (as we agreed with @i5hi) to get the `Show Derivation Status` feature in before today's code freeze. **This use case will need to be refactored properly to align with our architecture and internal guidelines.**

---

### 🎥 Demo

Video showcasing the new widget with all features enable:
https://github.com/user-attachments/assets/a8650661-35a3-4750-b65e-3bf34e8dd282


ethicnology:
https://github.com/user-attachments/assets/a8650661-35a3-4750-b65e-3bf34e8dd282

lib/features/import_mnemonic/router.dart
```diff
@@ -0,0 +1,60 @@
...
+import 'package:bb_mobile/features/import_mnemonic/ui/mnemonic_page.dart';
+import 'package:bb_mobile/features/import_mnemonic/ui/select_purpose_page.dart';
+import 'package:bb_mobile/features/wallet/ui/wallet_router.dart';
+import 'package:bb_mobile/locator.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:go_router/go_router.dart';
+
+enum ImportMnemonicRoute {
+  importMnemonicHome('/import-mnemonic-home'),
+  selectScriptType('/select-script-type');
```
kumulynja, line 12: Nothing critical, just wondering if select-script-type should be a root level route or not?
It is only ever seen after importing a mnemonic, right?

lib/features/import_mnemonic/ui/mnemonic_page.dart
```diff
@@ -0,0 +1,62 @@
...
+                content: BBText(
+                  state.error!.toString(),
+                  style: Theme.of(context).textTheme.bodyMedium,
+                  color: Theme.of(context).colorScheme.onError,
+                ),
+              ),
+            );
+          }
+        },
+        builder: (context, state) {
```
kumulynja: I think the builder here is not needed, since the `state` is not used. So replace the `BlocConsumer` with just a `BlocListener`.

lib/features/import_mnemonic/ui/select_purpose_page.dart
```diff
@@ -0,0 +1,197 @@
...
+                  state.error!.toString(),
+                  style: Theme.of(context).textTheme.bodyMedium,
+                  color: Theme.of(context).colorScheme.onError,
+                ),
+              ),
+            );
+          }
+        },
+        builder: (context, state) {
+          final cubit = context.read<ImportMnemonicCubit>();
```
kumulynja, line 44: If you already have the `state` from the builder params of the `BlocConsumer`, you don't have to get it from the Cubit in the context anymore.
So you can replace:

```
final cubit = context.read<ImportMnemonicCubit>();
final scriptType = cubit.state.scriptType;
```

with just `final scriptType = state.scriptType;`.

lib/features/import_mnemonic/router.dart
ethicnology: They are both subroutes of the ShellRouter

lib/features/import_mnemonic/ui/mnemonic_page.dart
ethicnology: 833b8820503a26a7ce982cbcb7e4bbb4fc3f5a7d

lib/features/import_mnemonic/ui/select_purpose_page.dart
ethicnology: 833b8820503a26a7ce982cbcb7e4bbb4fc3f5a7d

PR #1032, opened 2025-07-31
Exchange withdraw

kumulynja:
As good as finished, just MISSING SPECS about the payment processor codes and how they match with the recipients' payment processor values, they are not the same...
And also missing the specs about which info to show exactly in the recipients' card and on the confirmation screen based on the available information, which depends on the recipient type and payment processor used.


lib/features/withdraw/ui/widgets/new_recipient_form.dart
```diff
@@ -29,27 +31,124 @@ class _NewRecipientFormState extends State<NewRecipientForm> {
...
+      case FiatCurrency.cad:
+        return 'CA';
+      case FiatCurrency.eur:
+        return 'EU';
+      case FiatCurrency.mxn:
+        return 'MX';
+      case FiatCurrency.crc:
+        return 'CR';
+      case FiatCurrency.usd:
+        return 'CR'; // Default fallback
```
ethicnology, line 47: maybe we can add this value to the enum and access it `FiatCurrency.eur.code`

ethicnology (commented):
Improvement suggested

i5hi:
Basically working - all pages and endpoints are in; some bugs exist in certain currencies but we will address them in separate issues during the testing week.

PR #1048, opened 2025-08-04
Recoverbull content fix

i5hi:
Added maxlines: 5 to avoid content being cut


i5hi:
Also added Wallet Delete redirect home to this PR by accident.

ethicnology (approved):
Better to delete commented code rather than keeping it, because it's more likely to be deleted at some point

PR #1062, opened 2025-08-04
feat: `allowAutoFillWords` option to auto-fill mnemonic words when only one suggestion remains

ethicnology:


ethicnology:
https://github.com/user-attachments/assets/393989a1-2429-4454-b404-de7904a7a31e

PR #1068, opened 2025-08-05
Exchange auth close error

i5hi:
Use messages to make `x` in webview reroute back to wallet home. 

The reason for routing to wallet rather than exchange home is that exchange home can lead to looping back to the auth page. If auth is successful we anyway route to exchange.


kumulynja (approved):
Tested and works 👍

PR #1079, opened 2025-08-06
Better handling of broadcast errors. 

i5hi:
Made a common broadcast error for liquid and bitcoin.
Simple error message on broadcast fail.
Removed redundant call to getWalletUsecase in confirmTransaction - already called in broadcastTransaction

### DIFF
--- a/lib/features/send/presentation/bloc/send_cubit.dart
+++ b/lib/features/send/presentation/bloc/send_cubit.dart
@@ -2,6 +2,8 @@ import 'dart:async';
 
 import 'package:bb_mobile/core/blockchain/domain/usecases/broadcast_bitcoin_transaction_usecase.dart';
 import 'package:bb_mobile/core/blockchain/domain/usecases/broadcast_liquid_transaction_usecase.dart';
+import 'package:bb_mobile/core/errors/send_errors.dart'
+    show BroadcastTransactionException;
 import 'package:bb_mobile/core/exchange/domain/usecases/convert_sats_to_currency_amount_usecase.dart';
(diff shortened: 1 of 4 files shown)


lib/features/send/presentation/bloc/send_cubit.dart
```diff
@@ -1253,6 +1255,22 @@ class SendCubit extends Cubit<SendState> {
...
+        state.copyWith(
+          confirmTransactionException: ConfirmTransactionException(e.message),
+          broadcastingTransaction: false,
+        ),
+      );
+    } on BroadcastTransactionException catch (_) {
+      emit(
+        state.copyWith(
+          confirmTransactionException: ConfirmTransactionException(
+            'Failed to broadcast transaction. Check your network connection and try again.',
```
kumulynja, line 1269: Ok for now, but this doesn't permit localization of the error message for the user if we want that in the future.
Better to not make everything a ConfirmTransactionException here, but keep the concrete type and then in the UI have a switch for all concrete types and show a specific error message for it in the language of the user.

This would also require structuring the send_error file differently. We should have one sealed SendError type and all concrete types like BroadcastTransactionException should be a subtype of SendError. Like that we can just keep one error per domain in the cubit state and check it in the UI to show the correct error.
For example the GetWalletException would be a subtype of a WalletError that we can have as a field in the state, and BroadcastTransactionException can be added to a SendError type field of the state here. And then the UI can just show/hide the correct message based on the value.
Here we can then have `on WalletError catch(e)` and `on SendError catch(e)`, per domain that is used basically.
But yeah, we probably have to define and apply our whole error handling approach everywhere still at some point. For now this is ok.

PR #1082, opened 2025-08-06
feat: add logs viewer screen and integrate with share logs widget

ethicnology:
- Scrollable in two directions
- `SelectableText`
- Copy button for the whole line
⚠️ Only for superuser and non `INFO` logs


ethicnology:
demo
https://github.com/user-attachments/assets/afd7c4c7-d3d3-4c36-aa9b-337c6725f060

PR #1085, opened 2025-08-06
hide external wallets/signers and watch-only wallets to sell from

kumulynja:
Hide until we added selling from these types properly.

### DIFF
--- a/lib/features/wallet/ui/widgets/wallet_cards.dart
+++ b/lib/features/wallet/ui/widgets/wallet_cards.dart
@@ -7,9 +7,15 @@ import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:gap/gap.dart';
 
 class WalletCards extends StatelessWidget {
-  const WalletCards({super.key, this.padding, this.onTap});
+  const WalletCards({
+    super.key,
+    this.padding,
+    this.onTap,
+    this.hideWatchOnly = false,
+  });
 
   final EdgeInsetsGeometry? padding;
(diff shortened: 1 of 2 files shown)


lib/features/wallet/ui/widgets/wallet_cards.dart
```diff
@@ -30,7 +36,14 @@ class WalletCards extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final wallets = context.select((WalletBloc bloc) => bloc.state.wallets);
+    final wallets = context.select(
+      (WalletBloc bloc) =>
+          bloc.state.wallets
+              .where(
+                (w) => !hideWatchOnly || (!w.isWatchOnly && !w.signsRemotely),
```
ethicnology: You can replace `(!w.isWatchOnly && !w.signsRemotely)` by `w.signsLocally`

ethicnology (approved):

lib/features/wallet/ui/widgets/wallet_cards.dart
i5hi: maybe even rename `hideWatchOnly` to `localSignersOnly` ?

kumulynja: Yes, that's more accurate, will change it. Thanks!

kumulynja: Changed!

PR #1086, opened 2025-08-06
Remove coldcard from super user + minor UI edits

i5hi:


lib/features/settings/ui/screens/bitcoin/experimental_settings_screen.dart
ethicnology, line 1: Let's rename the screen as well ?

kumulynja (approved):

PR #1106, opened 2025-08-09
Remove esp

i5hi:


ethicnology (approved):
The version bump is already committed in other PR

PR #1107, opened 2025-08-10
Secure storage view

i5hi:


ios/Runner.xcodeproj/project.pbxproj
ethicnology, line 1: this is already committed in #1110 which may result in conflicts

ethicnology:
Analyze and Build fails with

```
error • Target of URI doesn't exist: 'package:bb_mobile/core/storage/secure_storage_backup.dart' • lib/main.dart:5:8 • uri_does_not_exist
  error • Undefined name 'SecureStorageBackup' • lib/main.dart:44:13 • undefined_identifier
```

PR #1110, opened 2025-08-11
KYC Webview Infinite reload fix

i5hi:
Increased the delay to check for refreshRequired from 5s to 30s


lib/features/exchange/ui/screens/exchange_kyc_screen.dart
```diff
@@ -113,7 +113,7 @@ class _ExchangeKycScreenState extends State<ExchangeKycScreen> {
             //  Flutter web app has been rendered successfully.
             // If it is still 0, it means that the Flutter web app has not been
             //  rendered correctly, and we should reload the WebView.
-            await Future.delayed(const Duration(seconds: 5));
+            await Future.delayed(const Duration(seconds: 30));
```
kumulynja, line 116: This will only delay the refresh, but not stop it.
PR #1112 solves this. So this should just stay at 5 seconds so in case the refresh needs to happen, people don't have to wait 30 seconds.

Issue #1105, opened 2025-08-09
Custom electrum server form requires string for testnet to be able to save 


BullishNode:
## Description

Custom electrum server form requires string for testnet to be able to save 

### Expected behavior
Same an electrum server without requiring testnet and use default testnet

ethicnology:
1. Make `testnet` non mandatory
2. Fill the `testnet` with default value

ethicnology:
1. The whole `ElectrumSettings` feature should be refactored since the code is highly complex. Many things doesn't works as expected.
2. We can't fill testnet with default values since we don't have default value for testnet.

Example of unexpected things:
- User should have hints about the URL format expected for example the specification of `ssl://` and the port `:51002`
- Connection should be tested before saving
- The app needs to be restarted to fetch the balances…

PR #1113, opened 2025-08-11
feat: Saving `Mainnet` custom electrum server does not require a custom `Testnet`

ethicnology:


ethicnology:
[comment](https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1105#issuecomment-3174961530)
> 1. The whole `ElectrumSettings` feature should be refactored since the code is highly complex. Many things doesn't works as expected.
> 2. We can't fill testnet with default values since we don't have default value for testnet.

> Example of unexpected things:
> - User should have hints about the URL format expected for example the specification of `ssl://` and the port `:51002`
> - Connection should be tested before saving
> - The app needs to be restarted to fetch the balances…

Issue #1109, opened 2025-08-11
Exchange: Handle redirection UX on specific errors


i5hi:
- [ ] If a user does not have KYC - on all flows that require KYC (Funding, Buy, [Sell - CA atleast Limited with Limits]) provide a link to `Complete KYC`
- [ ] If a user does not have enough balance for a BUY order provide a link to `Fund Your Account`

PR #1115, opened 2025-08-11
Exchange guide to kyc and funding

### LINKED ISSUE #1111: Error Box Styling
## Description
The red bar on the left of the error box does not cover the entire height of the error box.
This is likely to be an issue with this component and should be fixed across its usage.

### Reproduce 
Turn off the internet and try fund your account

### Expected behavior
The red bar should cover the height of the box.

## Context
- `BULL`: 5.4.0+77` 

### Platforms
- [x] Android
- [x] iOS

### Demo

Image

kumulynja:
Adds "fund your account" and "complete kyc" warnings and buttons as requested in #1109
Also fixes #1111 .


kumulynja:
@i5hi For funding and buy the "Complete KYC" button is shown for anyone that is not fully verified.
For sell as well, except if the currency is CAD, then I permitted limited KYC as well like this:

```
// For CAD, we allow selling with limited KYC.
// For other fiat currencies, we require full KYC.
final isKycLevelOk =
    isFullyVerifiedKycLevel ||
    fiatCurrency == FiatCurrency.cad && isLimitedKyc;
```

Is that correct?

i5hi:
> @i5hi For funding and buy the "Complete KYC" button is shown for anyone that is not fully verified. For sell as well, except if the currency is CAD, then I permitted limited KYC as well like this:
> 
> ```
> // For CAD, we allow selling with limited KYC.
> // For other fiat currencies, we require full KYC.
> final isKycLevelOk =
>     isFullyVerifiedKycLevel ||
>     fiatCurrency == FiatCurrency.cad && isLimitedKyc;
> ```
> 
> That's how I interpreted "[Sell - CA atleast Limited with Limits]" in #1109. Is that correct?

With CA this is a fine way to start but once the user attempts to complete an order if they are beyond the limits of their current KYC level then based on that error we have to show them a `Upgrade KYC Level` button to move to Light or Full verification.

PR #1120, opened 2025-08-12
permit navigation to incodesmile.com for exchange verification

kumulynja:
- [x] when clicking the link
- [ ] automatic pop-up


i5hi:
- [ ] do not fetch api key until user completed verification

PR #1124, opened 2025-08-13
Kyc allow camera

i5hi:


lib/features/exchange/ui/screens/exchange_kyc_screen.dart
```diff
@@ -76,17 +82,15 @@ class _ExchangeKycScreenState extends State<ExchangeKycScreen> {
...
-                isLogin ||
-                isBBLogo ||
-                isEmailVerification ||
-                isIncodeSmile;
+            final allow = isKyc || isLogin || isBBLogo || isEmailVerification;
 
             if (allow) {
               return NavigationDecision.navigate;
             } else {
               log.warning('Navigation blocked: ${url.path}');
```
ethicnology, line 90: Is navigation still blocked?

kumulynja: We block navigation intentionally here depending on the url to keep the user only on the allowed routes of the kyc flow.

PR #1159, opened 2025-08-28
Added account deletion and removed nfc for standard user

i5hi:


ethicnology (changes requested):
Build CI broken ⚠️ 

Missing file `delete_account_bottom_sheet.dart`
 
```
Error (Xcode): lib/features/settings/ui/screens/exchange/exchange_settings_screen.dart:1:8: Error: Error when reading 'lib/core/widgets/delete_account_bottom_sheet.dart': No such file or directory
```

PR #1176, opened 2025-09-02
feat: recoverbull list drive backups and check the balance and transactions status before import

ethicnology:


ethicnology:
- ⚠️ wallets are not loaded at the end of the flow
- liquid wallet check is missing

PR #1181, opened 2025-09-03
default new receive address for liquid to last unused

kumulynja:
Because of new address generation in swap resolutions, the address stop gap can become exceeded quickly, to already prevent it on Liquid, in which address reuse is less of a concern, we return the last unused as default now by keeping the `forceUnseenLiquidAddress` parameter on false in the `getNewReceiveAddress` function. And only set that param to true if the user explicitly wants to generate another address.

This is a quick fix though, since it is not good to have a param only for Liquid in a function for both bitcoin as liquid wallets.
Also does this not solve the problem for Bitcoin on-chain wallets exceeding the stop gap quickly. Therefore a more thorough solution is needed in the swaps management to only generate one refund address per swap and not more after every reinit of the swap watcher.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -103,9 +103,9 @@ PODS:
   - payjoin_flutter (0.20.0)
   - PromisesObjC (2.4.0)
   - ScreenProtectorKit (1.3.1)
-  - SDWebImage (5.21.1):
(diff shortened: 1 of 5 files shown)


i5hi:
I will make the fix in swap watcher to save the address and use the saved address if it exists.

PR #1195, opened 2025-09-09
feat: allow autoswap to any bitcoin walletId

i5hi:
Migration for `recipientWalletId` in autoswap is already completed on develop.

This PR just uses this new field and updates existing functionality to use a provided wallet ID rather than using the default wallet all the time. 

Since the value is set as null in the db; the code will now check for null and replace with defaultBitcoinWalletId; following which this value should never be set to null again. This will ensure existing behaviour does not change.

Users can then change the value in autoswap settings where providing a value for recipientWalletId is required.

### DIFF
--- a/lib/core/swaps/domain/usecases/auto_swap_execution_usecase.dart
+++ b/lib/core/swaps/domain/usecases/auto_swap_execution_usecase.dart
@@ -48,6 +48,8 @@ class AutoSwapExecutionUsecase {
   }) async {
     final swapRepository = isTestnet ? _testnetRepository : _mainnetRepository;
     final autoSwapSettings = await swapRepository.getAutoSwapParams();
+    // check if recipient wallet id is set
+
     if (!autoSwapSettings.enabled) {
       throw AutoSwapDisabledException('Auto swap is disabled');
     }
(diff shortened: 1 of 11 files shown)


lib/core/swaps/domain/usecases/auto_swap_execution_usecase.dart
```diff
@@ -154,28 +178,27 @@ class AutoSwapExecutionUsecase {
...
+    // sometimes sync fails and label is not set
+    final txLabel = Label.tx(
+      transactionId: txid,
+      walletId: defaultLiquidWallet.id,
+      label: 'Auto-Swap',
+    );
+    await _labelRepository.store(txLabel);
 
-    final walletTx = await _walletTxRepository.getWalletTransaction(
+    // sync at the end to ensure label is set
```
kumulynja, line 196: Not sure how the labeling is impacted by sync?

lib/features/autoswap/ui/widgets/autoswap_settings.dart
```diff
@@ -402,32 +404,156 @@ class _AlwaysBlockToggle extends StatelessWidget {
...
+                        ? context.colour.error
+                        : context.colour.primary,
+              ),
+            ),
+            contentPadding: const EdgeInsets.symmetric(
+              horizontal: 16,
+              vertical: 12,
+            ),
+          ),
+          hint: BBText(
```
kumulynja, line 481: Not a problem here, but better not to use BBText anymore.

kumulynja (changes requested):
@i5hi For the recipientWalletId to be stored in the autoswap settings, also the boltz_storage_datasource should be updated to pass this field to the insert in the `storeAutoSwapSettings` function, otherwise it will not be saved.

lib/core/swaps/domain/usecases/auto_swap_execution_usecase.dart
i5hi: its not; its that earlier it was before labelling and sometimes it would fail and the label would not be added

PR #1200, opened 2025-09-12
Pay fixes

i5hi:
- [x] Start flow with recipient selection
- [x] Sinpe movil recipient creation (new checkSinpe endpoint to validate number to ownerName) - 
- [x] Sinpe success screen 

Reference number not added since it requires a backend update to the getOrderSummary endpoint

### DIFF
--- a/lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
+++ b/lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
@@ -167,7 +167,6 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
     required String apiKey,
     required String orderId,
   }) async {
-    log.info('Confirm order request: $orderId');
     final resp = await _http.post(
       _ordersPath,
       data: {
@@ -178,7 +177,6 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
       },
       options: Options(headers: {'X-API-Key': apiKey}),
     );
-    log.info('Confirm order response: ${resp.data}');
     if (resp.statusCode != 200) throw Exception('Failed to confirm order');
     return OrderModel.fromJson(resp.data['result'] as Map<String, dynamic>);
   }
@@ -187,9 +185,6 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
     required String apiKey,
     required String orderId,
   }) async {
-    log.info('getOrderSummary request - orderId: $orderId');
-    log.info('getOrderSummary request - apiKey: $apiKey');
-
     final resp = await _http.post(
       _ordersPath,
       data: {
@@ -200,8 +195,6 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
       },
       options: Options(headers: {'X-API-Key': apiKey}),
     );
-
-    log.info('getOrderSummary response: ${resp.data}');
 
     if (resp.statusCode != 200) throw Exception('Failed to get order summary');
     return OrderModel.fromJson(
@@ -219,6 +212,7 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
         'method': 'listOrderSummaries',
         'params': {
           "sortBy": {"id": "createdAt", "sort": "desc"},
+          "paginator": {"page": 1, "pageSize": 50},
         },
       },
       options: Options(headers: {'X-API-Key': apiKey}),
@@ -287,7 +281,6 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
       },
       options: Options(headers: {'X-API-Key': apiKey}),
     );
-    log.info('getFundingDetails: ${resp.data}');
     if (resp.statusCode != 200) {
       throw Exception('Failed to get funding details');
     }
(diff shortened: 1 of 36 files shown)


ethicnology:
rebased on `develop` to propagate https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/1206

lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
```diff
@@ -213,6 +212,7 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
         'method': 'listOrderSummaries',
         'params': {
           "sortBy": {"id": "createdAt", "sort": "desc"},
+          "paginator": {"page": 1, "pageSize": 50},
```
kumulynja, line 215: If we will not do it in this PR, should we create an Issue to use vars instead of hardcoded values here so we can load more orders dynamically?

lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
```diff
@@ -509,7 +496,9 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
         'jsonrpc': '2.0',
         'id': '0',
         'method': 'listRecipients',
-        'params': {},
+        'params': {
+          "paginator": {"page": 1, "pageSize": 50},
```
kumulynja, line 500: Same here, we should probably use vars here and load more recipients dynamically while the user scrolls down through its list of recipients.

lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
```diff
@@ -532,16 +521,16 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
         'jsonrpc': '2.0',
         'id': '0',
         'method': 'listRecipientsFiat',
-        'params': {},
+        'params': {
+          "paginator": {"page": 1, "pageSize": 50},
```
kumulynja, line 525: Same here, create an issue to use pagination dynamically instead of have a hardcoded value here.

lib/features/exchange/ui/screens/exchange_kyc_screen.dart
```diff
@@ -138,19 +130,13 @@ class _ExchangeKycScreenState extends State<ExchangeKycScreen> {
...
 
               if (!isRendered) {
                 log.warning(
                   'Flutter view not successfully rendered. Reloading WebView.',
                 );
                 await _controller.reload();
-              } else {
-                log.info('Flutter view appears to be fully rendered.');
-              }
+              } else {}
```
kumulynja, line 139: No need to keep an empty else body

lib/features/pay/presentation/pay_bloc.dart
```diff
@@ -100,6 +101,10 @@ class PayBloc extends Bloc<PayEvent, PayState> {
     on<PayReplaceByFeeChanged>(_onReplaceByFeeChanged);
     on<PayUtxoSelected>(_onUtxoSelected);
     on<PayLoadUtxos>(_onLoadUtxos);
+    on<PayUpdateOrderStatus>(_onUpdateOrderStatus);
+    on<PayAmountInputBackPressed>(_onAmountInputBackPressed);
```
kumulynja: I don't think we should have BLoC events/handlers for "back button" presses. That should be kept to strictly navigation without side-effects if possible, which I think is possible here.

lib/features/pay/presentation/pay_bloc.dart
```diff
@@ -100,6 +101,10 @@ class PayBloc extends Bloc<PayEvent, PayState> {
...
+    on<PayWalletSelectionBackPressed>(_onWalletSelectionBackPressed);
```
kumulynja: Same here: I don't think we should have BLoC events/handlers for "back button" presses. That should be kept to strictly navigation without side-effects if possible, which I think is possible here.

A clean state should be set before handling the Confirmation/Continue again after having gone back, it shouldn't be done on the back press itself.

lib/features/pay/presentation/pay_state.dart
```diff
@@ -221,54 +224,57 @@ sealed class PayState with _$PayState {
...
-    );
-  }
-}
-
 extension PayAmountInputStateX on PayAmountInputState {
-  PayRecipientInputState toRecipientInputState({
-    required FiatAmount amount,
-    required FiatCurrency currency,
+  PayWalletSelectionState toWalletSelectionState({
+    required Recipient selectedRecipient,
```
kumulynja: If it progresses from the amount input state, then the amount and currency should be passed, not the selected recipient, since the recipient was already selected and passed before when going from the recipient input to amount input state/screen.

lib/features/pay/ui/pay_router.dart
```diff
@@ -13,12 +15,15 @@ import 'package:go_router/go_router.dart';
 
 enum PayRoute {
   pay('/pay'),
-  payRecipients('recipients'),
-  payWalletSelection('wallet-selection'),
-  payExternalWalletNetworkSelection('external-wallet-network-selection'),
-  paySendPayment('send-payment'),
-  payReceivePayment('receive-payment'),
-  paySuccess('success');
+  payRecipients('pay-recipients'),
```
kumulynja: It's no problem either, but I don't think the pay-prefix is really needed since it are all relative paths, so the full path will be /pay/* anyways

lib/features/pay/presentation/pay_state.dart
kumulynja: @i5hi I haven't tested the case to see how it behaves yet, but this feels like a bug

lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
i5hi: i think with recipients its okay to load them all because its a list that grows slower and the filters will work better if we have as many as possible

lib/features/pay/ui/pay_router.dart
i5hi: ahh okay
i will change this

lib/features/pay/presentation/pay_state.dart
i5hi: done

i5hi: this fixed some navigation issues too

PR #1227, opened 2025-09-24
feat: `Ark` integration prototype

ethicnology:
## Features

- The user goes to **Settings** and enables **Ark**.
- Enabling Ark derives a **BIP85 Secret Key** (Hex Application with index **11811**).
- The secret key initializes an **ArkWallet client** to communicate with the operator.
- When the wallet starts, it checks if the **11811 derivation** exists in the database and displays the **Ark Instant Payments** cards.
- Tapping on **Ark Instant Payments** triggers the Ark feature, where the UI/UX tries to reproduce the Bitcoin and Liquid wallets experience.
- The user can:
  - **Receive**: Displays the **boarding** and **offchain** addresses.
    - **Boarding**: A Bitcoin address where funds should be sent/settled to fund the Ark Wallet.
    - **Offchain**: The Ark address.
  - **Send**:
    - Send money to another Ark Wallet by inputting an Ark address.
    - Perform a **Collaborative Redeem** to send Ark Wallet funds back to a Bitcoin address.
  - **Settle**:
    - Commit the pending/unconfirmed VTXOs into committed/confirmed VTXOs.

### Balance & UI Features
- View both **confirmed** and **pending** balance.
- Input only an amount **≤ confirmed balance**.
- Tap **MAX** to send the entire confirmed balance.
- Check the **fiat equivalent** of the Ark balance in available currencies.
- **Hide amounts**.
- **Pull-to-refresh** the screen.

---

## Limitations

- No automatic sync.
- The prototype implements some Ark concepts without proper UX/UI abstractions. The UX should be improved (like auto-settling?).
- Does not display the amount of redeem transactions below dust.
- Does not inform the user about the **dust limit** under which funds cannot be redeemed/settled.
- Ark currently expects a **single `secret_key`** to initialize the wallet, which means:
  - A single Bitcoin boarding address and offchain address → **address reuse**.
  - There are ongoing discussions on how to handle **multiple keypair generation** for Ark Wallet.
- We could create a new derivation path in the **BIP32/SLIP32** fashion.
- We can derive as many **BIP85 derivations** as needed (the complexity is borne by the wallet instead of the package).
- We cannot sort transactions in the same style as the Bitcoin/Liquid wallet since Ark transactions do not necessarily have a date.
- **Better to communicate with Ark server through Tor**


ethicnology:
https://github.com/user-attachments/assets/f1dbad51-63ea-451e-a7ba-e3b85f914833

PR #1243, opened 2025-10-02
feat: delete PIN

ethicnology:
https://github.com/user-attachments/assets/ba43dc91-97ed-4823-a74a-477b118e4d79


i5hi:
We should not show the delete button if the PIN does not exist

Issue #1251, opened 2025-10-02
Amount input bug


BullishNode:
iOS 5.4.5 (fresh install)

Amount input bughttps://github.com/user-attachments/assets/71186fce-e68d-45e6-9a66-68eb3cde74ca

BullishNode:
Video shared on internal slack

PR #1256, opened 2025-10-02
remove unneccessary amount controller 

kumulynja:
An amount controller was not needed here and it was setting the new value after change, but that making the value change again, which triggers the controller again etc. creating some kind of infinite loop when trying to delete parts.


kumulynja:
Should fix: https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1251

i5hi (approved):

PR #1286, opened 2025-10-08
refactor: `DialPad` to improve the UX with recoverbull, PIN, send and receive

ethicnology:
# Before

## BEFORE Create/Confirm/Verify PIN code

https://github.com/user-attachments/assets/a8904252-8564-4cb8-8a0c-d68d37f382f6

## BEFORE Receive

https://github.com/user-attachments/assets/89985b36-5dd4-4d48-b420-2fcfd8da242a

## BEFORE Send

https://github.com/user-attachments/assets/f51d4b08-0abf-49f8-b509-201098c4b39f

## BEFORE Recoverbull

https://github.com/user-attachments/assets/9c6c1ca4-34cc-4aff-b789-600bc3baa47f

# After

## AFTER Create/Confirm/Verify PIN code

https://github.com/user-attachments/assets/ccccd313-d03a-4645-b748-6474ae7d4225

## AFTER Receive

https://github.com/user-attachments/assets/634af45f-c76d-44f2-afe6-5ea8d29c838f

## AFTER Send

https://github.com/user-attachments/assets/c6133c8c-ff1f-4b57-91cb-c2a96d68472f

## AFTER Recoverbull

https://github.com/user-attachments/assets/b80b175c-1b7d-4a48-a050-149a462c3c49

# Remaining issue with send MAX

This need a deeper refactoring of Send cubit that may introduce regressions since we need to compute the fees to be able to update the proper max amount

https://github.com/user-attachments/assets/e605eafc-3a7e-4e3d-ba93-53a30cfdcdb1

# Design flaw in the UX

I've disabled the cursor on the `InputPrice` widget. Having a `DialPad` is a way to offer the same experience across the platforms. The cursor and the actions related to the cursor are specifics for each platforms. Allowing the user to use the cursor create unexpect behaviors with pasting when it comes to synchronize the `InputPrice` with the `DialPad`.

I recommend to use either the platform inputs, removing our `DialPad` in favour of Input of the user platform, or to refactor the `InputPrice` to offer an alternative way to paste an amount.

### DIFF
--- a/lib/core/widgets/dialpad/dial_pad.dart
+++ b/lib/core/widgets/dialpad/dial_pad.dart
@@ -21,59 +21,45 @@ class DialPad extends StatefulWidget {
 }
 
 class _DialPadState extends State<DialPad> {
-  late final TextEditingController controller;
-  bool get hasDot => controller.text.contains('.');
-
-  @override
-  void initState() {
-    super.initState();
-    controller = TextEditingController();
-  }
-
-  @override
-  void dispose() {
-    controller.dispose();
-    super.dispose();
-  }
+  String text = '';
+  bool get hasDot => text.contains('.');
 
   void onTextChanged(String num) {
-    final text = controller.text;
-
     switch (widget.mode) {
       case DialPadMode.int:
         final intValue = int.tryParse(text + num);
         if (intValue == null) break;
-        controller.text = intValue.toString();
+        text = intValue.toString();
       case DialPadMode.double:
         if (hasDot && num == '.') break;
         if (!hasDot && num == '.') {
-          controller.text = text + num;
+          text = text + num;
           break;
         }
         if (!hasDot && num != '.') {
           final intValue = int.tryParse(text + num);
           if (intValue == null) break;
-          controller.text = intValue.toString();
+          text = intValue.toString();
           break;
         }
         if (hasDot) {
           final doubleValue = double.tryParse(text + num);
           if (doubleValue == null) break;
-          controller.text = text + num; // or you wont have 0
+          text = text + num; // or you wont have 0
           break;
         }
       case DialPadMode.pin:
-        controller.text += num;
+        text += num;
     }
 
-    widget.onChanged(controller.text);
+    widget.onChanged(text);
   }
 
   void onBackspacePressed() {
-    if (controller.text.isEmpty) return;
-    controller.text = controller.text.substring(0, controller.text.length - 1);
+    if (text.isEmpty) return;
+    text = text.substring(0, text.length - 1);
 
-    widget.onChanged(controller.text);
+    widget.onChanged(text);
   }
 
   Widget numPadButton(BuildContext context, String num) {


lib/core/widgets/dialpad/dial_pad.dart
```diff
@@ -2,27 +2,88 @@ import 'package:bb_mobile/core/themes/app_theme.dart';
 import 'package:bb_mobile/core/widgets/text/text.dart';
 import 'package:flutter/material.dart';
 
-class DialPad extends StatelessWidget {
+enum DialPadMode { int, double, pin }
+
+class DialPad extends StatefulWidget {
```
kumulynja, line 7: We should keep reusable widgets Stateless if possible, which in this case it is by just passing the TextEditingController as a parameter instead of creating and keeping it in the state of the DialPad itself. It should be kept in the state of the widget that needs the value of the TextEditingController, which is the parent widget that will use the DialPad. A lot of the problems with the current 'reusable' widgets is that they are Stateful  when the state should be managed by the widget that uses them. We should start avoiding those problems.

Update: see comment underneath, a TextEditingController is not even needed to be passed here.

lib/core/widgets/dialpad/dial_pad.dart
```diff
@@ -2,27 +2,88 @@ import 'package:bb_mobile/core/themes/app_theme.dart';
...
-        splashFactory: disableFeedback ? NoSplash.splashFactory : null,
-        highlightColor: disableFeedback ? Colors.transparent : null,
+        onTap: () => onTextChanged(num),
+        splashFactory: widget.disableFeedback ? NoSplash.splashFactory : null,
+        highlightColor: widget.disableFeedback ? Colors.transparent : null,
         child: SizedBox(
           height: 64,
-
           child: Center(
             child: BBText(
```
kumulynja, line 74: Change BBText for regular Text, otherwise the DialPad is not usable in certain layouts.

lib/core/widgets/dialpad/dial_pad.dart
```diff
@@ -2,27 +2,88 @@ import 'package:bb_mobile/core/themes/app_theme.dart';
...
+  final Function(String) onChanged;
   final bool disableFeedback;
+  final DialPadMode mode;
+
+  @override
+  State<DialPad> createState() => _DialPadState();
+}
+
+class _DialPadState extends State<DialPad> {
+  late final TextEditingController controller;
```
kumulynja: Since this widget doesn't have a `TextField` or `TextFormField`, a `TextEditingController` is not needed here. It is being used to keep a value in memory, this could be done by a normal state variable if you really need to, but even that is not needed here I think.

lib/core/widgets/dialpad/dial_pad.dart
ethicnology: It was already a `BBText` it shouldn't create regressions

Which layout are not compatible with `BBText` ? 

Since `BBText` is part of the current UI Kit we should make these layout compatible with it

lib/core/widgets/dialpad/dial_pad.dart
ethicnology: I've replaced the `TextEditingController`

lib/core/widgets/price_input/price_input.dart
```diff
@@ -65,7 +65,7 @@ class PriceInput extends StatelessWidget {
                                 controller: amountController,
                                 focusNode: focusNode,
                                 keyboardType: TextInputType.none,
-                                showCursor: true,
+                                showCursor: false,
```
kumulynja, line 68: This should not be false as this doesn't permit you to add a number anymore at any position you want.

lib/features/receive/ui/screens/receive_amount_screen.dart
```diff
@@ -83,7 +84,19 @@ class _AmountPageState extends State<AmountPage> {
...
           ),
-          ReceiveNumberPad(amountController: _amountController),
+          DialPad(
+            mode: switch (context.watch<ReceiveBloc>().state.bitcoinUnit) {
+              BitcoinUnit.btc => DialPadMode.double,
+              BitcoinUnit.sats => DialPadMode.int,
+              null => DialPadMode.int,
+            },
+            onChanged: (number) {
+              context.read<ReceiveBloc>().add(
```
kumulynja, line 94: This makes it impossible to be compatible with pasting and manually entering numbers. It breaks that you can combine manual and dialpad input.

lib/core/widgets/dialpad/dial_pad.dart
ethicnology: The previous iteration of the DialPad was not allowing to pass a `TextEditingController` and was working with callbacks as well `onBackspacePressed` and `onNumberPressed` which are simplified by `onChanged`

lib/features/send/ui/screens/send_screen.dart
kumulynja, line 1: Removing everything related to MAX here is what re-introduces the bug of escape not working after MAX is pressed anymore.

lib/core/widgets/price_input/price_input.dart
ethicnology: As explained in my PR description

> Design flaw in the UX

> I've disabled the cursor on the InputPrice widget. Having a DialPad is a way to offer the same experience across the platforms. The cursor and the actions related to the cursor are specifics for each platforms. Allowing the user to use the cursor create unexpect behaviors with pasting when it comes to synchronize the InputPrice with the DialPad.

> I recommend to use either the platform inputs, removing our DialPad in favour of Input of the user platform, or to refactor the InputPrice to offer an alternative way to paste an amount.

lib/features/send/ui/screens/send_screen.dart
ethicnology: Yes, we should refactor the way MAX is working in send.

Basically we should either pre-compute the maximum amount (fees deducted) the user can send before he is taping on max or while he is tapping on max.

This way, we will be able to update the state with the proper max sendable amount

lib/features/receive/ui/screens/receive_amount_screen.dart
ethicnology: Related to PR description

> Design flaw in the UX

> I've disabled the cursor on the InputPrice widget. Having a DialPad is a way to offer the same experience across the platforms. The cursor and the actions related to the cursor are specifics for each platforms. Allowing the user to use the cursor create unexpect behaviors with pasting when it comes to synchronize the InputPrice with the DialPad.

> I recommend to use either the platform inputs, removing our DialPad in favour of Input of the user platform, or to refactor the InputPrice to offer an alternative way to paste an amount.

PR #1300, opened 2025-10-12
Swap fixes

i5hi:
Various fixes in boltz swaps

### DIFF
--- a/lib/features/transactions/presentation/blocs/transaction_details/transaction_details_state.dart
+++ b/lib/features/transactions/presentation/blocs/transaction_details/transaction_details_state.dart
@@ -67,6 +67,34 @@ sealed class TransactionDetailsState with _$TransactionDetailsState {
     }
   }
 
+  /// Calculates the actual amount sent (received by recipient)
+  /// For swaps: amount - txFee + aggregateSwapFees
+  /// For regular transactions: amount - txFee
+  int getAmountSent() {
+    final swap = this.swap;
+    final txFee = walletTransaction?.feeSat ?? 0;
+    final amount = walletTransaction?.amountSat ?? 0;
+
+    if (swap == null) {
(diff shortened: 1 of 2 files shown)


lib/core/swaps/domain/entity/swap.dart
```diff
@@ -316,6 +316,44 @@ extension SwapFeePercent on Swap {
     final feePercent = getFeeAsPercentOfAmount();
     return feePercent > 5.0;
   }
+
+  /// Calculates the total aggregated fees for display in transaction details
```
kumulynja: Sounds more something to add in view models/presenters since it is to display, but no issue since we don't have that structure/layer yet for swaps or transaction details, so it's something we can do later, just mentioning it to have it in mind.

lib/features/swap/presentation/swap_bloc.dart
```diff
@@ -140,6 +140,10 @@ class SwapCubit extends Cubit<SwapState> {
 
       if (state.selectedFeeList == null) {
         await loadFees();
+        // Wait for fees to be loaded
+        while (state.loadingFees || state.selectedFeeList == null) {
+          await Future.delayed(const Duration(milliseconds: 100));
```
kumulynja, line 145: to avoid needing a while loop and delays here, I think all the logic of loading and estimating the fees and calculating the amount could be done in a use case that returns all of the needed calculated data as a single response struct, but again, that would be optimisation and not necessary for this PR I guess.

lib/core/swaps/domain/entity/swap.dart
i5hi: I can do it now its a simple change; i will move it to the state

i5hi: done

PR #1307, opened 2025-10-14
feat: Ark UI improvements

basantagoswami:
things done:

**dev mode gating**
- business logic checks for dev mode == on before turning on ark
- on wallet home page ark wallet is hidden when dev mode is off

**ark wallet setup**
- ark wallet can now gets revoked when dev mode is turned off and can be reactivated after turning on dev mode again

**balance display**
- balance card now shows total balance instead of separate confirmed/pending rows
- added info button that opens a balance breakdown modal, breakdown shows: confirmed, pending, and total

**txn details**
- new txn details page accessible by tapping any transaction. shows: txn id, type, status, amount etc
- transaction id is clickable (opens mempool.space for boarding transactions)
- some fields are copyable

**about page**
- new "about" page showing ark server information. displays things like the server URL, server pubkey, forfeit address, network, dust amount, session duration, exit delays, esplora URL
- relevant fields can be copied

**ui improvements**
- improved transaction history layout
- updated receive page layout to use bip21 unified qr code and uri
- refined settle bottom sheet UI

**other non ark stuff**
- dev mode toggle is a now a toggle instead of buttons

### DIFF
--- a/lib/features/ark_setup/setup_page.dart
+++ b/lib/features/ark_setup/setup_page.dart
@@ -38,6 +38,8 @@ Ark is still experimental.
 Your Ark wallet is derived from your main wallet's seed phrase. No additional backup is needed, your existing wallet backup restores your Ark funds too.
 
 By continuing, you acknowledge the experimental nature of Ark and the risk of losing funds.
+
+Developer note : Ark secret is derived from the main wallet seed using an arbitrary BIP-85 derivation (index 11811).
             '''),
                       const Spacer(),
                       if (error != null) ...[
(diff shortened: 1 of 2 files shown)


lib/features/ark/ui/transaction_history_widget.dart
```diff
@@ -19,7 +54,7 @@ class TransactionHistoryWidget extends StatelessWidget {
         child: Column(
           children: [
             const Gap(16),
-            BBText(
```
ethicnology, line 22: This should stay a `BBText`

basantagoswami: I was told BBText is deprecated. Does using Text cause any UI issues?

ethicnology: Not to my knowledge, we wish to have the opportunity to refactor/create a whole new UI Kit but the current one is `BBText`, `BBButton` etc… we must stay consistent to avoid extra work for the day we are refactoring the UI kit

PR #1311, opened 2025-10-15
Minor Fixes

i5hi:


ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -678,19 +678,19 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
-				CURRENT_PROJECT_VERSION = 96;
+				CURRENT_PROJECT_VERSION = 99;
 				DEVELOPMENT_TEAM = BX99T32YGS;
 				ENABLE_BITCODE = NO;
-				FLUTTER_BUILD_NAME = 5.4.5;
-				FLUTTER_BUILD_NUMBER = 96;
+				FLUTTER_BUILD_NAME = 6.0.0;
```
ethicnology: Should bump version be included in this PR?

pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: bb_mobile
 description: Bull Bitcoin Mobile Wallet
 publish_to: "none"
-version: 5.4.5+96
+version: 6.0.0+99
```
ethicnology: Do we want to include the bump in this PR
build number `97` can be used instead of `99` if previous was `96`

i5hi: ah yes; sorry ill revert - did that for the release on ios day before yesterday.

lib/features/pay/ui/widgets/pay_new_recipient_form.dart
```diff
@@ -794,17 +798,27 @@ class _BankTransferForm extends StatelessWidget {
...
   const _SepaForm({required this.formData, required this.onFormDataChanged});
 
   final Map<String, dynamic> formData;
   final Function(String, String) onFormDataChanged;
 
+  @override
+  State<_SepaForm> createState() => _SepaFormState();
+}
+
+class _SepaFormState extends State<_SepaForm> {
```
kumulynja, line 821: There are no widget state variables used, so this change to Stateful is unnecessary and discouraged as it is better to have StatelessWidget until you really keep state in the widget itself.

lib/features/withdraw/ui/widgets/new_recipient_form.dart
```diff
@@ -856,17 +856,27 @@ class _BankTransferForm extends StatelessWidget {
...
   const _SepaForm({required this.formData, required this.onFormDataChanged});
 
   final Map<String, dynamic> formData;
   final Function(String, String) onFormDataChanged;
 
+  @override
+  State<_SepaForm> createState() => _SepaFormState();
+}
+
+class _SepaFormState extends State<_SepaForm> {
```
kumulynja, line 879: Same comment as above, why use a StatefulWidget if no variables are kept in the widget state?

PR #1313, opened 2025-10-15
Improve Electrum Settings UI with extra and more clear messages, indicators and validations

kumulynja:


lib/core/electrum/application/usecases/set_custom_servers_priority_usecase.dart
```diff
@@ -16,7 +17,14 @@ class SetCustomServersPriorityUsecase {
     // Update each server's priority based on its position in the list
     final servers =
         request.servers.indexed.map((record) {
-          final server = record.$2.toDomain();
+          final dto = record.$2;
```
ethicnology: Better to use named records/tuple or class for deserialization rather than positioned `$1`, `$2`

lib/features/electrum_settings/frameworks/ui/widgets/set_advanced_options_bottom_sheet.dart
ethicnology, line 1: Why not re-use or update `BlurredBottomSheet` instead to keep the existing UI kit consistent across the codebase ?

ethicnology (changes requested):
- named records
- `BlurredBottomSheet` instead of a custom BottomSheet for consistency

lib/core/electrum/application/usecases/set_custom_servers_priority_usecase.dart
kumulynja: The positioned fields are because I am using `indexed` on a List to use the index as the priority, since the List has the servers in the same order as the user dragged them in the UI to change the priority. So the positioned fields do not come from a call that we define the return value from, it is the `indexed` getter from Dart itself that returns these positioned fields with $1 being the index of the element in the List and $2 the element itself.

lib/features/electrum_settings/frameworks/ui/widgets/set_advanced_options_bottom_sheet.dart
kumulynja: I will change it, I didn't know we had this component as the standard one to use. Good to know!

lib/core/electrum/application/usecases/set_custom_servers_priority_usecase.dart
ethicnology: If we can't do named, I've refactored the deserialization to be `final (index, dto) = record;` to avoid `$1` and `$2`

kumulynja: Awesome, great way to do it!

lib/features/electrum_settings/frameworks/ui/widgets/set_advanced_options_bottom_sheet.dart
kumulynja: BlurredBottomSheet is used now for the bottomsheet to add a server and for the advanced options bottom sheet 👍

PR #1320, opened 2025-10-15
fix: recipient filters in pay

i5hi:


kumulynja (approved):
Tested and working fine 👍

PR #1344, opened 2025-10-19
6.2.0

i5hi:
Various Minor Fixes:

    Input formatter in swaps for BTC values with decimals
    Duplicate watch only wallet entries is Transfer flow
    Complete swap without recieveTxid should retry
    Better background quality and correct orientation
    Fix in PaymentMethod enum preventing OrderSummary parsing
    Ark v8
    Default currency is now USD
    Add support to external descriptors
    Refactor CI (removing outdated and merge some flows to reduce github actions usage)
    refactor: sanitize \r on top of \t and \n and remove an unnecessary reverse by a sort in the filter


ethicnology:
+ ark v8
+ CAD to USD
+ Add support to external descriptors

PR #1376, opened 2025-10-24
Fix: Create both Bitcoin and Liquid wallets on import

echennells:
Previously the import wallet feature only created Bitcoin wallets, unlike the recovery flow which creates both.

Now creates both Bitcoin and Liquid wallets from the same seed, matching recovery behaviour.


ethicnology:
Hey @echennells 👋

I’m doing a bit of cleanup and closing some old or stale PRs. No worries about your work—the branch will still be there.

If you’re still interested in adding this, please open/bump the issue so we can discuss the approach before you spend more time on the implementation 💪

PR #1393, opened 2025-10-29
amount input cleanup (wip)

basantagoswami:


basantagoswami:
signal-2025-11-06-155805_002

Added a paste from clipboard button to make it easier to paste an amount. It only shows up when there in a numeric value in the clipboard

Also added validation so that user can't type amounts like 0001 sats, removes the leading zeros. Send screen should not have many of the issues now

PR #1398, opened 2025-10-30
Add support for Specter. Allow SeedSigner and Krux.

i5hi:
- Minor UI change to Import Wallet flow - no separate hardware wallet page; all shown on the same page now with dividers
- Added support for Specter (clone of SeedSigner)
- Show SeedSigner and Krux to all users


i5hi:
- [ ] Add Q instructrions update

PR #1468, opened 2025-11-11
feat: allow derivation path selection for raw xpubs Closes #1401

### LINKED ISSUE #1401: Allow derivation path selection for raw xpubs
Multiple users report not being able to access their balances and transactions when importing an xpub. This is due to either them not exporting the right public keys, or their wallet not allowing zpub export. Lets think of adding a derivation path selection when importing xpubs.

ethicnology:
⚠️ This feature introduce the possibility for a user to select a derivation path that does not match the one used to derive the given xpub.

### DIFF
--- a/lib/features/import_watch_only_wallet/presentation/watch_only_details_widget.dart
+++ b/lib/features/import_watch_only_wallet/presentation/watch_only_details_widget.dart
@@ -11,30 +11,15 @@ import 'package:bb_mobile/features/import_watch_only_wallet/watch_only_wallet_en
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:gap/gap.dart';
-import 'package:go_router/go_router.dart';
 import 'package:satoshifier/enums/derivation.dart' as satoshifier;
 
-class WatchOnlyDetailsWidget extends StatefulWidget {
-  const WatchOnlyDetailsWidget({super.key});
-
-  @override
-  State<WatchOnlyDetailsWidget> createState() => _WatchOnlyDetailsWidgetState();
-}
-
-class _WatchOnlyDetailsWidgetState extends State<WatchOnlyDetailsWidget> {
-  WatchOnlyWalletEntity? watchOnlyWallet;
-
-  @override
-  void didChangeDependencies() {
-    watchOnlyWallet =
-        context.watch<ImportWatchOnlyCubit>().state.watchOnlyWallet;
-    if (watchOnlyWallet == null) context.pop();
-    super.didChangeDependencies();
-  }
+class WatchOnlyDetailsWidget extends StatelessWidget {
+  final WatchOnlyWalletEntity watchOnlyWallet;
+  const WatchOnlyDetailsWidget({super.key, required this.watchOnlyWallet});
 
   @override
   Widget build(BuildContext context) {
-    return watchOnlyWallet!.when(
+    return watchOnlyWallet.when(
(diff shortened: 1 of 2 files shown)


ethicnology:
@i5hi added the disclaimer for the derivation path

localization/app_en.arb
```diff
@@ -7169,6 +7169,8 @@
   "@importWatchOnlyExtendedPublicKey": {
     "description": "Label for extended public key (xpub/ypub/zpub) display"
   },
+  "importWatchOnlyDisclaimerTitle": "Derivation Path Warning",
+  "importWatchOnlyDisclaimerDescription": "Make sure the derivation path you choose matches the one that produced the given xpub, as using the wrong path can lead to loss of funds",
```
i5hi: can we add "By verifying the first address derived from the wallet before use."?

lib/features/import_watch_only_wallet/presentation/watch_only_details_widget.dart
```diff
@@ -2,48 +2,55 @@ import 'package:bb_mobile/core/entities/signer_device_entity.dart';
...
 import 'package:gap/gap.dart';
+import 'package:go_router/go_router.dart';
+import 'package:satoshifier/enums/derivation.dart' as satoshifier;
 
-class WatchOnlyDetailsWidget extends StatelessWidget {
-  const WatchOnlyDetailsWidget({
-    required this.watchOnlyWallet,
-    required this.cubit,
-  });
+class WatchOnlyDetailsWidget extends StatefulWidget {
```
kumulynja: The only state field is coming from the cubit state already, so I don't think a StatefulWidget is needed here and you can remove all the `didChangeDependencies` etc.

lib/features/import_watch_only_wallet/presentation/watch_only_details_widget.dart
```diff
@@ -2,48 +2,55 @@ import 'package:bb_mobile/core/entities/signer_device_entity.dart';
...
+}
+
+class _WatchOnlyDetailsWidgetState extends State<WatchOnlyDetailsWidget> {
+  WatchOnlyWalletEntity? watchOnlyWallet;
+
+  @override
+  void didChangeDependencies() {
+    watchOnlyWallet =
+        context.watch<ImportWatchOnlyCubit>().state.watchOnlyWallet;
+    if (watchOnlyWallet == null) context.pop();
```
kumulynja: I am not sure if I understand this exactly, but it feels a bit weird that we have a widget for a watch only wallet, but then it is possible that there actually is no watch only wallet and so we go back. If there is no watch only wallet, shouldn't we avoid navigating here in the first place?

I feel this needs some explanation to understand what's going on.

kumulynja (changes requested):

lib/features/import_watch_only_wallet/presentation/watch_only_details_widget.dart
ethicnology: Good catch changed to stateless in 974874c89c0a5af7faf689ee1f1955caf2afec4a

lib/features/import_watch_only_wallet/presentation/watch_only_details_widget.dart
ethicnology: Good catch I've removed all that logic in 974874c89c0a5af7faf689ee1f1955caf2afec4a

localization/app_en.arb
ethicnology: Updated in c92c7bee1c739d18b130558199ce7d48950e399a

Issue #1437, opened 2025-11-05
Remove Typo form Transfer Warning


thibistaken:
In the green top banner, it says: "Transfer Bitcoin seamlessly between your wallets. Only keep funds in the Instant Payments wallet is for day-to-day spending". The 'is" in that sentence should be taken out. As in: "Only keep funds in the Instant Payments wallet for day-to-day spending."

<img width="1125" height="2436" alt="Image" src="https://github.com/user-attachments/assets/569796b3-e71e-49eb-8dae-e96e57328e61" />

Issue #1457, opened 2025-11-10
Formating fix for "Import wallet"


SovereignGG:
Minor change to be applied to the next version release if in the formatting of "Import Wallet", to align with the formatting of all other options the "wallet" should also have a capital letter. 

<img width="768" height="1713" alt="Image" src="https://github.com/user-attachments/assets/b5970bdd-19ee-44ba-b75e-4c666cc92716" />

PR #1477, opened 2025-11-12
Typo fixes in #1437 and #1457

i5hi:
Closes Issues #1437 and #1457


gluneau:
@ethicnology Do we need a process to check if the ES and FR translation need updating as well?  Otherwise the translations will drift out of sync.

Issue #1454, opened 2025-11-10
Very slow transaction loading


i5hi:
## Description
Loading transactions is very slow and gets slower if there are more transactions

**Likely causes:**

1. Label fetching:
In wallet_transaction_repository_impl.dart (lines 99-139), labels are fetched individually for each transaction, input, output, and address.
This results in many queries. For 100 transactions with 2 inputs and 2 outputs each:
100 transaction label queries
200 input label queries
200 output label queries
200 address label queries
Total: ~700 queries

2. Swap to transaction mapping:
n get_transactions_usecase.dart (lines 76-125), firstWhere is used for each transaction, which is O(n×m). This creates O(n×m) lookups and uses .remove() on lists (O(n)).

**Recommended fixes**
Batch fetch all labels: collect all refs (transaction IDs, input refs, output refs, addresses), fetch all labels in one or a few queries, then map in memory.

Use Maps for lookups: build Maps keyed by txId for swaps, payjoins, and orders, then use O(1) lookups instead of firstWhere.

Avoid .remove() on lists: use Sets or Maps to track what's been used.

thibistaken:
Depends on the re-arch. In that case transaction history relies on Wallet and more... see: https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/1785/changes#diff-0e40b74646edbfb06db3f21907c1b733d679d2cd52859d2e5885cb823a73fcaa

PR #1479, opened 2025-11-12
Faster load for transactions

i5hi:
This PR solves part of the problem in https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1454

It solves the following:
- Faster load of transaction details by passing data from the tx list cubit for faster initialization
- Cache transaction in the transaction cubit so user can still see all transactions on second visit to tx list page

What it still does not solve:
- First load of tx list is still slow (I attempted a few things but they did not really affect the speed much - will try out a few more things)

### DIFF
--- a/lib/features/transactions/presentation/blocs/transactions_cubit.dart
+++ b/lib/features/transactions/presentation/blocs/transactions_cubit.dart
@@ -15,6 +15,9 @@ part 'transactions_cubit.freezed.dart';
 part 'transactions_state.dart';
 
 class TransactionsCubit extends Cubit<TransactionsState> {
+  static final Map<String?, List<Transaction>> _transactionsCache = {};
+  static final Set<TransactionsCubit> _activeCubits = {};
+
   TransactionsCubit({
     String? walletId,
     required GetTransactionsUsecase getTransactionsUsecase,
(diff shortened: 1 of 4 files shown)


lib/features/transactions/presentation/blocs/transactions_cubit.dart
```diff
@@ -15,6 +15,9 @@ part 'transactions_cubit.freezed.dart';
 part 'transactions_state.dart';
 
 class TransactionsCubit extends Cubit<TransactionsState> {
+  static final Map<String?, List<Transaction>> _transactionsCache = {};
+  static final Set<TransactionsCubit> _activeCubits = {};
```
kumulynja, line 19: Back to adding cubits into cubits? :o 
I really don't think caching things in a cubit is the correct solution.
I think we need a thorough refactor of all underlying data structures and calls concerning wallets, transactions, addresses, labels, utxos, payjoins, swaps and how they are fetched and linked, and then have proper pagination.

i5hi: yeah; this is a naive/quick fix

i was thinking of pushing this as a temporary fix and then we prioritize the better solution in the next release

i5hi: let me try another approach

PR #1497, opened 2025-11-14
New Swap Master Key flow

i5hi:
First round of changes towards implementing the swap restore flow.

We are now using a single SwapMasterKey derived from the default wallet seed (handled by boltz library) for all swaps.


ethicnology:
@i5hi is this PR stale ?

BullishNode:
stale

PR #1534, opened 2025-11-19
BREAKING CHANGE: Allow users to define a custom Tor URL for the RecoverBull key server via app settings

ethicnology:
test server: `http://k7x5u62owmfd4qqpw6vzkogu4juhcgkx5qwtqsbsg3mqxayhetrjfzad.onion`


lib/core/storage/migrations/schema_9_to_10.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:bb_mobile/core/storage/sqlite_database.steps.dart';
+import 'package:drift/drift.dart';
+
+class Schema9To10 {
+  static Future<void> migrate(Migrator m, Schema10 schema10) async {
```
kumulynja, line 5: You need to seed here too, no?

Issue #1514, opened 2025-11-17
Tor via Orbot


i5hi:
As a first step towards supporting tor, we will allow users to connect to an external Orbot tor daemon and use this proxy while connected to bitcoin electrum. This should support using hidden service urls as well as normal urls.

Integration Checklist:
- [ ] Add a new TOR section in settings
- [ ]  Tor settings should have an active and deactive toggle and allow users to set the Tor proxy port (default for Orbot is 9950) and also show a connection status
- [ ] If Tor is activated; we must store this detail and always check when setting up Electrum for bitcoin as we need to always pass the proxy to the settings (maybe a good idea to add this as a new field in Electrum Settings useTorProxy: bool, torProxyPort: int) NOTE: If an entire url is required for the proxy use localhost
- [ ] On the landing/onboarding page where we have Create new wallet/ Recover Wallet; add a text at the bottom for `Advanced Options` where a user can configure Tor and/or add their own custom server.

Issue #427, opened 2025-01-09
[Tor] Add support


i5hi:
Add support for TOR via:

- [ ] Provided Tor Daemon: https://pub.dev/packages/tor
- [ ] External Tor (Orbot) Proxy?

The following libraries will also need support added:

- [ ] Boltz API Calls
- [ ] LWK Electrum Client

ethicnology:
related:
- https://github.com/SatoshiPortal/recoverbull-client-dart/issues/2
- https://github.com/Foundation-Devices/tor/issues/58

Issue #299, opened 2024-05-20
Support for connecting to Electrum server over Tor


katoog:
Please add support to connect to custom Electrum server with Tor address.

BullishNode:
concept ACK

i5hi:
To start with we can quite easily support setting a proxy port and using an external Tor daemon like Orbot. 

Additionally, bdk supports Tor but lwk currently does not. Will request lwk to add this.

thibistaken:
Yes, @emtreulapollaguera we'll prioritize this in our upcoming release in December.

PR #1542, opened 2025-11-21
connect to electrum servers using tor via orbot socks5 proxy

basantagoswami:
Related issues: #1514, #299 #427

Screenshots of changes done:

**Advanced option in during onboarding**

**Advanced options screen UI in onboarding flow**

**After enabling tor proxy**

**TOR proxy port edit option in onboarding**

**TOR settings screen - status connected**

**TOR proxy disabled**

**Disconnected (failed to connect to tor proxy)**

**Electrum servers error when all servers are offline and tor proxy is enabled**


lib/features/onboarding/ui/screens/advanced_options.dart
```diff
@@ -0,0 +1,159 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final advancedOptions = context.select(
+      (ElectrumSettingsBloc bloc) => bloc.state.advancedOptions,
+    );
+    final isLoading = context.select(
+      (ElectrumSettingsBloc bloc) => bloc.state.isLoading,
+    );
+
+    final useTorProxy = advancedOptions?.useTorProxy ?? false;
```
i5hi: should be in state

i5hi: all these params we should read directly from state

lib/features/electrum_settings/frameworks/ui/widgets/tor_proxy_error_banner.dart
```diff
@@ -0,0 +1,54 @@
...
+import 'package:bb_mobile/core/widgets/cards/info_card.dart';
+import 'package:bb_mobile/features/electrum_settings/interface_adapters/presenters/bloc/electrum_settings_bloc.dart';
+import 'package:bb_mobile/features/tor_settings/presentation/bloc/tor_settings_cubit.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+
+class TorProxyErrorBanner extends StatelessWidget {
+  const TorProxyErrorBanner({super.key});
+
+  bool _areAllServersOffline(ElectrumSettingsState state) {
```
i5hi: this method should be in the state file

lib/features/tor_settings/presentation/bloc/tor_settings_cubit.dart
```diff
@@ -0,0 +1,128 @@
...
+
+    emit(state.copyWith(status: TorStatus.connecting));
+
+    try {
+      // Test SOCKS5 proxy functionality by attempting SOCKS5 handshake
+      // This verifies both that the port is open AND that our app can use it
+      final socket = await Socket.connect(
+        '127.0.0.1',
+        state.torProxyPort,
+        timeout: const Duration(seconds: 3),
```
i5hi: should we give some more time for this?

lib/features/tor_settings/ui/screens/tor_settings_screen.dart
```diff
@@ -0,0 +1,189 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final advancedOptions = context.select(
+      (ElectrumSettingsBloc bloc) => bloc.state.advancedOptions,
+    );
+    final isLoading = context.select(
+      (ElectrumSettingsBloc bloc) => bloc.state.isLoading,
+    );
+
+    final useTorProxy = advancedOptions?.useTorProxy ?? false;
```
i5hi: same as previous comment; defaults should be set in the state and we should fetch directly from state

lib/features/tor_settings/ui/widgets/tor_port_input_bottom_sheet.dart
```diff
@@ -0,0 +1,124 @@
+import 'package:bb_mobile/core/themes/app_theme.dart';
+import 'package:bb_mobile/core/widgets/buttons/button.dart';
+import 'package:bb_mobile/core/widgets/text/text.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:gap/gap.dart';
+import 'package:go_router/go_router.dart';
+
+class TorPortInputBottomSheet extends StatefulWidget {
```
i5hi, line 9: we should use a separate state file for all these variables and methods.

Generally we should avoid StatefulWidget where possible - the only place we use it is if there is a form 

And we should avoid having too much logic in the UI

kumulynja: In this case, it actually should have a Form, since it is collecting user input for the port.
It should have a Form and use BBTextFormField instead of a normal TextField, then validator fields can be used instead of using onChange for it.

@basantagoswami you can check the electrum feature to see how Forms, BBTextFormField and the validator is used. Then you also shouldn't keep `errorText` like this, you return the error messages in the validator function.

lib/core/storage/database_seeds.dart
```diff
@@ -61,6 +61,8 @@ class DatabaseSeeds {
         stopGap: 20,
         timeout: 5,
         retry: 5,
+        useTorProxy: false,
```
kumulynja: I think we need a new function as well that only seeds these two columns, otherwise I think we will overwrite all settings for everyone.

So this function is ok to seed with `onCreate`, but in schema 9 to 10, we need a function that only seeds the tor columns in electrum settings and doesn't overwrite the other settings.

lib/core/storage/migrations/schema_9_to_10.dart
```diff
@@ -4,5 +4,9 @@ import 'package:drift/drift.dart';
 class Schema9To10 {
   static Future<void> migrate(Migrator m, Schema10 schema10) async {
     await m.createTable(schema10.recoverbull);
+
+    final electrumSettings = schema10.electrumSettings;
+    await m.addColumn(electrumSettings, electrumSettings.useTorProxy);
+    await m.addColumn(electrumSettings, electrumSettings.torProxyPort);
```
kumulynja: Seeding for both recoverbull as for the newly added electrum settings columns is needed after this.

lib/features/electrum_settings/frameworks/ui/widgets/tor_proxy_error_banner.dart
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    return allServers.every(
+      (server) => server.status == ElectrumServerStatus.offline,
+    );
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final electrumState = context.watch<ElectrumSettingsBloc>().state;
```
kumulynja: Better to use `context.select` for only the fields of the state that you need in the widget to avoid re-rendering it changes of all fields of the state.

lib/features/onboarding/ui/screens/advanced_options.dart
```diff
@@ -0,0 +1,159 @@
...
+    super.initState();
+    // Load Electrum settings for Bitcoin Mainnet
+    context.read<ElectrumSettingsBloc>().add(
+      const ElectrumSettingsLoaded(isLiquid: false),
+    );
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final advancedOptions = context.select(
```
kumulynja: `context.select` shouldn't be used in `StatefulWidget`s as they will not make the widget re-render if their value in the state changes. If you need the value at the first render, you need to set it in the `initState` function with a `context.read` instead.

lib/features/onboarding/ui/screens/advanced_options.dart
kumulynja: Yeah, I think `advancedOptions` should be a `late` param of the `_AdvancedOptionsState` here and it should set in the `initState` function by reading from the BLoC's state. Then it can be used in the rest of the widget.

lib/features/onboarding/ui/screens/advanced_options.dart
```diff
@@ -0,0 +1,159 @@
...
+      (ElectrumSettingsBloc bloc) => bloc.state.advancedOptions,
+    );
+    final isLoading = context.select(
```
kumulynja: Same comment, context.select in a StatefulWidget will only get the value of isLoading from the first render of the widget, it will not re-render the widget when the state changes. If you need to rebuild the stateful widget when bloc.state.isLoading changes, there are a couple of options:

- Add the isLoading as a `late` param of the `_AdvancedOptionsState`, set the initial value in `initState` and also subscribe to the bloc state stream in the `initState` and use `setState({isLoading= state.isLoading;})` in the listener.
- Use `BlocSelector` to wrap the widgets on this screen that needs to change when `state.isLoading` changes.

lib/features/onboarding/ui/screens/advanced_options.dart
```diff
@@ -0,0 +1,159 @@
...
+      _AdvancedOptionsState();
+}
+
+class _AdvancedOptionsState
+    extends State<AdvancedOptions> {
+  @override
+  void initState() {
+    super.initState();
+    // Load Electrum settings for Bitcoin Mainnet
+    context.read<ElectrumSettingsBloc>().add(
```
kumulynja: You could do this already where the BLoC is created, so probably in the router where it is provided. That's where we usually do it if it is an initialization event to load the initial state.

lib/features/tor_settings/presentation/bloc/tor_settings_cubit.dart
```diff
@@ -0,0 +1,128 @@
...
+import 'package:bb_mobile/core/utils/logger.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+part 'tor_settings_cubit.freezed.dart';
+part 'tor_settings_state.dart';
+
+class TorSettingsCubit extends Cubit<TorSettingsState> {
+  TorSettingsCubit({
+    required ElectrumSettingsRepository electrumSettingsRepository,
```
kumulynja: You shouldn't directly use a Repository in a Cubit. A Cubit should not have any business logic or any logic that is unrelated with transforming data from the usecases to the state and from the UI to the usecases. You should create and inject use cases and these usecases should contain the business/orchestration logic.

lib/features/tor_settings/presentation/bloc/tor_settings_cubit.dart
```diff
@@ -0,0 +1,128 @@
...
+      log.warning('Tor proxy connection failed: $e');
+    } on TimeoutException catch (e) {
+      // Connection timeout - Tor proxy is not responding
+      emit(state.copyWith(status: TorStatus.offline));
+      log.warning('Tor proxy connection timeout: $e');
+    } catch (e) {
+      // Other errors - mark as unknown
+      emit(state.copyWith(status: TorStatus.unknown));
+      log.severe('Unexpected error checking Tor status: $e');
+    }
```
kumulynja: This is way too much logic for a Cubit function and it uses a very low/infrastructure level thing which is the `Socket`. All of this should be abstracted away in a reusable UseCase that orchestrates between the ElectrumSettingsRepository and a SocketPort interface and return what you want to obtain here which is just the TorStatus.

lib/features/tor_settings/presentation/bloc/tor_settings_cubit.dart
```diff
@@ -0,0 +1,128 @@
...
+  }
+
```
kumulynja, line 75: This is way too much logic for a Cubit function and it uses a very low/infrastructure level thing which is the Socket. All of this should be abstracted away in a reusable UseCase that orchestrates between the ElectrumSettingsRepository and a SocketPort interface and return what you want to obtain here which is just the TorStatus.

lib/features/tor_settings/ui/screens/tor_settings_screen.dart
```diff
@@ -0,0 +1,189 @@
...
+}
+
+class _TorSettingsScreenState extends State<TorSettingsScreen> {
+  @override
+  void initState() {
+    super.initState();
+    context.read<ElectrumSettingsBloc>().add(
+      const ElectrumSettingsLoaded(isLiquid: false),
+    );
+    context.read<TorSettingsCubit>().init();
```
kumulynja: These can be done already where the BLoC and Cubit are provided.

lib/features/tor_settings/ui/screens/tor_settings_screen.dart
```diff
@@ -0,0 +1,189 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final advancedOptions = context.select(
+      (ElectrumSettingsBloc bloc) => bloc.state.advancedOptions,
+    );
+    final isLoading = context.select(
+      (ElectrumSettingsBloc bloc) => bloc.state.isLoading,
+    );
+
```
kumulynja: See comments above about using context.select in a StatefulWidget. It doesn't work as you'd expect, it doesn't make the widget re-render on state change.

kumulynja (changes requested):

basantagoswami:
Had to do a complete refactor of the PR because storing torProxyPort etc for each Electrum server doesn't make sense. Moved those to settings

Issue #1546, opened 2025-11-23
Minor Fixes for 6.4.0


i5hi:
- [ ] Clicking on configure electrum server settings error banner on wallet home screen still takes you to settings, not electrum server settings

- [ ] The Delete Custom Server popup has a look that doesn't match the rest of the app, need to update at least the 'Delete' button, it's red on black now.

- [ ] The status indicator on wallet home screen still has too small of a touch area, works well on emulators, but on real physical devices, it's hard to tap on

- [ ] We turn off Ark when dev mode is turned off, but we need to make sure that we also turn off dev mode when superuser mode is turned off. It's not handled right now

- [ ] The electrum server error banner doesn't go away even after the status indicator goes green because servers are back

basantagoswami:
Working on the branch [minor-fixes-for-6.4.0](https://github.com/SatoshiPortal/bullbitcoin-mobile/tree/minor-fixes-for-6.4.0) for this issue

basantagoswami:
Converted the Recoverbull embedded Tor client over SOCKS5 proxy thing to an issue of its own

PR #1566, opened 2025-11-26
Minor fixes for 6.4.0

basantagoswami:
Changed made:

1. Electrum URL Parser - New utility to parse Electrum server URLs with :s/:t suffix support and smart
   SSL defaults (.onion → TCP, clearnet → SSL)
2. Electrum Settings UI - Auto-detect SSL from URL format, show "(Auto-detected)" label, improved
  validation using parser
3. Delete Server Dialog - Redesigned with BBButton styling, better layout, and improved error button
  colors
4. Backup Settings - Adjusted button spacing (5→12 gap between buttons, 20 gap at bottom)
5. BIP329 Labels - Added SafeArea, adjusted padding (24→32 horizontal, 16→12 gap), centered heading
6. Recoverbull PIN Screen - Fixed dialpad alignment to bottom, added proper spacing, prevented input
  field jump with fixed 60px height container
7. Settings Reorganization - Moved Tor Settings from Bitcoin Settings to App Settings
8. Electrum Error Navigation - Error banner now navigates to Electrum Settings (not general Settings)
9. Text Fixes - Changed "Bitcoin Settings" → "App Settings" in advanced options during onboarding
10. Localization - Fixed spacing in privacy notice text (added extra newline)
11. RecoverBull Settings UI - Redesigned with view/edit mode toggle, improved URL display, back button
  navigation, and "Learn more" link
12. RecoverBull Server Confirmation - Added back button, changed Continue button from red to black, added
  "Learn more" link

Related issue: https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1546

### DIFF
--- a/lib/core/utils/electrum_url_parser.dart
+++ b/lib/core/utils/electrum_url_parser.dart
@@ -10,10 +10,16 @@ class ElectrumUrlParser {
   ///   - `.onion` addresses default to TCP (no SSL)
   ///   - Clearnet addresses default to SSL
   ///
-  /// Returns null if URL is invalid or incomplete
-  static ({String cleanUrl, bool enableSsl})? parse(String input) {
+  /// Throws ElectrumUrlValidationError if URL is invalid or incomplete
+  static ({String cleanUrl, bool enableSsl}) parse(String input) {
     final trimmedInput = input.trim();
-    if (trimmedInput.isEmpty) return null;
+    if (trimmedInput.isEmpty) throw ElectrumUrlValidationError.empty;
+
+    // Check for protocol prefix
+    final protocolPattern = RegExp('^([a-zA-Z]+)://');
+    if (protocolPattern.hasMatch(trimmedInput)) {
+      throw ElectrumUrlValidationError.hasProtocol;
+    }
 
     // Check for :s or :t suffix
     final suffixMatch = RegExp(
@@ -40,34 +46,15 @@ class ElectrumUrlParser {
       return (cleanUrl: trimmedInput, enableSsl: !isOnion);
     }
 
-    return null;
+    throw ElectrumUrlValidationError.invalidFormat;
   }
 
-  /// Validates Electrum server URL format
-  ///
-  /// Accepts:
-  /// - `host:port` format
-  /// - Optional `:s` or `:t` suffix for SSL/TCP specification
-  ///
-  /// Rejects:
-  /// - Protocol prefixes (ssl://, tcp://, etc.)
-  /// - Invalid characters or format
-  static ElectrumUrlValidationError? validate(String input) {
-    final trimmedInput = input.trim();
-    if (trimmedInput.isEmpty) return ElectrumUrlValidationError.empty;
-
-    // Check for protocol prefix
-    final protocolPattern = RegExp('^([a-zA-Z]+)://');
-    if (protocolPattern.hasMatch(trimmedInput)) {
-      return ElectrumUrlValidationError.hasProtocol;
+  /// This is a non-throwing version of parse
+  static ({String cleanUrl, bool enableSsl})? tryParse(String input) {
+    try {
+      return parse(input);
+    } catch (_) {
+      return null;
     }
-
-    // Validate host:port format (with optional :s or :t suffix)
-    final hostPortPattern = RegExp(r'^[a-zA-Z0-9.-]+:\d+(:[st])?$');
-    if (!hostPortPattern.hasMatch(trimmedInput)) {
-      return ElectrumUrlValidationError.invalidFormat;
-    }
-
-    return null;
   }
 }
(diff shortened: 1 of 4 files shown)


lib/core/utils/electrum_url_parser.dart
kumulynja, line 1: This is good, but since both `parse` and `validate` repeat some validation and regexes, maybe it is better to have only `parse` and `tryParse` like this:

```
/// Returns parsed values or throws error if not valid
static ({String cleanUrl, bool enableSsl}) parse(String input) {
      /// Do parsing and validation and throw `ElectrumUrlValidationError` on error
}

/// Returns parsed values or null if not valid
static ({String cleanUrl, bool enableSsl})? tryParse(String input) {
        try {
            final result = parse(input);
            return result;
        } catch {
            return null;
        }
}
````

This keeps all validation and parsing logic in one and the same function (`parse`) while also giving a function that doesn't throw errors. It is more idiomatic and the same as the native int and double parsing api.
Of course instead of receiving an error as return parameter, you have to try/catch it.

File: lib/features/electrum_settings/frameworks/ui/widgets/add_custom_server_bottom_sheet.dart [hunk 1]
```diff
@@ -42,21 +44,46 @@ class _AddCustomServerBottomSheetState
...
   void _submit() {
     if (_formKey.currentState!.validate()) {
       // Unfocus to close keyboard before popping (optional, just looks nicer)
       FocusScope.of(context).unfocus();
-      Navigator.of(context).pop(
-        CustomServerInput(url: _controller.text.trim(), enableSsl: _enableSsl),
-      );
+
+      final result = ElectrumUrlParser.parse(_controller.text.trim());
+      final cleanUrl = result?.cleanUrl ?? _controller.text.trim();
```
kumulynja: Is this still needed here if the validator already does this?

If it is still needed because the domain layer doesn't process the :s or :t suffix, then it is better to do it in the BLoC, because of separation of concerns:

- UI permits inputting of url with suffix and uses that to toggle ssl, so for the UI it's valid input and it doesn't really need the clean url for something itself, so also don't bother about it here.
- Domain/application layer expects the clean url, and the single responsibility of a BLoC is transforming and passing data between UI and domain/application, so do the parsing to get the clean url in the BLoC before calling the use case, not here, for the UI it's validated already.

lib/features/recoverbull/ui/pages/settings_page.dart
```diff
@@ -102,36 +121,102 @@ class _SettingsPageState extends State<SettingsPage> {
...
+                          padding: const EdgeInsets.all(16),
+                          decoration: BoxDecoration(
+                            color: context.colour.surfaceContainerHighest,
+                            borderRadius: BorderRadius.circular(8),
+                            border: Border.all(
+                              color: context.colour.outline.withValues(alpha: 0.2),
+                            ),
+                          ),
+                          child: BBText(
+                            _urlController.text,
```
kumulynja: Use the original url here, since this is the non-editable one that is supposed to be stored, so also use that value. The controller is for the text field to edit it.

lib/features/recoverbull/ui/pages/settings_page.dart
```diff
@@ -102,36 +121,102 @@ class _SettingsPageState extends State<SettingsPage> {
...
+                        mainAxisAlignment: MainAxisAlignment.spaceBetween,
+                        children: [
+                          BBText(
+                            'Key Server URL',
+                            style: context.font.titleMedium,
+                            color: context.colour.onSurface,
+                          ),
+                          if (!_isEditing)
+                            TextButton.icon(
+                              onPressed: () => setState(() => _isEditing = true),
```
kumulynja: On onPressed here, you should do _urlController.text = _originalText as well.
Since this is the moment you want to edit what is in _originalText.

lib/features/recoverbull/ui/pages/settings_page.dart
```diff
@@ -42,6 +44,7 @@ class _SettingsPageState extends State<SettingsPage> {
     try {
       final url = await _fetchUrlUsecase.execute();
       _urlController.text = url.toString();
```
kumulynja: Not needed here, only when editing is pressed.

lib/features/recoverbull/ui/pages/settings_page.dart
```diff
@@ -56,14 +59,31 @@ class _SettingsPageState extends State<SettingsPage> {
...
     } catch (e) {
       log.warning('Error saving recoverbull url: $e');
     } finally {
       if (mounted) setState(() => _isSaving = false);
     }
   }
 
+  void _cancelEdit() {
+    setState(() {
+      _urlController.text = _originalUrl;
```
kumulynja: Not needed here, only when pressing the edit button.

kumulynja (approved):
Thanks for the small changes, great job @basantagoswami! 👍

PR #1571, opened 2025-11-27
implement observer pattern for settings dependent features (ark, dev mode)

basantagoswami:
Issue: #1570


basantagoswami:
When `superuser mode` is disabled, `dev mode` disables itself. When `dev mode` is disabled, `Ark` disabled itself. Also the UI in settings and wallet home screen responds reactively to the change.

PR #1604, opened 2025-12-02
Seed viewer in superuser mode + Minor changes

i5hi:
Seed viewer to view all seeds in secure storage.
Allow users to delete seeds that are not linked to a wallet.
Move recoverbull from app settings to wallet backup.
Updated Icon


lib/core/seed/domain/usecases/get_all_seeds_from_secure_storage_usecase.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:bb_mobile/core/seed/data/repository/seed_repository.dart';
+import 'package:bb_mobile/core/seed/domain/entity/seed.dart';
+import 'package:bb_mobile/core/utils/logger.dart';
+
+class GetAllSeedsFromSecureStorageUsecase {
```
kumulynja: I would rename this to just GetAllSeedsUsecase, as it shouldn't matter if it is stored in secure storage or not, the repository abstracts away the details of where it is stored and how to fetch it.

lib/features/all_seed_view/presentation/all_seed_view_cubit.dart
```diff
@@ -0,0 +1,98 @@
...
+
+      // Process and separate seeds into existing and old wallets
+      final result = _processAndSeparateSeedsUsecase.execute(
+        seeds: seeds,
+        existingFingerprints: existingFingerprints,
+      );
+
+      emit(
+        state.copyWith(
+          loading: false,
```
kumulynja: To avoid repeating `loading: false` both here and in the `catch` clause, you can add a `final` clause that just does `emit(state.copyWith(loading: false)` , it makes it easier to not forget to update in two places later if needed at some point.

lib/features/all_seed_view/ui/all_seed_view_screen.dart
```diff
@@ -0,0 +1,335 @@
...
+        }
+      });
+    }
+  }
+
+  @override
+  void dispose() {
+    // Clear state when navigating away
+    _cubit?.clearState();
+    _cubit?.hideSeeds();
```
kumulynja: A Cubit/Bloc has a `dispose` function itself where resources should be cleaned up when it's closed, this shouldn't be done here.

lib/features/all_seed_view/ui/all_seed_view_screen.dart
```diff
@@ -0,0 +1,335 @@
...
+    // Fetch seeds after the first frame is rendered to avoid blocking navigation
+    if (!_hasStartedFetch) {
+      _hasStartedFetch = true;
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (mounted) {
+          _cubit?.fetchAllSeeds();
+        }
+      });
+    }
+  }
```
kumulynja: This is all unnecessary, this shouldn't be a StatefulWidget, why would you need to store the cubit in the state of the widget if it is already accessible through the context?
Also the `_hasStartedFetch` variable seems not needed since the cubit state has the `isLoading` already that reflects if fetching is going on or not.
So change to StatelessWidget and remove all of this and just use the Cubit directly everywhere.

lib/features/all_seed_view/ui/all_seed_view_screen.dart
```diff
@@ -0,0 +1,335 @@
...
+          backgroundColor: context.colour.onPrimary,
+          title: Text(
+            'Security Warning',
+            style: context.font.headlineSmall?.copyWith(
+              color: context.colour.onSurface,
+            ),
+          ),
+          content: SingleChildScrollView(
+            child: Text(
+              'Displaying seed phrases is a security risk. Anyone who sees your seed phrase can access your funds. Make sure you are in a private location and that no one can see your screen.',
```
kumulynja: Best to use translation files directly now in new features for all labels.

lib/core/seed/data/datasources/seed_datasource.dart
```diff
@@ -43,6 +44,41 @@ class SeedDatasource {
...
+          final json = jsonDecode(value) as Map<String, dynamic>;
+          final seedModel = SeedModel.fromJson(json);
+          seeds.add(seedModel);
+        } catch (e) {
+          // Skip keys that are not seed objects
+          continue;
+        }
+      }
+
+      return seeds;
```
kumulynja, line 75: What I find weird here is that the actual fetching from Secure Storage is outside of the isolate, so the problem isn't with the secure storage being slow then. But anyways if we can parse the data in an isolate and that improves performance, that's good 👍

lib/core/seed/data/repository/seed_repository.dart
```diff
@@ -73,4 +73,37 @@ class SeedRepository {
   }
 
   Future<void> delete(String fingerprint) => _source.delete(fingerprint);
+
+  Future<List<MnemonicSeed>> getAllMnemonicSeeds() async {
+    try {
+      final models = await _source.getAll();
```
kumulynja, line 79: Same as above, the actual interaction with Secure Storage is done here and the isolate is only used to parse the content, so was the parsing maybe done in a blocking way that made it hang?

Anyways if we can parse the data in an isolate and that improves performance, that's good 👍

PR #1644, opened 2025-12-09
Enable autoswap by default

i5hi:
- [x] Enable autoswap by default
- [x] Migrations added to 10To11
- [x] Make Autoswap settings its own page
- [x] Show warning for the first time and use `Do not show again` option to hide

### DIFF
--- a/lib/features/autoswap/ui/screens/autoswap_settings_screen.dart
+++ b/lib/features/autoswap/ui/screens/autoswap_settings_screen.dart
@@ -43,34 +43,18 @@ class AutoSwapSettingsScreen extends StatelessWidget {
                           children: [
                             const Gap(16),
                             _EnabledToggle(),
-                            const Gap(16),
-                            Stack(
-                              children: [
-                                Column(
-                                  crossAxisAlignment: CrossAxisAlignment.start,
-                                  mainAxisSize: MainAxisSize.min,
-                                  children: [
-                                    _AmountThresholdField(),
-                                    const Gap(16),
-                                    _TriggerBalanceField(),
-                                    const Gap(16),
-                                    _FeeThresholdField(),
-                                    const Gap(16),
-                                    _WalletSelectionDropdown(),
-                                    const Gap(16),
-                                    _AlwaysBlockToggle(),
-                                  ],
-                                ),
-                                if (!enabled)
-                                  Positioned.fill(
-                                    child: IgnorePointer(
-                                      child: Container(
-                                        color: context.appColors.overlay,
-                                      ),
-                                    ),
-                                  ),
-                              ],
-                            ),
+                            if (enabled) ...[
+                              const Gap(16),
+                              _AmountThresholdField(),
+                              const Gap(16),
+                              _TriggerBalanceField(),
+                              const Gap(16),
+                              _FeeThresholdField(),
+                              const Gap(16),
+                              _WalletSelectionDropdown(),
+                              const Gap(16),
+                              _AlwaysBlockToggle(),
+                            ],
                           ],
                         ),
                 ),


lib/core/swaps/domain/usecases/auto_swap_execution_usecase.dart
```diff
@@ -129,15 +130,13 @@ class AutoSwapExecutionUsecase {
...
-            : ApiServiceConstants.bbElectrumUrl;
+    final btcElectrumUrl = defaultBitcoinWallet.isTestnet
+        ? ApiServiceConstants.bbElectrumTestUrl
+        : ApiServiceConstants.bbElectrumUrl;
 
-    final lbtcElectrumUrl =
-        defaultLiquidWallet.isTestnet
-            ? ApiServiceConstants.publicElectrumTestUrl
-            : ApiServiceConstants.bbLiquidElectrumUrlPath;
+    final lbtcElectrumUrl = defaultLiquidWallet.isTestnet
```
kumulynja, line 137: Not possible with our current structure and core DI, but in the future, we should use an electrum port here and fetch the electrum servers to use from the electrum servers feature through it.

lib/features/autoswap/presentation/autoswap_settings_cubit.dart
```diff
@@ -285,4 +393,96 @@ class AutoSwapSettingsCubit extends Cubit<AutoSwapSettingsState> {
       ),
     );
   }
+
+  void toggleBitcoinUnit() {
```
kumulynja, line 397: A lot of the functions in this cubit do not send any data to a usecase and so I would personally manage them in the widget, only passing them to the cubit on confirmation. Everything needed from data to calculate stuff could be loaded in the state when the cubit is created. This to avoid it is possible to change things after confirmation and have discrepancies between the confirmed and cubit state. But if it works for now I guess its ok, nothing critical.

lib/features/autoswap/ui/screens/autoswap_settings_screen.dart
```diff
@@ -0,0 +1,548 @@
...
+          'Base Instant Wallet Balance',
+          style: context.font.bodyLarge?.copyWith(
+            color: context.appColors.text,
+          ),
+        ),
+        const Gap(8),
+        Row(
+          children: [
+            Expanded(
+              child: BBInputText(
```
kumulynja, line 137: Better to use BBTextFormField, add everything in a form and add validator and formatting as we use for other amount fields everywhere

lib/features/autoswap/ui/screens/autoswap_settings_screen.dart
```diff
@@ -0,0 +1,548 @@
...
+          'Trigger At Balance',
+          style: context.font.bodyLarge?.copyWith(
+            color: context.appColors.text,
+          ),
+        ),
+        const Gap(8),
+        Row(
+          children: [
+            Expanded(
+              child: BBInputText(
```
kumulynja, line 246: Same as above, better to use BBTextFormField, it is available in the recipients feature, but it can/should move to the core ui kit for reuse in all features needing form fields.

lib/features/autoswap/ui/screens/autoswap_settings_screen.dart
kumulynja: It is available in the recipients feature, but it can/should move to the core ui kit for reuse in all features needing form fields.

lib/features/autoswap/ui/screens/autoswap_settings_screen.dart
```diff
@@ -0,0 +1,548 @@
...
+          context.loc.autoswapMaxFeeLabel,
+          style: context.font.bodyLarge?.copyWith(
+            color: context.appColors.text,
+          ),
+        ),
+        const Gap(8),
+        Row(
+          children: [
+            Expanded(
+              child: BBInputText(
```
kumulynja, line 344: Same, use BBTextFormField with validator and formatting. Put all fields in a form for validation when pressing Confirm/Continue.

lib/features/autoswap/ui/autoswap_settings_router.dart
```diff
@@ -1,10 +1,11 @@
...
 import 'package:flutter/material.dart';
+import 'package:go_router/go_router.dart';
 
 /// Router for the Auto Swap Settings feature
 class AutoSwapSettingsRouter {
-  /// Shows the Auto Swap Settings bottom sheet
+  /// Shows the Auto Swap Settings screen
   static void showAutoSwapSettings(BuildContext context) {
-    AutoSwapSettingsBottomSheet.showBottomSheet(context);
+    context.pushNamed(SettingsRoute.autoswapSettings.name);
```
kumulynja: This doesn't feel like a very useful class and doesn't really has any routes to be called a router class, context.pushNamed can just be used wherever showAutoSwapSettings is used now.

lib/features/autoswap/presentation/autoswap_settings_cubit.dart
i5hi: ive moved it to state now

PR #1661, opened 2025-12-14
feat: Wallet selection on SendAmountScreen #1529 and remove network selection #1296

wired-pasteque:
- Remove the NetworkDisplay from Send #1296
- Allow wallet selection on Send feature (I've used the dropdown similar to buy as mentioned in #1529)
- Replaces hardcoded default wallet labels with localized keys (Wallet.displayLabel)
- Fixes layout on the Send Amount screen pinning the Continue button to the bottom of the screen

[Screenshot](https://github.com/user-attachments/assets/0d4cd394-d88b-42b2-a044-336cd95490c7)

Issue #1529, opened 2025-11-18
Add Wallet Selector in Send and Receive


thibistaken:
Feature request from @BullishNode but needs Figma to define the UI. Only for onchain, because otherwise we default to Instant Payments wallet.

BullishNode:
We could use same kind of component: dropdown with balance. But make the dropdown smaller, neater. That way, we dont use any new real estate, we just replace the what we already are using 


![InShot_20251122_104846850.jpg](https://github.com/user-attachments/assets/5c73a9fe-d22a-431c-a659-26294e600b61)

![InShot_20251122_104914337.jpg](https://github.com/user-attachments/assets/94046c8e-48e8-4149-a5cc-c7a924a84ca6)

i5hi:
Can be done along with [#1296](https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1296)

PR #1662, opened 2025-12-14
feat: Add wallet selection to receive bitcoin #1529

wired-pasteque:
- This PR is based on the changes I've made in my yesterday's one https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/1661
- It adds wallet selection on receive as requested in #1529 only on bitcoin, since @ethicnology told me there can be only one liquid wallet. Also, the wallet selector is shown only if there is more than one bitcoin wallet

[Screenshot](https://github.com/user-attachments/assets/4b9ba1e4-7259-48ad-9e43-251c8d02b32e)


BullishNode:
What do you guys think about putting it above the QR code?

PR #1664, opened 2025-12-15
update: default electrum server to fulcrum

i5hi:
- [x] Add fulcrum.bullbitcoin.com:50002 as top priority default
- [x] Remove fulcrum.bullbitcoin.com:50002 from existing custom server entries for users who added it manually
- [x] Added to migration10To11

### DIFF
--- a/lib/core/storage/migrations/schema_10_to_11.dart
+++ b/lib/core/storage/migrations/schema_10_to_11.dart
@@ -1,5 +1,6 @@
 import 'package:bb_mobile/core/storage/sqlite_database.dart';
 import 'package:bb_mobile/core/storage/sqlite_database.steps.dart';
+import 'package:bb_mobile/core/utils/constants.dart';
 import 'package:drift/drift.dart';
 
 class Schema10To11 {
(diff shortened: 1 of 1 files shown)


lib/core/storage/migrations/schema_10_to_11.dart
```diff
@@ -39,5 +39,58 @@ class Schema10To11 {
     await db.managers.settings.update(
       (f) => f(themeMode: const Value('system')),
     );
+
+    final electrumServers = schema11.electrumServers;
+    const fulcrumUrl = 'ssl://fulcrum.bullbitcoin.com:50002';
```
kumulynja: Use the constant `ApiServiceConstants.fulcrumElectrumUrl` here too.

lib/core/storage/migrations/schema_10_to_11.dart
```diff
@@ -39,5 +39,58 @@ class Schema10To11 {
...
+    const fulcrumUrlWithoutProtocol = 'fulcrum.bullbitcoin.com:50002';
```
kumulynja: If really needed, dynamically take the protocol part off of `fulcrumUrl`. But we never stored servers without the protocol for mainnet right? Or did we?

kumulynja (changes requested):

lib/core/storage/migrations/schema_10_to_11.dart
i5hi: we store bitcoin with the protocol prefix but not liquid

lib/core/storage/migrations/schema_10_to_11.dart
i5hi: fixed.

lib/core/storage/migrations/schema_10_to_11.dart
i5hi: fixed.

PR #1668, opened 2025-12-17
Coinselection continue button

i5hi:
Only enable continue on coinseletion when amount reqested < amount selected


ethicnology:
Can I cherry-pick this commit in #1669 to avoid dealing with conflicts?

Issue #319, opened 2024-08-20
[Delete Wallet] Remove bdk/lwk wallet db on delete


i5hi:
We removed deleting wallet db because of the issue with interference with `sync` 

We have to ensure that there is no ongoing sync when a wallet is being deleted and bring back removal of bdk/lwk db files.

PR #1672, opened 2025-12-17
feat: remove BDK/LWK wallet DB files on wallet deletion #319

ethicnology:


lib/core/wallet/data/datasources/bdk_facade.dart
kumulynja, line 1: Why is this called a `facade` if bdk isn't independent of the `wallet` feature?
Facades are used to expose the methods/services that should not directly be accessed by the caller. But in the case of `bdk` and `lwk`, there is no problem that the wallet datasources use bdk/lwk's methods directly. This class that is now called `facade` is really just a `factory`, since it just creates bdk/lwk wallet instances and nothing else. No need to change this now though, just mentioning it.

lib/core/wallet/data/datasources/bdk_facade.dart
```diff
@@ -119,6 +121,10 @@ class BdkFacade {
     return wallet;
   }
 
+  static Future<String> getWalletDbPath(WalletModel walletModel) async {
```
kumulynja: Instead of exposing the db path here, it would be better to expose a delete function here and the delete function here uses the private _getDbPath and deletes the wallet. Then the datasource shouldn't know the details about what's behind creating/deleting a wallet. And then both create and delete on this level are in the same class, which makes more sense. Leaking the details of a db path that is needed for a bdk/lwk wallet is not something useful to the datasource.

lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
```diff
@@ -612,6 +604,20 @@ class BdkWalletDatasource {
...
+
+  Future<void> delete({required WalletModel wallet}) async {
+    try {
+      final dbPath = await BdkFacade.getWalletDbPath(wallet);
+      final dbFile = File(dbPath);
+
+      if (!await dbFile.exists()) throw WalletNotFoundError();
+
+      await dbFile.delete();
+      log.fine('Deleted wallet ${wallet.id} BDK database');
```
kumulynja: Move this get path, getting the file and deleting the file to the BdkFacade (factory) and just call the delete function here. Like that all path and file related things, both for create as for delete are maintained in a single place and not leaked.

lib/core/wallet/data/datasources/lwk_facade.dart
kumulynja, line 1: Same comments for lwk_facade as for bdk_facade.

lib/core/wallet/data/datasources/lwk_wallet_datasource.dart
kumulynja, line 1: Same comments for lwk_wallet_datasource as for bdk_wallet_datasource regarding not handling the path and file stuff in the datasource, but containing it in the factory.

lib/core/wallet/data/repositories/wallet_repository.dart
kumulynja, line 380: There is not really a risk here I think since bdk/lwk recreate the deleted files everytime we need a wallet instance, but it seems more logical to first delete the metadata and then delete the lwk/bdk data.
In case something goes wrong with deleting the metadata, at least the wallet files of it are still there, otherwise it still looks like the wallet exists, but the wallet data would already be deleted. (Which is not really a risk or issue with bdk/lwk, so I am just nitpicking here, but in other situations or if bdk/lwk at some point doesn't automatically recreate the files anymore, and creating and using a wallet need separate actions regarding the creating of the folders/files, it might give problems then)

kumulynja (changes requested):

lib/core/wallet/data/datasources/bdk_facade.dart
ethicnology: [Facade is a structural design pattern that provides a simplified interface to a library, a framework, or any other complex set of classes.](https://refactoring.guru/design-patterns/facade)

This PR didn't introduce the pattern. When I was working on Background Tasks, I needed bdk functions that was available only in the datasource, without the need of the datasource itself. This can be refactored but this file is harmless

lib/core/wallet/data/datasources/bdk_facade.dart
ethicnology: will do

lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
ethicnology: related to previous comment, i will refactor that

lib/core/wallet/data/repositories/wallet_repository.dart
ethicnology: I'm taking the opposite angle:

Deleting from our environment (sqlite) is less likely to fail than an external environment.

My fear is considering that we cleared the lwk/bdk database while it failed. Since we need the wallet metadata to perform the deletion by getting the wallet db path, we could re-attempt if we still have the metadata.

If we don't have the metadata anymore the operation will be more costly in time since we will have to look up all bdk/lwk database to find the proper wallet

lib/core/wallet/data/datasources/bdk_facade.dart
ethicnology: cc360506e4a75b3366c797aff90b157e8067f6b7

lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
ethicnology: cc360506e4a75b3366c797aff90b157e8067f6b7

lib/core/wallet/data/datasources/lwk_wallet_datasource.dart
ethicnology: cc360506e4a75b3366c797aff90b157e8067f6b7

lib/core/wallet/data/datasources/lwk_facade.dart
ethicnology: cc360506e4a75b3366c797aff90b157e8067f6b7

lib/core/wallet/data/datasources/bdk_facade.dart
kumulynja: > [Facade is a structural design pattern that provides a simplified interface to a library, a framework, or any other complex set of classes.](https://refactoring.guru/design-patterns/facade)

Yes, the BDK and LWK's own public API is their facade as a library already. In the context of us using bdk or lwk, we don't have to create a facade for BDK/LWK itself anymore, we have to create the boundaries and facades of our own system. But yes, it is harmless, I just mentioned it to avoid we start calling every interface a facade now.

lib/core/wallet/data/repositories/wallet_repository.dart
kumulynja: "Less likely" is not a good measure imo. We should do things in a way that even if there is a slight chance of failure, we make sure that failure doesn't break anything.

> If we don't have the metadata anymore the operation will be more costly in time since we will have to look up all bdk/lwk database to find the proper wallet

The important thing here is not to compare based on operation cost, but to avoid failures because of missing data.
It is similar to making sure we don't have wallet metadata of a local signer anymore if the mnemonic is already deleted in secure storage, because:

A) Wallet metadata deleted, mnemonic still present => NO PROBLEM, wallet will not be shown/used anymore as we wanted since we wanted to delete it  👍 
B) Mnemonic already deleted, wallet metadata still present => ERROR! Can't find mnemonic for existing wallet.

(B can break the app if not properly catched, A will never break the app since there will never be that error)

The same would be true in case deleting the bdk/lwk data was as sensitive and unrecoverable as deleting a mnemonic, which of course it isn't at this point, but since we might add other types of wallets later in which there might be data that is unrecoverable if deleted, it's better to set the right example already.

To delete db/wallet data for wallets that are not in the wallet metadata anymore, we could have a small task remove them in which I really doubt performance to look them up is that important.

lib/core/wallet/data/datasources/bdk_facade.dart
ethicnology: > I just mentioned it to avoid we start calling every interface a facade now.

`Facade` is a common design pattern to wrap/expose libraries or a subset of functions from dependencies. It can be used to avoid the propagation of the external library inside the codebase to easily swap it. Likely useless for bdk / lwk thought. We are using various naming to achieve this in the codebase Service / Static classes etc…

> I just mentioned it to avoid we start calling every interface a facade now

We should discuss and agrees on the design patterns we want to allow and enforce accross the codebase to maintain it's consistency rather than dealing with heterogeneous architecture

lib/core/wallet/data/repositories/wallet_repository.dart
ethicnology: Good to mention that the mnemonic is not deleted from the secure storage anyway

lib/core/wallet/data/datasources/bdk_facade.dart
kumulynja: > Facade is a common design pattern to wrap/expose libraries

Yes, the library's public api itself so it can be used by others. So we, as users, don't have to create a facade of a facade without adding our own concepts to it. And here it is clearly just a factory to create instances.

> It can be used to avoid the propagation of the external library inside the codebase to easily swap it.

That would be an abstract class/interface, no facade. A facade is concrete, it's not to swap things out easily.
To easily swap things out and not have tight coupling between external libraries and our business logic, abstract classes/interfaces have to be used, not facades.

> We are using various naming to achieve this in the codebase Service / Static classes etc

I have no idea of that. Again, the only way to swap things out easily and reduce tight coupling is by defining interfaces that abstract things away from the dependencies into things that are useful and have meaning for our system/business logic. And from what I know of our codebase those are not Static classes and I haven't seen them being named Services either. We used to have datasource and repository abstract classes for that, never anything else from what I know.

lib/core/wallet/data/repositories/wallet_repository.dart
kumulynja: Yes, that was just a clearer/realer example of the problem this order of execution could cause if the deleted data is critical or unrecoverable, which for bdk/lwk data is not the case, but in the future for other wallet types or if bdk/lwk internals change, it can be.

PR #1676, opened 2025-12-18
Allow bitcoin-bitcoin internal transfer + Coin selection in transfer flow

i5hi:
- [x] Allow user to select bitcoin wallets for both From and To wallets in tranfer
- [x] This should trigger a standard bitcoin transfer instead of a swap
- [x] Add Advanced Options for both swaps and regular onchain transactions
- [x] Allow settings fees for regular onchain transactions
- [x] Minor fixes in color scheme for light mode in transfer flow (only in the feature, no fixes to color scheme class)

### DIFF
--- a/lib/core/widgets/coin_selection_bottom_sheet.dart
+++ b/lib/core/widgets/coin_selection_bottom_sheet.dart
@@ -11,14 +11,14 @@ import 'package:flutter/material.dart';
 import 'package:gap/gap.dart';
 import 'package:go_router/go_router.dart';
 
-class CommonCoinSelectionBottomSheet extends StatelessWidget {
+class CommonCoinSelectionBottomSheet extends StatefulWidget {
   final BitcoinUnit bitcoinUnit;
   final double exchangeRate;
   final String fiatCurrency;
   final List<WalletUtxo> utxos;
-  final List<WalletUtxo> selectedUtxos;
+  final List<WalletUtxo> initialSelectedUtxos;
   final int amountToSendSat;
-  final Function(WalletUtxo) onUtxoSelected;
+  final Function(List<WalletUtxo>) onDone;
 
   const CommonCoinSelectionBottomSheet({
     super.key,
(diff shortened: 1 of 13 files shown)


lib/core/widgets/coin_selection_bottom_sheet.dart
kumulynja, line 1: Would be better to make this a StatefulWidget and keep the selected utxos and total amount in the widget state and only passing a list of utxo's to the bloc when 'Done' is pressed.

lib/features/swap/presentation/transfer_event.dart
```diff
@@ -17,4 +17,13 @@ sealed class TransferEvent with _$TransferEvent {
       TransferExternalAddressChanged;
   const factory TransferEvent.receiveExactAmountToggled(bool enabled) =
       TransferReceiveExactAmountToggled;
+  const factory TransferEvent.replaceByFeeChanged(bool replaceByFee) =
+      TransferReplaceByFeeChanged;
+  const factory TransferEvent.utxoSelected(WalletUtxo utxo) =
```
kumulynja: Would be better to pass a list and only emit this event when 'Done' is called in the coin selection widget.

lib/features/swap/presentation/transfer_bloc.dart
```diff
@@ -732,6 +826,171 @@ class TransferBloc extends Bloc<TransferEvent, TransferState> {
...
+  ) async {
+    final selectedUtxos = List.of(state.selectedUtxos);
+    if (selectedUtxos.contains(event.utxo)) {
+      selectedUtxos.remove(event.utxo);
+    } else {
+      selectedUtxos.add(event.utxo);
+    }
+    emit(state.copyWith(selectedUtxos: selectedUtxos));
+    await _rebuildTransaction(emit);
+  }
```
kumulynja: This naming of the event and what it does is confusing and is a clear example of why mixing UI state and BLoC state is not a good idea. StatefulWidgets are really nothing bad if used for what they are needed.

The name of the event is onUtxoSelected, but if the utxo is already in the list of the bloc state, it actually doesn't select a utxo, it removes the utxo. Mimicking the behavior of a radio button or check box. But, a radio button or check box is UI and so this behavior is already done by the UI, it should not be mimicked in the BLoC again. The way to do that is by not triggering a BLoC event directly on an onChanged method of a UI component like a radio button if there is also a 'Done/Continue' button.
Only passing a list of selected Utxo's to the BLoC event when 'Done' is pressed is also more efficient since _rebuildTransaction is only called then instead of on every utxo toggle.

kumulynja (changes requested):

kumulynja:
My comments above are not blocking for merging this btw, would just be improvements if possible.

PR #1688, opened 2025-12-22
feat: custom mempool url

basantagoswami:


lib/core/storage/database_seeds.dart
```diff
@@ -115,4 +115,42 @@ class DatabaseSeeds {
           ),
         );
   }
+
+  static Future<void> seedDefaultMempoolServers(SqliteDatabase db) async {
+    final serversData = [
```
i5hi, line 120: These should be in the contants file and read here.

lib/core/utils/constants.dart
```diff
@@ -60,14 +60,16 @@ class PayjoinConstants {
 class ApiServiceConstants {
   // Bitcoin mempool
   static const bbMempoolUrlPath = 'mempool.bullbitcoin.com';
-  static const publicMempoolUrlPath = 'mempool.space';
+  static const publicMempoolUrlPath = 'mempool.space'; // note: not used
```
i5hi, line 63: lets remove it if we arent using it.

lib/core/utils/constants.dart
```diff
@@ -60,14 +60,16 @@ class PayjoinConstants {
...
   static const testnetMempoolUrlPath = 'mempool.space/testnet';
 
   // Liquid mempool
   static const bbLiquidMempoolUrlPath = 'liquid.bullbitcoin.com';
   static const bbLiquidMempoolTestnetUrlPath = 'liquid.bullbitcoin.com/testnet';
-  static const publicLiquidMempoolUrl = 'https://liquid.network';
-  static const publicLiquidMempoolTestnetUrl = 'https://liquid.network/testnet';
+  static const publicLiquidMempoolUrl =
```
i5hi, line 69: same here

lib/core/storage/migrations/schema_10_to_11.dart
```diff
@@ -93,5 +94,13 @@ class Schema10To11 {
...
           }),
         );
+
+    // Add mempool tables
+    await m.createTable(schema11.mempoolServers);
+    await m.createTable(schema11.mempoolSettings);
+
+    // Seed default mempool data
+    await DatabaseSeeds.seedDefaultMempoolServers(db);
+    await DatabaseSeeds.seedDefaultMempoolSettings(db);
```
ethicnology: The seed function should only be used in the initial migration.
In the migration from 10 to 11, it shouldn’t be used, since the schema may evolve.
Data should instead be inserted using the manager of this specific version

PR #1701, opened 2025-12-24
use BLoC event transformer to ignore new confirmations while one is processing

kumulynja:
Added the [bloc_concurrency](https://pub.dev/packages/bloc_concurrency) package to solve this with a "quick" but definitive fix. This doesn't mean we shouldn't solve the underlying cause of the state variable being used wrongly in the confirmation event handler anymore, but using the bloc_concurrency package here and in other BLoCs is a great first line of defense to avoid concurrency issues even if the underlying code isn't concurrency safe yet.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -172,6 +172,14 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "9.1.0"
+  bloc_concurrency:
+    dependency: "direct main"
+    description:
+      name: bloc_concurrency
+      sha256: "86b7b17a0a78f77fca0d7c030632b59b593b22acea2d96972588f40d4ef53a94"
+      url: "https://pub.dev"
+    source: hosted
+    version: "0.3.0"
   blockchain_utils:
     dependency: transitive
     description:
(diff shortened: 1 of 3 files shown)


pubspec.lock
```diff
@@ -306,10 +314,10 @@ packages:
     dependency: transitive
     description:
       name: camera_avfoundation
-      sha256: "0efb057a1fecdbf9b697272fbf79afbd47ac0e7bd69b4d900d3f304b31d93bad"
+      sha256: "087a9fadef20325cb246b4c13344a3ce8e408acfc3e0c665ebff0ec3144d7163"
       url: "https://pub.dev"
     source: hosted
-    version: "0.9.22+7"
+    version: "0.9.22+8"
```
ethicnology, line 320: I hope this build change doesnt fuck us like last time 🤣

kumulynja: Oh, is this something I should reverse maybe? I just did a `make setup` to start clean and it probably updated this automatically.

kumulynja: Or fix it on a specific version that we know works? @ethicnology

ethicnology: In theory it shouldn't be dangerous, but I don't know if you remember once this camera package fucked us on a build bump.

No worries we should check if camera still works after

PR #1717, opened 2025-12-30
feat: add announcements feature to exchange module

mocodesmo:
- Introduced GetAnnouncementsUsecase for fetching announcements.
- Updated ExchangeLocator to register the new use case.
- Implemented listAnnouncements method in ExchangeUserRepository.
- Added API call in BullbitcoinApiDatasource to retrieve announcements.
- Enhanced ExchangeCubit to load and manage announcements state.
- Integrated AnnouncementBanner in ExchangeHomeScreen for display.


lib/core/exchange/data/repository/exchange_user_repository_impl.dart
```diff
@@ -90,4 +92,36 @@ class ExchangeUserRepositoryImpl implements ExchangeUserRepository {
...
+
+  @override
+  Future<List<Announcement>> listAnnouncements() async {
+    try {
+      final apiKey = await _bullbitcoinApiKeyDatasource.get(
+        isTestnet: _isTestnet,
+      );
+      if (apiKey == null) {
+        throw ApiKeyException(
+          'API key not found. Please login to your Bull Bitcoin account.',
```
i5hi, line 104: Text should go in the localizations files

ethicnology: I'm not sure but I think this is an error that doesn't reach the user level, so only in logs

mocodesmo: none of the APIKeyExceptions are localised (35 occurrences)  . 
@i5hi  pls confirm what to do here

PR #1727, opened 2026-01-05
Exchange settings

mocodesmo:
## 1. New Exchange Settings Menu Items

The Exchange Settings screen now includes several new functional menu entries (previously commented out/placeholder):

- **Recipients** — Now navigates to a fully functional Recipients screen using the existing `RecipientsScreen` and `RecipientsBloc`
- **Transactions** — Now shows a dedicated exchange transactions view with buy/sell filtering
- **Default Bitcoin Wallets** — Fully functional wallet address management
- **App Settings** — User preferences (language, currency, email notifications)
- **Secure File Upload** — KYC document upload with status indicators
- **Statistics** — New screen showing trading statistics

---

## 2. Default Bitcoin Wallets Screen

**File:** `lib/features/settings/ui/screens/exchange/bitcoin_wallets_screen.dart`

Complete redesign with functional wallet management:

- **Description text** explaining the purpose
- **Three wallet address fields**: Bitcoin, Lightning, Liquid
- **Edit/Save/Cancel workflow** for each address type
- **Delete functionality** per address
- **Address truncation** for display (shows first 10 and last 8 characters)
- **Success/Error snackbars** for save operations
- **Loading and error states** with retry button

---

## 3. App Settings Screen

**File:** `lib/features/settings/ui/screens/exchange/app_settings_screen.dart`

Enhanced with new functionality:

- **Email Notifications Toggle** — New switch with label and description to enable/disable email notifications
- **Improved styling** — Updated colors, shadows, and typography
- **Auto-refresh** on screen load via `initState`
- Fixed enum references (e.g., `TextAlign.center`, `SnackBarBehavior.floating`)

---

## 4. Secure File Upload Screen

**File:** `lib/features/settings/ui/screens/exchange/file_upload_screen.dart`

Completely redesigned with functional upload capability:

- **Upload option card** similar to BB-Exchange styling
- **Status indicators**:
  - `Upload` (ready state with upload button)
  - `In Review` (warning-styled badge)
  - `Accepted` (success-styled badge)
- **Error display** with dismiss button
- **Loading states** during user fetch and file upload
- **Success snackbar** on successful upload
- **Disabled state** when file is already uploaded or under review

---

## 5. New Statistics Screen

**File:** `lib/features/settings/ui/screens/exchange/statistics_screen.dart`

Brand new screen showing user trading statistics:

- **Order Statistics Section**:
  - Buy/Sell Ratio card
  - Buy Statistics (volume, trade count, average price)
  - Sell Statistics (volume, trade count, average price)
  - Total Trading Volume
- **Biller Statistics Section** (when available):
  - Stats by payment method/biller
  - Total amount and trade count per biller
- **Loading, error, and empty states**
- **"As of" timestamp** showing when data was last updated
- **Formatted numbers** (K/M suffixes for large values)

---

## 6. Transactions Screen Enhancements

**Files:**
- `lib/features/settings/ui/screens/exchange/transactions_screen.dart`
- `lib/features/transactions/presentation/blocs/transactions_cubit.dart`
- `lib/features/transactions/presentation/blocs/transactions_state.dart`
- `lib/features/transactions/ui/widgets/txs_filter_row.dart`

The exchange transactions screen now:

- Uses the existing `TransactionsScreen` with an `exchangeOnly` mode
- **Filters limited to**: All, Sell, Buy (hides Send, Receive, Swap, Payjoin filters)
- Only shows order transactions (no wallet transactions)

---

## 7. Recipients Screen

**File:** `lib/features/settings/ui/screens/exchange/recipients_screen.dart`

Now fully functional using the existing recipients infrastructure:

- Reuses `RecipientsScreen` widget
- Integrates `RecipientsBloc` for data management
- Allows navigation to pay flow when recipient is selected

---

## 8. New Localization Strings

**File:** `localization/app_en.arb`

~40+ new localization keys added for:

- Email notifications labels/descriptions
- Bitcoin wallets screen texts
- File upload statuses and messages
- Statistics screen labels (order stats, biller stats, volume, trade count, etc.)
- Generic `retry` button label

---

## Navigation Flow

```
Exchange Settings
├── Recipients → Full recipients list with pay flow
├── Transactions → Exchange-only transactions (buy/sell orders)
├── Default Bitcoin Wallets → Manage BTC/Lightning/Liquid addresses
├── App Settings → Language, Currency, Email Notifications
├── Secure File Upload → KYC document upload with status
├── Statistics → Trading statistics dashboard
├── Referrals → (existing)
├── Security → (existing)
├── Account → (existing)
└── Log Out → (existing)
```

---

### DIFF
--- a/lib/router.dart
+++ b/lib/router.dart
@@ -161,7 +161,6 @@ class AppRouter {
       ...ImportQrDeviceRouter.routes,
       RecoverBullRouter.route,
       RecoverBullGoogleDriveRouter.route,
-      RecipientsRouter.route,
       LabelsRouter.route,
       StatusCheckRouter.route,
     ],


lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
```diff
@@ -585,6 +588,236 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
     return resp.data['result'] as Map<String, dynamic>;
   }
 
+  // ==================== Recipients API ====================
+
+  /// List recipients with optional filters for default wallets
+  Future<List<Map<String, dynamic>>> listMyRecipients({
```
kumulynja, line 604: Why repeat this here if the recipients feature already implements this?
Features should be self-contained, putting everything in core was a mistake and we should not continue doing that.

lib/features/settings/ui/screens/exchange/recipients_screen.dart
```diff
@@ -1,30 +1,25 @@
...
-          onBack: () => context.pop(),
-        ),
-      ),
-      body: SafeArea(
-        child: Padding(
-          padding: const EdgeInsets.all(16.0),
-          child: Center(child: Text(context.loc.exchangeRecipientsComingSoon)),
-        ),
-      ),
+    return BlocProvider<RecipientsBloc>(
```
kumulynja: Why wrap the Recipients screen from the recipients feature in a screen in settings instead of just navigating to the recipients feature itself?

kumulynja (changes requested):
Quickly saw these two things already, should review the rest still too.

lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
ethicnology: This should be discussed and planned because in the current organization `data` and `domain` layers are stored in a `lib/core/<feature>` folder while `presentation`/`ui` layer is store in `lib/features/<feature>`

I do not oppose on the idea of merging the two feature folders together, as it is kind of redundant, if we all agree on that, but I suggest we discuss and plan this change in a dedicated PR

cc: @i5hi @basantagoswami your thoughts on this, should we plan this refactoring during this sprint?

lib/features/settings/ui/screens/exchange/recipients_screen.dart
mocodesmo: @kumulynja 

Why This Pattern?

Separation of concerns: The TransactionsScreen and RecipientsScreen are pure UI widgets—they just render whatever state the Bloc provides. They don't know or care about context.

Context-specific business logic: Different entry points need different configurations (exchange-only vs all txs, different callbacks for selection).

Avoids route parameter complexity: Instead of passing complex objects/callbacks through route parameters (which is awkward with GoRouter), each entry point wraps the screen with its own configured Bloc.

Reusability without duplication: The UI code lives in one place; only the Bloc wiring differs.


This is a standard pattern in Flutter Clean Architecture where the feature's internal screen is "dumb" and the caller decides how to configure the business logic layer.

lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
mocodesmo: @ethicnology @i5hi @kumulynja  
what's the plan here ? 
is the refactor necessary for this specific pr ?
let me know if this needs changing thanks

lib/features/settings/ui/screens/exchange/recipients_screen.dart
mocodesmo: pls advise if we want to go with "just navigating to the recipients feature itself"

lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
i5hi: yes; we should be reusing `features/recipients` and create a facade that exposes methods needed by other features. 
In the features that need to use these methods we should create a `port` that can import the recipients facade. Then in your usecase in the exchange settings feature you can import that `port`

kumulynja: @i5hi that's indeed one way to integrate between features, but the facade is more if you want to call into other features from the usecases/adapters.
If you want to integrate UI, you can do that directly either by navigating or by wrapping the screens/widgets of other features, since this doesn't really couple anything in the business logic, since it is still the logic of the feature that owns the UI/BLoC that is being called into independent of which feature navigated to it or wrapped it.

lib/features/settings/ui/screens/exchange/recipients_screen.dart
kumulynja: > each entry point wraps the screen with its own configured Bloc.

> Reusability without duplication: The UI code lives in one place; only the Bloc wiring differs.

Yes, this is true, but it isn't done like this, you are reusing the same `RecipientsBloc` from the recipients feature, so you are not wiring any different BLoC and so it is not different from just navigating to the recipients feature route in which that same `RecipientsBloc` and `RecipientsScreen` are used already in the same way.

But, I agree that passing the recipients filters and `onSelected` callback through the route is not optimal and it was a quick fix because of time constraints at the moment.

The best solution would be a mix of both ways, not navigating, but wrapping the screen as you did, but still reusing most of the `RecipientsBloc` and only for the `onSelected` callback use the BLoC of the feature that is using the screen. In more details, like this:

1. Add the filter and `onSelected` callback as params of the `RecipientsScreen`
2. Have the reusable `RecipientsScreen` wrap the reusable `RecipientsBloc` in the recipients feature and pass the filter from the screen param to the BLoC and use the `onSelected` callback param instead of adding the `RecipientsSelected` event to the BLoC. The `RecipientsSelected` event and handler, as well as the callback param can be completely removed from the `RecipientsBloc` now.
3. Remove the route in the recipients feature and indeed let every feature wrap the recipients screen into its own screen, passing the appropriate filter and passing a function from it’s own BLoC to handle the selected recipient. This indeed gives more control for the calling feature to control its own logic and flow, while still reusing as much as possible from the UI and general recipients logic and UI.

kumulynja: I just did this and also from the settings it just navigates to the pay feature without wrapping the recipients screen again there, as I saw it was navigating to it already after the recipient selected. Withdraw and Pay themselves do not use navigation anymore, but wrap the RecipientsScreen as you started.

kumulynja: It would be good to still check what recipients related stuff can be removed from the core exchange datasource if it is already done by the recipients feature, and also move new recipients features functionality from the core exchange datasource to the recipients feature as well. After that I think we can merge this already.

PR #1750, opened 2026-01-08
Double payment prevention

i5hi:
If the send bloc has a txid (only possible if a broadcast is successful) - do not allow a successive broadcast. This will only allow 1 transaction to be broadcast in a single send flow.


kumulynja (approved):
LGTM! 👍

PR #1755, opened 2026-01-08
fix: various dark theme patches breaking PSBT flow

ethicnology:
- [fix: QR color background must be hardcoded in white since most device (ColdCard) can only read black QR on white background](https://github.com/SatoshiPortal/bullbitcoin-mobile/commit/f169eda539705035c438ebb8dc91dc0782cbd671)
- fix: hide instruction when they are empty + theme color for the button in Scan PSBT
- fix: custom fee theme

### DIFF
--- a/lib/features/psbt_flow/show_animated_qr/show_animated_qr_widget.dart
+++ b/lib/features/psbt_flow/show_animated_qr/show_animated_qr_widget.dart
@@ -123,7 +123,7 @@ class _ShowAnimatedQrViewState extends State<_ShowAnimatedQrView> {
                 data: state.parts[state.currentIndex],
                 // This color must stay hardcoded and not adapt to the theme.
                 // ColdCard can read only QR with white background.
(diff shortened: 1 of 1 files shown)


lib/features/psbt_flow/show_animated_qr/show_animated_qr_widget.dart
```diff
@@ -121,8 +121,9 @@ class _ShowAnimatedQrViewState extends State<_ShowAnimatedQrView> {
             children: [
               QrImageView(
                 data: state.parts[state.currentIndex],
-                // ignore: deprecated_member_use
-                foregroundColor: context.appColors.secondary,
+                // This color must stay hardcoded and not adapt to the theme.
+                // ColdCard can read only QR with white background.
+                backgroundColor: Colors.white,
```
i5hi: use a *fixed color from the theme

PR #1760, opened 2026-01-09
25-01-09_Update app_de.arb

bsn21m:
25-01-09_Update app_de.arb
Update into the German language.


bsn21m:
Hello, I created the JSON file because the translation in Weblate is not good. With JSON, I am more flexible and faster. Thank you.
Am 13. Jan. 2026, 15:05 +0100 schrieb Azad J. EMERY ***@***.***>:

PR #1775, opened 2026-01-12
Exchange - WebSockets instead of polling (for User and Chat updates)

mocodesmo:
---
### 🔔 1. Real-Time Notifications via WebSocket

**Before:** User data (balance, KYC status, groups) was polled every 5 seconds using a timer, causing unnecessary network requests and delayed updates.

**After:** Implemented WebSocket-based real-time notifications that push updates instantly when changes occur:

- **Balance updates** – User sees balance changes immediately
- **KYC status changes** – KYC approval/rejection reflected in real-time
- **Group membership updates** – Instant notification of user group changes
- **Support chat messages** – New messages appear without manual refresh

### 💬 2. Enhanced Support Chat Experience

**Real-time message delivery:**
- New chat messages from support are now pushed instantly via WebSocket
- No more manual refresh or polling required
- Messages appear immediately in the chat UI

**Improved attachment handling:**
- Cleaner image picker flow with proper permission handling
- Better error messages for permission issues:
  - "Permission denied." (temporary)
  - "Permission denied. Please grant photo library access in Settings." (permanent)
  - "Failed to pick files. Please try again." (general error)
- Streamlined file sharing functionality

### 🔄 3. Smarter Connection Lifecycle Management

**Login/Logout awareness:**
- WebSocket automatically connects when user logs in
- WebSocket automatically disconnects when user logs out
- No lingering connections or unnecessary reconnects

**Network environment switching (Mainnet ↔ Testnet):**
- WebSocket reconnects to the correct server when switching networks
- Seamless transition without manual intervention

**Auto-reconnect on disconnection:**
- If connection drops unexpectedly, auto-reconnects after 5 seconds
- Manual disconnects (logout) don't trigger auto-reconnect

### 🏗️ 4. Architecture Improvements (Supporting Better UX)

| Component | Change |
|-----------|--------|
| `ExchangeListener` | New widget wrapping the app to manage WebSocket connection lifecycle based on user authentication state |
| `ExchangeCubit` | Removed polling timer, added WebSocket connection management (`connectWebSocket`, `disconnectWebSocket`, `reconnectWebSocket`) |
| `ExchangeSupportChatCubit` | Listens to notification stream for instant message updates |

### 📁 New Files Added

| File | Purpose |
|------|---------|
| `exchange_listener.dart` | Manages WebSocket lifecycle based on login state |
| `exchange_notification_datasource.dart` | WebSocket connection handling |
| `exchange_notification_usecase.dart` | Environment-aware notification routing |
| `pick_image_attachments_usecase.dart` | Clean abstraction for image selection |
| `share_attachment_usecase.dart` | Clean abstraction for file sharing |
| `permission_datasource.dart` | Platform-specific permission handling |
| `image_picker_datasource.dart` | Image picker abstraction |
| `file_share_datasource.dart` | File sharing abstraction |

---

### ✨ Key User-Facing Benefits

1. **Faster feedback** – Balance and status updates appear instantly instead of up to 5 seconds later
2. **Real-time chat** – Support messages appear immediately without refreshing
3. **Less battery drain** – No more constant polling; WebSocket is more efficient
4. **Smoother UX** – Connection state is managed intelligently based on user actions
5. **Better error handling** – Clearer permission error messages guide users on how to fix issues

---

### DIFF
--- a/lib/core/exchange/exchange_locator.dart
+++ b/lib/core/exchange/exchange_locator.dart
@@ -37,9 +37,6 @@ import 'package:bb_mobile/core/exchange/domain/usecases/refresh_price_history_us
 import 'package:bb_mobile/core/exchange/domain/usecases/save_exchange_api_key_usecase.dart';
 import 'package:bb_mobile/core/exchange/domain/usecases/save_user_preferences_usecase.dart';
 import 'package:bb_mobile/core/exchange/domain/usecases/send_support_chat_message_usecase.dart';
-import 'package:bb_mobile/core/labels/data/label_datasource.dart';
-import 'package:bb_mobile/core/labels/data/label_repository.dart';
-import 'package:bb_mobile/core/labels/domain/batch_labels_usecase.dart';
 import 'package:bb_mobile/features/labels/labels_facade.dart';
 import 'package:bb_mobile/core/settings/data/settings_repository.dart';
 import 'package:bb_mobile/core/storage/data/datasources/key_value_storage/key_value_storage_datasource.dart';


lib/core/exchange/data/datasources/image_picker_datasource.dart
kumulynja, line 1: This is pure UI, this should not be in a datasource. Picking an image literally is an interaction with a UI. It should be done from a widget/screen and then the image data (PickedImage) should be passed to the BLoC and then the BLoC can use a usecase to send the image to wherever it has to be send to.

lib/core/exchange/data/datasources/file_share_datasource.dart
kumulynja, line 1: Same as with the image picker, this should not be a datasource, share_plus should be used in the screen/widget directly, since it literally is asking a UI interaction from the user.

lib/core/exchange/data/repository/exchange_notification_repository_impl.dart
kumulynja, line 1: This isn't a repository. Repositories are secondary/driven adapters, called upon by usecases.
This is a primary/driving adapter, since it gets websocket events and based on that it executes something.
I know we have made this mistake in other parts of the code as well to have websockets as datasources and passed through repositories, but we shouldn't make the same mistake with new code anymore.

Instead of this being a repository, this is more like a controller that gets events from the websocket and can then based on the event call the correct usecase show a notification or whatsoever.

lib/core/exchange/domain/usecases/pick_image_attachments_usecase.dart
kumulynja, line 1: This usecase should not exist, instead, there should be one to send an already picked image.

lib/core/exchange/domain/usecases/share_attachment_usecase.dart
kumulynja, line 1: Same as with the image picker, this usecase shouldn't exist, it should send an already picked file/file content.

lib/core/exchange/data/datasources/image_picker_datasource.dart
mocodesmo: when writing tests this will be hard to mock if the package is used directly in the bloc or a usecase

kumulynja: You should not use it in the bloc or usecase either. It is pure UI. And we should not test the package.

kumulynja: In the UI you use the package, you get the image data from it in the UI and you pass that through a BLoC function. Now that BLoC function calls a usecase to send that image to support.

kumulynja: The thing to test here is not the image picker. The thing to unit test would be the usecase to make sure the file is send correctly or the correct error is thrown if it was not possible to send the image.

mocodesmo: if we want to write widget tests
we wont be able to mock this if its coupled with the widget ui code

mocodesmo: the third party image picker package is the opposite of "pure ui"

mocodesmo: it involves native code how is it pure ui 

i suggest all external interaction like native code packages and third party packages always come
from a data source 

the widget, bloc and usecase layers should never have any direct comms with them

kumulynja: Why should it be mocked in a widget test? A widget/screen doesn't even have any DI, so what would you even mock there? If you want to test the rest of the widget without opening the picker UI, you just set a picked image in the widget state already. There is no need to mock this for widget tests.

mocodesmo: image picker package wont work in widget tests it needs to be mocked if we decide test coverage matters eventually

kumulynja: > it involves native code how is it pure ui

Where did you write any native code?
It's not because the package itself uses native code that what we use it for, which is letting our users pick a file/image, is not pure UI.

kumulynja: > i suggest all external interaction like native code packages and third party packages always come
from a data source

All Flutter widgets itself are third party packages...  it depends what you use the package for.

kumulynja: > image picker package wont work in widget tests it needs to be mocked if we decide test coverage matters eventually

And what would you test about it exactly? It is just a button that opens the image picker. You shouldn't test that. You can test the screen the button is on maybe, but then you can just set image data without the image picker.

mocodesmo: image picker package is mostly native code

kumulynja: So what? That doesn't matter. What are you using it for? That's what matters.

mocodesmo: write a widget test with image picker and see if it runs 
it wont
it needs to be mocked

kumulynja: Why would I test a button that opens image picker? What's there to test?

mocodesmo: we already have these layers implemented as a architecture pattern 
why would u couple the couple it on the ui layer

kumulynja: And you can still wrap it in another class/service that you can indeed mock and use that class/service directly in your UI. You don't need a datasource for that and definitely not pass through a usecase, repository etc for it.

kumulynja: > we already have these layers implemented as a architecture pattern

We used and abstracted the wrong layers a lot in the code base. We are starting to fix it all, so better to not add new code with the wrong patterns.

mocodesmo: so your argument is widget test coverage will never matter ? 

the reason for doing the rearchitecture project last year is so that every thing is uncoupled into layers and easier to test but u are suggesting we go against that principle and couple it 
why exactly?

kumulynja: > why would u couple the couple it on the ui layer

It is pure UI. Letting the user pick an image is something that litterally shows images on the screen and the user picks the images through an interface by pressing the screen to select the image, literally User Interaction with a User Interface. There are 0 business, orchestration or infrastructure interactions needed here. It's purely the user using an interface to select an image.

mocodesmo: the package runs alot of native code . widget test do not run native code. i do not want to add it to the ui layer as this is an anti pattern . and will need cleaning up down the road.
if you were talking about a pure dart flutter ui package then i would not argue but this definitely needs to go somewhere else.
let's have a meeting about this tomorrow

kumulynja: > so your argument is widget test coverage will never matter ?

That's not what I said, I said testing the image picker itself, or the button that opens the image picker is not a useful test. And you can still wrap the image picker in an own class if you need to mock it. It shouldn't be a datasource.

> the reason for doing the rearchitecture project last year is so that every thing is uncoupled into layers and easier to test but u are suggesting we go against that principle and couple it
why exactly?

Because the hard thing and the thing to get right is not just putting things into whatever layer like a datasource, the hard thing is to understand what every layer's responsability is and what is not and putting the right things in the right layers and having the correct dependency direction and flow between them is what matters. We didn't do all of that right from the first try last year. And it creates problems that we now with more understanding of architectural principles are busy to correct.

kumulynja: > the package runs alot of native code . widget test do not run native code. i do not want to add it to the ui layer as this is an anti pattern . and will need cleaning up down the road.

Wrong, the pattern of putting it in a datasource is the anti-pattern. Which will need to be cleaned up.

> if you were talking about a pure dart flutter ui package then i would not argue but this definitely needs to go somewhere else.

You know that dart is written in C++, right?
It doesn't matter anything if it has native code or not, what matters is what it does and what we use it for.

> let's have a meeting about this tomorrow

No thanks, I am tired of discussing.

mocodesmo: > Because the hard thing and the thing to get right is not just putting things into whatever layer like a datasource, the hard thing is to understand what every layer's responsability is and what is not and putting the right things in the right layers and having the correct dependency direction and flow between them is what matters. We didn't do all of that right from the first try last year. And it creates problems that we now with more understanding of architectural principles are busy to correct.

we regard decouple and then move things around rather than start with the wrong solution.

mocodesmo: > We didn't do all of that right from the first try last year. And it creates problems that we now with more understanding of architectural principles are busy to correct.

where the issue for this ? is it being discussed somewhere ?

kumulynja: You are not decoupling anything with this approach, since you are coupling your usecases to a filepicker and image picker instead of decoupling it of how they are picked by just passing the file/file content or image data in the usecase. That's the decoupling that matters.

kumulynja: > there is an interface (abstract) class and impl file .

So what? That doesn't mean anything, that's just a tool. I can use abstract classes and still couple things or decouple the wrong things unneccessarily.

mocodesmo: ok ill revert the chat feature to the code on develop and only make this pr about the websockets 

i thought i was cleaning things up

lib/core/exchange/domain/usecases/pick_image_attachments_usecase.dart
mocodesmo: reverting this back to develop branch implementation

lib/core/exchange/domain/usecases/share_attachment_usecase.dart
mocodesmo: reverting this back to develop branch implementation

lib/core/exchange/data/datasources/file_share_datasource.dart
mocodesmo: reverting this back to develop branch implementation

PR #1778, opened 2026-01-13
validate orbot availability before initiating recoverbull connection

### LINKED ISSUE #1691: Embedded Tor client initializes even when Orbot is running
From the output in my terminal:

```
I/flutter (27817): CONFIG	Starting Tor...
I/flutter (27817): INFO	Waiting for Tor to be ready...
x21
I/flutter (27817): CONFIG	Tor enabled
I/flutter (27817): INFO	Waiting for Tor to be ready...
x10
I/flutter (27817): CONFIG	Tor started
I/flutter (27817): CONFIG	Tor ready
I/flutter (27817): FINE	Tor started in 103s on port 38899
I/flutter (27817): FINE	Recoverbull server ready after 1 attempts
I/flutter (27817): FINE	Recoverbull server ready after 1 attempts
```

I had seen previously that Recoverbull didn't work when Orbot is running.

Output of `adb logcat -d | grep -i "tor\|recoverbull" | tail -100`:

```
11-27 19:14:35.449 21847 29570 D Orbot :
WARN: Not attempting connection to [scrubbed]:10444 because the network would reject it.
Are you trying to send Tor traffic over Tor? This traffic can be harmful to the Tor network.
If you really need it, try using a bridge as a workaround.
```

We had fixed it by only initializing embedded Tor when Orbot is not running. If it's running, we use it to route the Recoverbull traffic through Orbot

So I am not sure how this bug popped up or how it's working

basantagoswami:
Fixes #1691 

Before: Recoverbull checked the useTorProxy user setting to decide whether to use Orbot or start the embedded client. That worked only when the setting was enabled by user. If it was not enabled by the user, the app tried to start the embedded client and Orbot would block Tor over Tor connections

This PR fixes that issue, by checking if Orbot is actually running on the default port (9050)

This PR doesn't fix the case where Orbot is running in some custom port, and the user has not enabled the setting to use external Tor proxy, which is unlikely

Future work could be to somehow detect these Tor over Tor errors (if Orbot returns them), and display that to users


lib/core/tor/tor_locator.dart
```diff
@@ -40,5 +49,11 @@ class TorLocator {
     locator.registerFactory<TorStatusUsecase>(
       () => TorStatusUsecase(locator<TorRepository>()),
     );
+
+    locator.registerFactory<CheckTorProxyConnectionUsecase>(
```
kumulynja: UseCases shouldn't go in core. Usecases are for business/orchestration logic. Core is for shared infrastructure/simple primitives without any business logic.
`CheckTorProxyConnectionUsecase` feels like it should be part of the tor settings feature and nothing else. If other features need to know if tor is enabled and on which port, the tor settings feature's facade could return this info. 
The only thing about Tor that I feel should be in core is the embedded Tor client, more specifically a wrapper around it to start it and get info from it like the port (which can then be used in the adapters in other features that want to use tor). 
Maybe because of the current architecture it is difficult and there are usecases there already, but since this usecase was in features and is moved into core in this PR, it feels like a setback instead of an improvement.

kumulynja (approved):
Good catch and good fix! Tested with Orbot running without setting an external proxy from develop branch and indeed got an error, then tested with this branch with Orbot and RecoverBull worked 👍 

Just some structure and naming conventions that I would like to see changed, but that's a more general problem unrelated to this issue, so will be tackled in other PR's.
Just for sharing, this is how I think all Tor related things should be structured:

- Only thing in `core`: Some `factory` class to get a http client from a Tor proxy port, like the function that is currently in the tor_datasource in core. This can be used then by any adapter in any feature that needs tor and either the embedded tor client port or the external configured proxy port can be passed to it when needed.
- Just one `Tor` feature with:
    - Embedded Tor client wrapper that could be started and where the port can be returned from.
    - Management of the Tor settings like enabling/disabling the embedded client, the external proxy port etc.
    - Facade api to:
        - Get port if Tor is already available (either from running embedded client port or external proxy port). Can return null if none is available
        - Enable/disable embedded client from other features if returned port above is null.
        - Maybe a third function to get the Tor status with some more details and a real check if the connection is working, which can be useful for the Status feature.
- For the specific problem of the PR, the Tor settings feature should also add a `Port` (abstract class) to check for any external running Tor clients even if they are not set by the user. This can then be used in the usecases before starting the embedded client or saving the external proxy port to settings. It can be reused to both check any predefined default ports like 9050 before starting the embedded client, as well as to check the external port set by the user before saving it.
- Other features that need tor, like `Recoverbull` feature can then just get the proxy port to use from the `Tor` feature (so they all depend on Tor settings) and pass that port to whatever has a socks5 param, or use the Http client factory of core to pass the port and use that client for their calls. This would be done in the specific adapters of those features, like in a `RecoverBullGatewayPort` adapter for example. (It’s perfectly fine to use a facade of a feature it depends on in its own adapters, since it’s just another dependency from that features’ point of view)

PR #1784, opened 2026-01-15
BREAKING CHANGE: refactor labels with the new architecture

ethicnology:
### Schema Migration (v11 → v12)
**Labels table changes:**
- Add `id` autoincrement primary key
- Rename `ref` → `reference`
- Remove `spendable` column
- Add unique constraint on `(label, reference)`

### Key Changes
- **Domain**: `LabelEntity` with reference validators (tx, pubkey, input/output, xpub)
- **Facade**: Public API returning primitive types, uses `StoreLabelEnvelope` for input
- **Ports/Adapters**: Clean separation - usecases depend on abstract ports, not implementations
- **i18n**: Simplified export success message (removed singular/plural)

### Files
- `schema_11_to_12.dart` - Migration preserving existing data
- `label_entity.dart` - Domain entity with validators
- `labels_facade.dart` - Public API
- `*_port.dart` / `*_adapter.dart` - Hexagonal boundaries

### DIFF
--- a/lib/features/labels/domain/usecases/fetch_label_by_reference_usecase.dart
+++ b/lib/features/labels/domain/usecases/fetch_label_by_reference_usecase.dart
@@ -1,5 +1,4 @@
 import 'package:bb_mobile/features/labels/application/labels_repository_port.dart';
-import 'package:bb_mobile/features/labels/domain/label_entity.dart';
 import 'package:bb_mobile/features/labels/domain/label_error.dart';
 
 class FetchLabelByReferenceUsecase {
@@ -8,9 +7,10 @@ class FetchLabelByReferenceUsecase {
   FetchLabelByReferenceUsecase({required LabelsRepositoryPort labelRepository})
     : _labelRepository = labelRepository;
 
-  Future<List<LabelEntity>> execute(String reference) async {
+  Future<List<String>> execute(String reference) async {
     try {
-      return await _labelRepository.fetchByReference(reference);
+      final labels = await _labelRepository.fetchByReference(reference);
+      return labels.map((label) => label.label).toList();
     } on LabelError {
       rethrow;
     } catch (e) {
--- a/lib/features/labels/labels_facade.dart
+++ b/lib/features/labels/labels_facade.dart
@@ -27,8 +27,7 @@ class LabelsFacade {
        _deleteLabelUsecase = deleteLabelUsecase;
 
   Future<List<String>> fetchByReference(String reference) async {
-    final labels = await _fetchLabelByReferenceUsecase.execute(reference);
-    return labels.map((label) => label.label).toList();
+    return await _fetchLabelByReferenceUsecase.execute(reference);
   }
 
   Future<Set<String>> fetch() async =>


lib/core/exchange/domain/usecases/label_exchange_orders_usecase.dart
```diff
@@ -1,22 +1,16 @@
...
-import 'package:bb_mobile/core/labels/domain/batch_labels_usecase.dart';
-import 'package:bb_mobile/core/labels/domain/label.dart';
-import 'package:bb_mobile/core/labels/label_system.dart';
 import 'package:bb_mobile/core/utils/logger.dart';
+import 'package:bb_mobile/features/labels/labels_facade.dart';
 
 class LabelExchangeOrdersUsecase {
-  final LabelDatasource _labelDatasource;
-  final BatchLabelsUsecase _batchLabelsUsecase;
+  final LabelsFacade _labelsFacade;
```
kumulynja, line 7: You should not directly import a `Facade` here, as this tightly couples the "what" (adding a label) with the "how" (through the labels feature). A usecase should only get `Ports` injected, not concrete implementations, so you should create an `ExchangeLabelsPort` and implement that `Port` with the `LabelsFacade`. You could make the interface of the port something like `addLabelsForOrder` and then just pass the order to it, so in the concrete implementation of the `ExchangeLabelsPort` you can check the order type etc and decide which labels to add based on that. I would also move the system labels check and everything to it, as those are implementation details, the usecase just wants to make sure the order is labeled, it shouldn't know how exactly this is done.

ethicnology: This usecase is out of the scope as it belongs to the exchange feature. This PR focus on refactoring the labels folder

ethicnology: This usecase is out of the scope as it belongs to the exchange feature. This PR focus on refactoring the labels folder

lib/core/exchange/domain/usecases/label_exchange_orders_usecase.dart
```diff
@@ -1,22 +1,16 @@
 import 'package:bb_mobile/core/exchange/domain/entity/order.dart';
 import 'package:bb_mobile/core/exchange/domain/usecases/list_all_orders_usecase.dart';
-import 'package:bb_mobile/core/labels/data/label_datasource.dart';
-import 'package:bb_mobile/core/labels/domain/batch_labels_usecase.dart';
-import 'package:bb_mobile/core/labels/domain/label.dart';
-import 'package:bb_mobile/core/labels/label_system.dart';
 import 'package:bb_mobile/core/utils/logger.dart';
+import 'package:bb_mobile/features/labels/labels_facade.dart';
 
 class LabelExchangeOrdersUsecase {
```
kumulynja, line 6: I think this is not a good usecase, this labeling should be part of the usecase that creates the orders.
But I don't know if we should refactor that when we apply the new architecture guidelines to the exchange features and we should focus on the labels feature here.

lib/core/exchange/domain/usecases/label_exchange_orders_usecase.dart
kumulynja: Oh ok, just wrote another comment saying the same and only now see you responded haha.
Will focus only on the labels feature code itself.

lib/core/exchange/domain/usecases/label_exchange_orders_usecase.dart
ethicnology: The labeling is already part of the usecase that creates orders. This usecase is executed once the first time the user connect to the exchange to ensure it is labeling the orders that might have been created before the app was supporting the exchange.

The day we attack the exchange folder, let's rename it better!

LabelExistingExchangeOrdersUsecase ?

lib/core/exchange/domain/usecases/label_exchange_orders_usecase.dart
ethicnology: We will refactor this usecase in when we attack the exchange feature

lib/features/labels/domain/usecases/fetch_label_by_reference_usecase.dart
```diff
@@ -0,0 +1,22 @@
...
+import 'package:bb_mobile/features/labels/domain/label_entity.dart';
+import 'package:bb_mobile/features/labels/domain/label_error.dart';
+
+class FetchLabelByReferenceUsecase {
+  final LabelsRepositoryPort _labelRepository;
+
+  FetchLabelByReferenceUsecase({required LabelsRepositoryPort labelRepository})
+    : _labelRepository = labelRepository;
+
+  Future<List<LabelEntity>> execute(String reference) async {
```
ethicnology: here i shouldn't return an entity to avoid exposing it to the bloc ? @kumulynja

lib/features/labels/primitive/label_system.dart
kumulynja, line 1: This can just be moved to the domain folder. Primitives are part of the domain. No separate `primitives` folder needed.

lib/features/labels/primitive/label_type.dart
kumulynja, line 1: This can just be moved to the domain folder. Primitives are part of the domain. No separate `primitives` folder needed.

lib/features/labels/domain/usecases/fetch_label_by_reference_usecase.dart
ethicnology: refactored

lib/features/labels/domain/label_entity.dart
kumulynja, line 1: The LabelEntity is too linked to bip329 specs still, it is also unclear if it is a system label or not. This is something crucial for our app, since it is the business that wants us to label things automatically. So this should reflect in the domain and be more explicit instead of needing to parse the label itself (which isn’t even clear from looking at the domain if that’s the way to know). Introducing the Provenance concept would help. I will add a general comment at the end or in another issue as to how we could model the domain better.

lib/features/labels/domain/label_entity.dart
```diff
@@ -0,0 +1,97 @@
+import 'package:bb_mobile/features/labels/primitive/label_type.dart';
+import 'package:bs58check/bs58check.dart' as base58;
+import 'package:convert/convert.dart';
+
+class LabelEntity {
+  final int? id;
```
kumulynja, line 6: id should not be optional, one of the main characteristics of an entity is that it has an identity.

lib/features/labels/domain/usecases/delete_label_usecase.dart
kumulynja, line 1: I can not comment on a folder here, but the whole usecases folder should move to the application layer, since usecases are application logic, not domain logic.

lib/features/labels/application/labels_repository_port.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:bb_mobile/core/storage/sqlite_database.dart';
+import 'package:bb_mobile/features/labels/domain/label_entity.dart';
+
+abstract class LabelsRepositoryPort {
+  LabelsRepositoryPort({required SqliteDatabase database});
```
kumulynja, line 5: Do not specify a constructor here, an abstract class doesn’t need a constructor and especially not with an implementation detail like the SqliteDatabase as param.

lib/features/labels/application/store_label_model.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:bb_mobile/features/labels/domain/label_entity.dart';
+import 'package:bb_mobile/features/labels/primitive/label_type.dart';
+
+/// DTO for storing labels.
+///
+/// This DTO is used as input to the [StoreLabelsUsecase] to avoid exposing
+/// domain entities to outer layers. The use case is responsible for converting
+/// this DTO into the domain [LabelEntity] entity.
+class StoreLabelModel {
```
kumulynja, line 9: `StoreLabelModel` naming is a bit confusing since models are usually more related to frameworks in the Clean Architecture context, or models usually have behaviour (extra functions) as well, but here we just want plain data to cross the application layer border, so I wouldn’t call it a model. Shall we define to call it `ApplicationLabel` so the boundary is explicit in the naming and we don’t use Model or DTO or any other term that we don't like?

lib/features/labels/domain/usecases/delete_label_usecase.dart
```diff
@@ -0,0 +1,22 @@
...
+
+class DeleteLabelUsecase {
+  final LabelsRepositoryPort _labelRepository;
+
+  DeleteLabelUsecase({required LabelsRepositoryPort labelRepository})
+    : _labelRepository = labelRepository;
+
+  Future<void> execute({
+    required String label,
+    required String reference,
```
kumulynja, line 12: The delete label usecase should use the entity id imo instead of bip329 terms. It gives a cleaner api.

lib/features/labels/application/labels_repository_port.dart
```diff
@@ -0,0 +1,16 @@
...
+
+  Future<void> store(List<LabelEntity> labels);
+
+  Future<List<LabelEntity>> fetchByLabel(String label);
+
+  Future<List<LabelEntity>> fetchByReference(String reference);
+
+  Future<void> trashLabel({required String label, required String reference});
```
kumulynja, line 13: This should use the id instead of label or reference for a cleaner api.

kumulynja: That doesn't mean there can't be a trashByLabelAndReference function, but if that is not explicit, you would expect it is by id if it is about an entity like label.

lib/features/labels/domain/usecases/export_labels_usecase.dart
```diff
@@ -0,0 +1,24 @@
...
+  }) : _labelRepository = labelRepository,
+       _labelConverter = labelConverter;
+
+  Future<String> call(LabelFormat format) async {
+    final labels = await _labelRepository.fetchAll();
+    final formattedLabels = _labelConverter.convertTo(
+      format: format,
+      labels: labels,
+    );
+    return (formattedLabels as FormattedLabelsBIP329).jsonl;
```
kumulynja, line 22: TODO in future: check the actual label format that is passed and use the correct port for it. Best is to use a port registry for all different formats. But since now it will always be bip329 export, no need for it now though. Reminder to create an issue.

lib/features/labels/domain/usecases/fetch_label_by_reference_usecase.dart
kumulynja: That's right, we shouldn't leak entities here.

lib/features/labels/domain/usecases/fetch_distinct_labels_usecase.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:bb_mobile/features/labels/application/labels_repository_port.dart';
+import 'package:bb_mobile/features/labels/domain/label_error.dart';
+import 'package:bb_mobile/core/utils/logger.dart';
+
+class FetchDistinctLabelsUsecase {
```
kumulynja, line 5: This usecase's use is not very understandable. Is it even possible that two labels on the same target are the same if the constraints are correctly set in the db? And if this is for filtering for showing it in the UI, I don’t think this use case should even exist here. In that case, the presentation layer should do the filtering and decide which label of which target to show or to show that this label is set on both address as utxo for example. But this is not something to have a separate usecase for I feel.

lib/features/labels/domain/usecases/fetch_label_by_reference_usecase.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:bb_mobile/features/labels/application/labels_repository_port.dart';
+import 'package:bb_mobile/features/labels/domain/label_entity.dart';
+import 'package:bb_mobile/features/labels/domain/label_error.dart';
+
+class FetchLabelByReferenceUsecase {
```
kumulynja, line 5: This feels like a pure BIP329 usecase, better to have a getLabelById usecase now that we have id's.

kumulynja: Or at least have both and if the callers that need this have the id it would be better if they fetch by id if possible.

lib/features/labels/domain/label_error.dart
kumulynja, line 1: abel_error mixes UI build context with domain errors. This should be done in the UI layer. The domain errors should stay strictly domain related and shouldn’t even be the same errors exposed to the UI.

lib/features/labels/presentation/page.dart
kumulynja, line 1: Add page.dart into UI folder outside of presentation folder. We might create other screens and label related widgets that could be put in UI too then and pure UI is something that can be embedded in other features without problems, while the presentation folder should be more about the controllers (BLoC/Cubit) which I think shouldn't be used from other features.

lib/features/labels/domain/label_entity.dart
kumulynja, line 6: Also, the id is now added here, but it is not used anywhere, so this addition doesn't add what it should be adding at the moment, everything is still by reference. We need the facade and use cases to use id, not the bip329 reference, otherwise there is still no way to edit a label cleanly and easily

lib/features/labels/store_label_envelope.dart
```diff
@@ -0,0 +1,53 @@
...
+  }) {
+    return StoreLabelEnvelope(
+      type: LabelType.address,
+      label: label,
+      reference: address,
+      origin: origin,
+    );
+  }
+
+  StoreLabelModel toModel() {
```
kumulynja, line 45: This shouldn’t have a `toModel`behavior since this is supposed to be a plain data object used by other features that never need this. Better to define a mapper class for it or if you really want to have a function like this somewhere and avoid an extra class, then I think it is better on the `ApplicationLabel` class since it is more internal at least. Otherwise you are leaking application related behaviour in the public object.

lib/features/labels/store_label_envelope.dart
```diff
@@ -0,0 +1,53 @@
...
+  final LabelType type;
+  final String label;
+  final String reference;
+  final String? origin;
+
+  StoreLabelEnvelope({
+    required this.type,
+    required this.label,
+    required this.reference,
+    required this.origin,
```
kumulynja, line 14: Origin shouldn’t be required here since it is an optional field.

lib/core/exchange/domain/usecases/label_exchange_orders_usecase.dart
kumulynja: I was thinking some more about this and I actually don't even think this should be done from the exchange feature. In the end it is the labels feature that needs a migration of the labels, or a seeding in some sense of missed values. So I was thinking that we should actually remove this usecase and let the labels feature use the core BBX api client directly to fetch the orders and seed the labels in a real db migration/seeding in the labels feature. It is a pure migration (framework) thing from the labels, no usecase needed and certainly not in exchange features.

PR #1785, opened 2026-01-15
Architecture doc update

kumulynja:
In this first iteration I tried to share the concepts as understandable as possible, without too much bloat. We can add more concrete practices like naming terminology etc. as we go defining it more.
Also if the need to clarify concepts more in depth arises, we can also add it later.
I didn't add a migration from the old to new ways section yet, but I think the `Common Architectural Pitfalls` section might be sufficient and even better already to avoid previous mistakes from now on.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,147 +1,158 @@
 # 🏗 Architecture
 
-## 📁 Project structure
+This project follows a **feature-based** and **layered architecture** to ensure **separation of concerns**, **scalability** and **maintainability**.
 
-This project follows a **feature-based** and **layered architecture** to ensure **scalability**, **maintainability**, and **separation of concerns**.
+## 📦 Feature-Based Architecture
 
-The lib/ directory is organized into the following main sections:
+Features are self-contained modules that encapsulate all code related to a specific functionality, domain, or resource of the application.
 
-- [Core](lib/_core/): Contains fundamental **UI-independent** components that are **used across multiple features**. It has its own set of layers and a dependency locator:
-  - [Data](lib/_core/data/): Contains **data sources**, **models**, and concrete **repository and service implementations** that are shared across multiple features.
-  - [Domain](lib/_core/domain/): Contains **entities**, abstract **repository and service contracts**, and **use cases** that are shared across multiple features.
-  - [Presentation](lib/_core/presentation/): Contains **BLoCs** that are used in multiple features.
-  - [Core Locator](lib/_core/): Manages **dependency injection setup** for the core layer.
-- [UI](lib/_ui/): Contains **shared UI components**, like widgets, fonts, colors and themes that are used across multiple features.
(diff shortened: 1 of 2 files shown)


FEATURES.md
```diff
@@ -0,0 +1,192 @@
...
+   - API Clients (REST/GraphQL clients)
+   - Factory for a HTTP client to connect to Tor
+   - UI Kit (shared widgets, theme)
+   - DI setup and interfaces (Service Locator pattern)
+   - Router setup and interfaces (Navigation)
+   - PIN encrypted storage
+
+2. **Core Primitives**:
+   - Located in `/lib/core/primitives/`
+   - Examples: `Secret`, `SecretUsagePurpose`, `Fingerprint`, `Address`, `Amount`, etc.
```
ethicnology, line 151: let's keep the primitive inside the feature exported in the facade to avoid the replication of a feature tree into `core/primitives` or we let them all there without sub-tree ?

ethicnology: maybe it's better to let them all there without sub-tree to ensure no primitive duplicates and a quick lookup?

ARCHITECTURE.md
ethicnology, line 1: I feel we should add a graph on the communications within the same feature that include the naming convention of the interfaces and data structures?

FEATURES.md
kumulynja: Yes, keeping them in the `public` folder of the feature and exporting it together with the facade is maybe the best for things that still have a clear relation/owner with one specific feature, like `Secret`, `SecretUsagePurpose` etc.
And then things like `Amount` that are really general, just put those in `/core/primitives` directly without extra folder nesting?

ARCHITECTURE.md
kumulynja: Yeah, I thought of doing that once we have some features done and then link to them as examples.
For this initial draft I wanted to make sure the essence was better captured and the bloat and unneccessary patterns and mistakes of the previous architecture are gone already.

FEATURES.md
ethicnology: I suggest we commit strictly to one approach: either co-locating primitives within their specific features or centralizing them all in core. We should avoid mixing the two strategies to ensure the architectural rules remain easy to enforce.

Keeping primitives within features prevents the core/primitive folder from becoming cluttered, though it risks code duplication across different features. Moving everything to core solves the duplication issue but inevitably leads to a massive, centralized folder. We need to decide which trade-off we prefer

kumulynja: Yeah, thing is, in which feature would `Amount` go if we go with keeping them in features? That is so general and used app wide that it feels like a core thing haha, while Secret, SecretUsagePurpose etc clearly are strictly related to secrets. So I wouldn't mind making that the rule to enforce: 

a) Clearly related to a feature --> feature domain, exposed in public
b) Very general, not related to one feature in specific, just used everywhere as commons --> core/primitives

I don't see a reason why this couldn't be enforceable too.

ethicnology: Harder to enforce but it makes sense

i5hi (approved):

Issue #1810, opened 2026-01-20
Add page.dart into UI folder outside of presentation folder. We might create other screens and label related widgets that could be put in UI too then and pure UI is something that can be embedded in other features without problems, while the presentation folder should be more about the controllers (BLoC/Cubit) which I think shouldn't be used from other features.


ethicnology:

kumulynja:
> since we are not using any acronyms for folder name, I suggest we use user_interface rather than ui

I don't know about that, `ui` is very common to see, and easier to write. I really don't have any problem with acronyms that make sense and are clear. I mean, if the acronym is not common I agree it is better to be explicit, but I don't think anyone working on a mobile app will think UI means anything else.

Issue #1808, opened 2026-01-20
`FetchLabelByReferenceUsecase` feels like a pure BIP329 usecase, better to have a getLabelById usecase now that we have id's. Or at least have both and if the callers that need this have the id it would be better if they fetch by id if possible.


ethicnology:

ethicnology:
I can add the id version, this has nothing to do with Bip329 but with the unique constraint in our database that guarantee us that a single label exists per reference

Issue #1807, opened 2026-01-20
`FetchDistinctLabelsUsecase` is not very understandable. Is it even possible that two labels on the same target are the same if the constraints are correctly set in the db? And if this is for filtering for showing it in the UI, I don’t think this use case should even exist here. In that case, the presentation layer should do the filtering and decide which label of which target to show or to show that this label is set on both address as utxo for example. But this is not something to have a separate usecase for I feel.


ethicnology:

kumulynja:
Yeah, but I feel like it is not a thing to create a usecase for. This is more presentation/UI filtering/grouping after just quering all labels. Because then at least you have the info to show that this label is used on different things. If you have the usecase just returning strings, you don't know this and it doesn't make it very useful.

Issue #1805, opened 2026-01-20
The delete label usecase should use the entity id imo instead of bip329 terms. It gives a cleaner api.


ethicnology:
merging with:
>  trashLabel function in the repository should be by id. That doesn't mean there can't be a trashByLabelAndReference function, but if that is not explicit, you would expect it is by id if it is about an entity like LabelEntity.

kumulynja:
> These are the name of our columns in the database

So still domain based on the db (framework) and the db clearly build with only bip329 in mind.
Instead of defining the domain with useful concepts and identity of the entity and then defining your db schema to store the domain info.

> I can do the refactoring it will require to encapsulate the String in a class with the id

Not sure what you mean with this exactly. Which string do you mean to encapsulate?
The id is an `int` unrelated to the bip329 reference or any other string, no?

PR #1826, opened 2026-01-21
refactor: labels feature issues Closes #1795

### LINKED ISSUE #1795: Labels feature issues
- [ ] Move primitives to domain: Primitives are part of the domain. No separate primitives folder needed.
- [ ] The LabelEntity is too linked to bip329 specs still, it is also unclear if it is a system label or not. This is a custom concept for our app, since it is the business that wants us to label things automatically. So this should reflect in the domain and be more explicit instead of needing to parse the label itself (which isn’t even clear from looking at the domain if that’s the way to know). Introducing a `Provenance` concept would help.
- [ ] `id` should not be optional, but required, in the entity
- [ ] Also, the id was added, but it is not used anywhere, so this addition doesn't add what it should be adding at the moment, everything is still by reference. We need the facade and use cases to use id, not the bip329 reference, otherwise there is still no way to edit a label cleanly and easily.
- [ ] The whole usecases folder should move to the application layer, since use cases are application logic, not domain logic.
- [ ] Do not specify a constructor in the `LabelsRepositoryPort`, an abstract class doesn’t need a constructor and especially not with an implementation detail like the `SqliteDatabase` as param.
- [ ] `StoreLabelModel` naming is a bit confusing since models are usually more related to frameworks in the Clean Architecture context, or models usually have behaviour (extra functions) as well, but here we just want plain data to cross the application layer border, so I wouldn’t call it a model. Shall we define to call it `ApplicationLabel` so the boundary is explicit in the naming and we don’t use Model or DTO or any other term that we don't like?
- [ ] The delete label usecase should use the entity id imo instead of bip329 terms. It gives a cleaner api.
- [ ] `trashLabel` function in the repository should be by id. That doesn't mean there can't be a `trashByLabelAndReference` function, but if that is not explicit, you would expect it is by `id` if it is about an entity like `LabelEntity`.
- [ ] Set `ExportLabelUsecase` up correctly to support different export formats by actually checking the label format that is passed and use the correct port for it. Best is to use a port registry for all different formats.
- [ ] `FetchDistinctLabelsUsecase` is not very understandable. Is it even possible that two labels on the same target are the same if the constraints are correctly set in the db? And if this is for filtering for showing it in the UI, I don’t think this use case should even exist here. In that case, the presentation layer should do the filtering and decide which label of which target to show or to show that this label is set on both address as utxo for example. But this is not something to have a separate usecase for I feel.
- [ ] `FetchLabelByReferenceUsecase` feels like a pure BIP329 usecase, better to have a getLabelById usecase now that we have id's. Or at least have both and if the callers that need this have the id it would be better if they fetch by id if possible.
- [ ] `LabelError` mixes UI build context with domain errors. This should be done in the UI layer. The domain errors should stay strictly domain related and shouldn’t even be the same errors exposed to the UI.
- [ ] Add page.dart into UI folder outside of presentation folder. We might create other screens and label related widgets that could be put in UI too then and pure UI is something that can be embedded in other features without problems, while the presentation folder should be more about the controllers (BLoC/Cubit) which I think shouldn't be used from other features.
- [ ] The facade `Label` shouldn’t have a `toModel` behavior since this is supposed to be a plain data object used by other features that never need this. Better to define a mapper class for it or if you really want to have a function like this somewhere and avoid an extra class, then I think it is better on the ApplicationLabel class since it is more internal at least. Otherwise you are leaking application related behaviour in the public object.
- [ ] Origin shouldn’t be required in the facade `Label` since it is an optional field.

ethicnology:
Closes sub-issues:
    #1798
    #1802
    #1803
    #1809
    #1811
    #1812
    #1800
    #1801
    #1804
    #1805
    #1807
    #1808
    #1810
    #1806


lib/features/labels/domain/new_label_entity.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:bb_mobile/features/labels/domain/primitive/label_type.dart';
+
+class NewLabelEntity {
```
kumulynja: This is not an entity, since it has no identity (id) yet, this is a simple Value Object used to create entities from it. (sorry for the nitpicking, but removing the Entity-suffix from it is more correct)

lib/features/labels/application/labels_repository_port.dart
```diff
@@ -1,16 +1,18 @@
-import 'package:bb_mobile/core/storage/sqlite_database.dart';
 import 'package:bb_mobile/features/labels/domain/label_entity.dart';
+import 'package:bb_mobile/features/labels/domain/new_label_entity.dart';
 
 abstract class LabelsRepositoryPort {
-  LabelsRepositoryPort({required SqliteDatabase database});
+  Future<LabelEntity> store(NewLabelEntity newLabel);
```
kumulynja: I find `store` confusing since `patch` also "stores" the update. Why not the more common terms `create` or `add`?

kumulynja: It might be clear because of the New in NewLabel though, but it would be more clear with `create` or `add` imo

lib/features/labels/application/labels_repository_port.dart
```diff
@@ -1,16 +1,18 @@
...
+  Future<void> patch(LabelEntity label);
 
   Future<List<LabelEntity>> fetchByLabel(String label);
 
   Future<List<LabelEntity>> fetchByReference(String reference);
 
-  Future<void> trashLabel({required String label, required String reference});
+  Future<LabelEntity?> fetchById(int id);
+
+  Future<void> trashLabel(LabelEntity label);
```
kumulynja: It feels weird to have to pass the complete label to delete it, just by id is more common in a repository port.

lib/features/labels/application/usecases/delete_label_usecase.dart
```diff
@@ -7,12 +9,11 @@ class DeleteLabelUsecase {
   DeleteLabelUsecase({required LabelsRepositoryPort labelRepository})
     : _labelRepository = labelRepository;
 
-  Future<void> execute({
-    required String label,
-    required String reference,
-  }) async {
+  Future<void> execute(ApplicationLabel label) async {
```
kumulynja: I feel like just passing the id is enough and the usecase can be called `DeleteLabelByIdUseCase`, no need to pass the complete entity and do all the mappings through all layers. Just the id is enough to know which entity to delete.

lib/features/labels/application/labels_repository_port.dart
ethicnology: > I find store confusing since patch also "stores" the update
store –> insert
patch –> update

I prefer store which is explicitly related to storage rather than create/add that exist in various context helpers for lists etc

Also we can reduce to a single `store` function that update on conflict based on the id btw

lib/features/labels/application/usecases/delete_label_usecase.dart
ethicnology: 615ec839c2008abcaa30bae7a0f933c2a28f96e4

lib/features/labels/application/labels_repository_port.dart
ethicnology: 615ec839c2008abcaa30bae7a0f933c2a28f96e4

lib/features/labels/domain/new_label_entity.dart
ethicnology: refactored in 909b9ecf71f4142db6b89ca9178dbd5bb6d374f2 but now there is confusion between `NewLabel` as a value object in the domain and `NewLabel` as a public data structure for the facade

lib/features/labels/application/labels_repository_port.dart
kumulynja: > Also we can reduce to a single store function that update on conflict based on the id btw

Yeah, I would prefer this if possible with the new label not having an id yet?

lib/features/labels/domain/new_label_entity.dart
kumulynja: what about using native params instead of NewLabel in the facade?

lib/features/labels/application/labels_repository_port.dart
ethicnology: dd61bee9e66d1e97c4af2208327a0fce74bc39f9

`NewLabel` has now an optional `id` in case the new label is supposed to replace an existing one

lib/features/labels/domain/new_label_entity.dart
ethicnology: We can use native params, but it will be 3/4 params instead of a single struct

PR #1834, opened 2026-01-26
Add Bitaxe connect

anipy1:
This PR adds Bitaxe Connect feature. It lets you connect the app to a Bitaxe mining device on your local network. You enter the device IP, choose a Bitcoin wallet from the app, and the app configures the Bitaxe to mine to that wallet.

It is experimental, so the feature is hidden by default. The Connect Bitaxe option only appears in [`all_settings_screen.dart`](https://github.com/anipy1/bullbitcoin-mobile/blob/64905d9f3b44eef7f289a8f915aa0586a438bd02/lib/features/settings/ui/screens/all_settings_screen.dart#L110
) when superuser mode is enabled.


anipy1:
I'm closing this one as I built it mostly as a fun experiment, but thinking more about it, I don't think it earns its place in the app: AxeOS already ships a perfectly good interface for setting the mining address, and anyone running a Bitaxe is already comfortable configuring it there. Duplicating that flow in the wallet adds surface area (device discovery, IP handling, a superuser-gated screen) for something users can do in a browser tab.

Thanks for the time either way.

PR #1874, opened 2026-02-14
replace bdk_flutter with bdk_dart

### LINKED ISSUE #1958: Upgrade to BDK dart library

### LINKED ISSUE #1994: Migrate to latest `bdk_dart` with named parameters
Migrate to latest bdk-dart version which uses named parameters. Existing PR needs update to latest bdk-dart, then internal testing can begin. Goal is to prevent recurring electrum server issues.

- [x] Update PR to latest bdk-dart version (named parameters)
- [ ] Internal testing on staging

kumulynja:
This PR fully replaces `bdk_flutter` with `bdk_dart`. The migration is necessary because `bdk_flutter` is based on a pre-v1 version of `bdk` that contains several critical bugs affecting wallet functionality, while `bdk_dart` is up-to-date with `bdk` v2.3.0 at the moment.

The most critical bugs we have experienced in BBW that this migration aims to eliminate are related to address indexes being increased exponentially, which have been reported and fixed in newer BDK versions:
- https://github.com/bitcoindevkit/bdk/issues/701
- https://github.com/bitcoindevkit/bdk/issues/1578
- https://github.com/bitcoindevkit/bdk/issues/915

# Note on `bdk_dart` Maturity

Unlike the `bdk` Rust library, `bdk_dart` itself is relatively new. This PR serves as the initial integration to begin comprehensive testing of `bdk_dart` in BBW.

We can keep an eye on the state here and release accordingly after internal testing: https://github.com/bitcoindevkit/bdk-dart/issues/16

# Testing requirements

Wallet functionality and thus the use of `bdk_dart` is being dependent on by as good as all features, so everything should be tested. Critical areas to test specifically:

- Wallet creation and restoration
- Wallet synchronization and balance retrieval
- PSBT signing and finalization
- Transaction broadcasting
- Replace-by-fee (RBF) operations
- Address derivation and verify address index increments correctly
- Ledger hardware wallet integration
- BBQR
- Data migration from Hive to SQLite

# Primary success criteria:

- [ ] Address index exponential increment bug is resolved
- [ ] No new regressions introduced
- [ ] All existing features continue to work as expected


pubspec.lock
ethicnology, line 1: To protect this PR from being screwed by the update of a dependency unralted to the topic, you may discard the change on the `pubspec.lock` and `flutter pub get`

It should show the diff only related to `bdk_dart`, if no, you can select the line to be committed with something like Zed, or discard the changes unrelated to `bdk_dart`

lib/core/bbqr/bbqr.dart
```diff
@@ -88,8 +87,7 @@ class Bbqr {
 
   static Future<List<String>> splitPsbt(String psbt) async {
     try {
-      final bdkPsbt = await bdk.PartiallySignedTransaction.fromString(psbt);
-      final psbtBytes = bdkPsbt.serialize();
```
ethicnology, line 92: Note: that the psbt is no longer validated

kumulynja: Good point, I thought it was only used to get the bytes, but couldn't find the function for that in bdk_dart, so just did a base64.decode. I can still add back the bdk.Psbt(psbt) before doing the decode though.

lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
```diff
@@ -171,26 +169,39 @@ class BdkWalletDatasource {
...
-          .map((utxo) => bdk.OutPoint(txid: utxo.txId, vout: utxo.vout))
+          .map(
+            (utxo) => bdk.OutPoint(bdk.Txid.fromString(utxo.txId), utxo.vout),
+          )
           .toList();
       txBuilder.addUtxos(selectableOutPoints);
     }
-    if (replaceByFee) txBuilder.enableRbf();
+
+    // TODO: Investigate if bdk_dart always has enableRbf enabled by default as option doesn't seem to be exposed in the TxBuilder, if so remove the replaceByFee parameter and logic
```
i5hi: I think the way they do it is by directly letting you set `nSequence` which should be 0xfffffffd or lower for enabled

i5hi: https://github.com/bitcoindevkit/bdk-dart/blob/main/lib/bdk.dart#L22065C13-L22065C29

lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
```diff
@@ -255,27 +266,20 @@ class BdkWalletDatasource {
...
-      signOptions: const bdk.SignOptions(
-        trustWitnessUtxo: true,
-        allowAllSighashes: false,
-        removePartialSigs: true,
-        tryFinalize: true,
-        signWithTapInternalKey: false,
-        allowGrinding: true,
-      ),
+      psbt,
+      bdk.SignOptions(true, null, true, true, false, true),
```
i5hi: wish bdk-dart used named parameters for this :(

ethicnology: You are right, it's error prone to have all these boolean without named parameters.

If bdk-dart do not want to use named parameters, maybe we could make an interface over it ?

lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
ethicnology: they should expose [.enable_rbf()](https://docs.rs/bdk/latest/bdk/wallet/tx_builder/struct.TxBuilder.html#method.enable_rbf) which sets the `nSequence`

is there a particular reason why they don't expose the convenience methods from bdk ?

lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
reez: I agree. And we had one user request it previously https://github.com/bitcoindevkit/bdk-dart/issues/21

This is currently an upstream uniffi-dart limitation (named parameters aren’t generated yet)

It's actually an upstream uniffi-dart thing where that needs to be enabled there. I hadn't opened an issue on the uniffi-dart repo yet but just did https://github.com/Uniffi-Dart/uniffi-dart/issues/110 so thanks for the feedback

lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
reez: enable rbf is the default that's the only reason you're not seeing it 

bdk: https://github.com/bitcoindevkit/bdk/pull/1616

bdk-ffi: https://github.com/bitcoindevkit/bdk-ffi/pull/600

let me know if that works for ya'll or if there's any thing else needed related to this

ethicnology: Cool, good to know

pubspec.lock
kumulynja: Done, only kept bdk_dart diff.

lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
spacebear21: uniffi-dart now generates named parameters https://github.com/Uniffi-Dart/uniffi-dart/pull/112

reez: `bdk-dart` is now updated, just merged (breaking) changes into main https://github.com/bitcoindevkit/bdk-dart/pull/54

kumulynja: Just updated here too so we use bdk-dart with the named parameters.

ethicnology (approved):

kumulynja:
@reez I am seeing `Use bdk-ffi 3.0` in the production-ready checklist of bdk-dart now. Is going with 2.3 ok as well already, or is there any specific reason we better wait for 3.0?

reez:
> @reez I am seeing `Use bdk-ffi 3.0` in the production-ready checklist of bdk-dart now. Is going with 2.3 ok as well already, or is there any specific reason we better wait for 3.0?

yeah going with 2.3 is ok as well, waiting for 3.0 isn't a blocker, I'll update the wording on that item in the checklist

PR #1875, opened 2026-02-16
Handle lwk database corruptions due to parallel updates

i5hi:
UpdateOnDifferentStatus errors from lwk can only be recovered by deleting the lwk db and resyncing it. This issue deletes wallet on the balance and sync method if the error is detected.

The reason for doing it in the balance method is because it is called on app startup and reading the db at this time can cause an app startup failure.


pubspec.lock
ethicnology, line 1: To protect this PR from being screwed by a dependency update, and since there is no `pubspec.yaml` among the commits I recommend to discard this file, `pubspec.lock`

i5hi: done

ethicnology:
Should we replace `develop` by `init-error-screen-fss-fork` as a target for this PR ?

Issue #1891, opened 2026-02-26
LWK database corruption from background task race condition blocks app startup


thibistaken:
## Description

Corrupted LWK (Liquid Wallet Kit) database caused by background sync tasks prevents the app from starting. When two LWK sync instances run concurrently (e.g. a background task fires while the app is starting), they race to write to the DB, corrupting it. The app then shows a startup error and cannot be recovered by restarting the app or the device.

## Root Cause

- LWK sync background tasks can fire concurrently with app startup sync
- Two instances writing to the LWK DB simultaneously corrupts it
- App startup currently depends on LWK DB being in a good state, so corruption = app won't open
- Background tasks get cancelled and re-registered on each app relaunch, but timing is unpredictable
- Possible underlying SQLite non-concurrency issue (similar to BDK's one-db-per-wallet workaround)

## Impact

- **Critical**: User cannot open the app at all
- If user has no seed backup, they cannot access their funds until the bug is fixed
- Restarting app/device does not resolve the issue

## Acceptance Criteria

- [ ] Detect corrupted LWK DB on startup and automatically wipe/rebuild it
- [ ] App startup must NOT depend on LWK DB being in a valid state — a corrupted or missing LWK DB should never block the app from opening
- [ ] Prevent concurrent LWK sync (mutex/lock to ensure only one sync runs at a time)
- [ ] User must always be able to access their seed regardless of LWK DB state
- [ ] Background task logging captures any LWK sync failures for debugging

## Notes

- Ishi confirmed this will be fixed in v6.8.0 and described it as a simple fix
- The previous catch that wiped the LWK DB on corruption may have been removed
- Only secure storage being wiped would cause actual fund loss — LWK DB corruption is recoverable

thibistaken:
## Test Case

1. Open app, then immediately force close during sync
2. Reopen the app
3. Background the app during sync, reopen
4. **Verify**: app always recovers and starts, even if LWK DB was corrupted
5. **Verify**: seed/wallet access is never blocked by DB state

PR #1893, opened 2026-02-26
6.8.0 Updates

i5hi:
- [x] Lwk db fixes (#1891)
- [x] Sqlite migration catch blocks for lables
- [x] Error screen in main to prevent freeze on startup errors
- [x] Remove background tasks except log prune
- [x] Point to fss fork with migration fixes for android
- [ ] Show update notifications as a bottom sheet when new updates are available. Link users to play store to enable/disable autoupdate


lib/main.dart
```diff
@@ -104,41 +105,239 @@ class Bull {
...
       runApp(const BullBitcoinWalletApp());
     },
     (error, stackTrace) async {
       log.severe(error: error, trace: stackTrace);
+      runApp(BullBitcoinWalletAppInitErrorScreen(error: error));
     },
   );
 }
 
+class BullBitcoinWalletAppInitErrorScreen extends StatefulWidget {
```
basantagoswami: Maybe this entire screen can be in a different file? main is already a bit too cluttered

i5hi: Good point. Move it to features/app_startup or core/widgets?

@kumulynja @ethicnology  wdyt?

basantagoswami: Probably in core/screens next to `route_error_screen.dart`

basantagoswami: Everything else looks good to me

lib/main.dart
```diff
@@ -104,41 +105,239 @@ class Bull {
...
+                      style: theme.textTheme.bodyMedium,
+                    ),
+                    const SizedBox(height: 8),
+                    GestureDetector(
+                      onTap: () => launchUrl(
+                        Uri.parse(SettingsConstants.webSupportLink),
+                        mode: LaunchMode.externalApplication,
+                      ),
+                      child: Text(
+                        'app.bullbitcoin.com/support',
```
basantagoswami: Instead of hardcoding it, maybe do SettingsConstants.webSupportLink.replaceFirst('https://', '') here just in case the URL changes some day and we have to update it

lib/main.dart
ethicnology: > Probably in core/screens next to route_error_screen.dart

Agreed

lib/main.dart
```diff
@@ -33,11 +33,12 @@ import 'package:lwk/lwk.dart';
...
 import 'package:payjoin_flutter/common.dart';
 import 'package:sentry_flutter/sentry_flutter.dart';
+import 'package:share_plus/share_plus.dart';
+import 'package:url_launcher/url_launcher.dart';
 import 'package:workmanager/workmanager.dart';
 
 class Bull {
   static Future<void> init() async {
-    await initLogs();
     await initFlutterRustBridgeDependencies();
```
ethicnology, line 41: Do we really need to extract `initLogs` from `Bull.init` ?

lib/main.dart
```diff
@@ -104,41 +105,239 @@ class Bull {
...
+  @override
+  State<BullBitcoinWalletAppInitErrorScreen> createState() =>
+      _BullBitcoinWalletAppInitErrorScreenState();
+}
+
+class _BullBitcoinWalletAppInitErrorScreenState
+    extends State<BullBitcoinWalletAppInitErrorScreen> {
+  Locale _locale = const Locale('en');
+
+  static const _languageNames = {
```
ethicnology: also there is 11 more languages now

lib/main.dart
i5hi: the reason it was there is because its initialization to be guarenteed incase something throws and we land in the error page, logs need to be available. I have now moved it to init - first call.

basantagoswami (approved):

PR #1909, opened 2026-03-04
check self spent before doing payjoin send

### LINKED ISSUE #1384: Prevent payjoin with self-spend
Override payjoun if self-spend

kumulynja:
Added a check for a self-spent in the Bitcoin transaction preparation phase, which is then stored in the Cubit state and used further in the cubit at transaction confirmation and broadcasting to skip any payjoin handling if it is a self-spent.

As a little extra, since we have it in the state now, I added the Self-transfer row to the confirmation screen (see screenshot below) in case it is a self-spent, so the user is aware and it is not by mistake. Nothing is shown if it is not a self-spent.

Translations of the "Self-transfer" label in all languages are done by Claude, so not sure about all of them. Let me know if I should remove them or if it's ok?

Screenshot 2026-03-04 at 21 28 49

### DIFF
--- a/lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
+++ b/lib/core/wallet/data/datasources/bdk_wallet_datasource.dart
@@ -144,6 +144,21 @@ class BdkWalletDatasource {
     return isMine;
   }
 
+  Future<bool> isAddressMine(
+    String address, {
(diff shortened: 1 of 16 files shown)


i5hi (approved):
tACK

PR #1914, opened 2026-03-06
Add consent scam warning for all exchange funding flows

i5hi:
- [x] Create datasource/repo method to consentScamWarning api
- [x] Check if userSummary has the CONSENT_SCAM group - if not, show a page for user to accept and gain the group


lib/features/fund_exchange/presentation/bloc/fund_exchange_bloc.dart
```diff
@@ -110,4 +118,46 @@ class FundExchangeBloc extends Bloc<FundExchangeEvent, FundExchangeState> {
...
+    FundExchangeScamWarningConsentNavigationHandled event,
+    Emitter<FundExchangeState> emit,
+  ) {
+    emit(
+      state.copyWith(
+        scamWarningConsentSubmittedSuccessfully: false,
+        hasConfirmedNoCoercion: false,
+      ),
+    );
+  }
```
kumulynja: No usecases are called here, so there is no need for a controller (Bloc) function here. No data needs to be passed to or come from any other layer, so there is no point in changing the state here and so no point in having this function here.

lib/features/fund_exchange/presentation/bloc/fund_exchange_event.dart
```diff
@@ -11,5 +11,9 @@ sealed class FundExchangeEvent with _$FundExchangeEvent {
   const factory FundExchangeEvent.fundingDetailsRequested({
     required FundingMethod fundingMethod,
   }) = FundExchangeFundingDetailsRequested;
+  const factory FundExchangeEvent.scamWarningConsentSubmitted() =
+      FundExchangeScamWarningConsentSubmitted;
+  const factory FundExchangeEvent.scamWarningConsentNavigationHandled() =
+      FundExchangeScamWarningConsentNavigationHandled;
```
kumulynja: I don't think this event is needed as nothing has to be send or fetched from other layers/usecases on it

lib/features/fund_exchange/presentation/bloc/fund_exchange_state.dart
```diff
@@ -11,6 +11,9 @@ sealed class FundExchangeState with _$FundExchangeState {
     @Default(false) bool hasConfirmedNoCoercion,
     FundingDetails? fundingDetails,
     GetExchangeFundingDetailsException? getExchangeFundingDetailsException,
+    @Default(false) bool isSubmittingScamWarningConsent,
+    @Default(false) bool scamWarningConsentSubmittedSuccessfully,
```
kumulynja: Can't this be derived from the updated user summary?

lib/features/fund_exchange/presentation/bloc/fund_exchange_state.dart
```diff
@@ -11,6 +11,9 @@ sealed class FundExchangeState with _$FundExchangeState {
     @Default(false) bool hasConfirmedNoCoercion,
```
kumulynja: Should this not be removed then? It feels like this is the same flag, no?

lib/features/fund_exchange/ui/screens/fund_exchange_warning_screen.dart
```diff
@@ -17,171 +18,116 @@ class FundExchangeWarningScreen extends StatelessWidget {
...
-    );
+    return BlocConsumer<FundExchangeBloc, FundExchangeState>(
+      listenWhen:
+          (previous, current) =>
+              !previous.scamWarningConsentSubmittedSuccessfully &&
+              current.scamWarningConsentSubmittedSuccessfully,
+      listener: (context, state) {
+        context.read<FundExchangeBloc>().add(
+          const FundExchangeEvent.scamWarningConsentNavigationHandled(),
+        );
```
kumulynja: I don't understand the point of this.

lib/features/fund_exchange/ui/screens/fund_exchange_warning_screen.dart
```diff
@@ -17,171 +18,116 @@ class FundExchangeWarningScreen extends StatelessWidget {
...
+                    shape: RoundedRectangleBorder(
+                      borderRadius: BorderRadius.circular(8.0),
+                    ),
+                    value: hasConfirmedNoCoercion,
+                    onChanged:
+                        isLoading
+                            ? null
+                            : (value) {
+                              context.read<FundExchangeBloc>().add(
+                                FundExchangeEvent.noCoercionConfirmed(
```
kumulynja: Best to just keep this as local widget state in a Stateful widget to disable/enable the submit button and then only on submit add the bloc event and so trigger a bloc function and bloc state change.

lib/features/fund_exchange/ui/widgets/fund_exchange_method_list_tile.dart
```diff
@@ -32,11 +32,20 @@ class FundExchangeMethodListTile extends StatelessWidget {
...
+                ?.hasConsentedScamWarning ??
+            false;
+
+        if (hasConsented) {
+          context.pushNamed(FundExchangeRoute.routeNameFor(method));
+        } else {
+          // Reset checkbox state before showing the warning screen.
+          context.read<FundExchangeBloc>().add(
+            const FundExchangeEvent.noCoercionConfirmed(false),
+          );
```
kumulynja: By removing the noCoercingConfirmed field from the Bloc state and just keeping it as local state in the warning screen, this hack to reset the checkbox is not needed anymore.

kumulynja (changes requested):

lib/features/fund_exchange/presentation/bloc/fund_exchange_state.dart
i5hi: true.

lib/features/fund_exchange/ui/screens/fund_exchange_warning_screen.dart
i5hi: yeah; the whole approach was dumb. this was to reset the state of scamWarningConsentSubmittedSuccessfully - which we dont need if we handle in the stateful widget

lib/features/fund_exchange/ui/widgets/fund_exchange_method_list_tile.dart
i5hi: yeah; this was a real overkill

kumulynja (approved):
👍

PR #1928, opened 2026-03-12
fix: import mnemonic passphrase and label input keyboard lag

### LINKED ISSUE #1867: Keyboard Lag Issue on Wallet Import Screen
🔗 [View Slack thread](https://slack.com/app_redirect?channel=C09899B4C9Z&message_ts=1770652384.561479)

👤 Asked by: [Basanta Goswami](https://satoshi-portal.slack.com/team/U09E0QJSRH7)

🐞 Users experience significant lag when trying to bring up the keyboard by pressing the input label while importing a new wallet. The keyboard response is so slow that it becomes nearly unusable

basantagoswami:
Fixes: #1867 (Keyboard Lag Issue on Wallet Import Screen)

- addPostFrameCallback guard (hints.first != widget.words[_focusedDisplayIndex])

When you finish typing a word and there's only one possible match, autofill fires: it schedules _onHintTap which sets the word and moves focus to the next field. That's the intended behaviour. But once all 12 words are complete and focus is still on the last word, hints always has exactly one match: the word that's already there. The old code scheduled _onHintTap unconditionally, which called onWordChanged -> parent setState -> rebuild -> scheduled _onHintTap again -> infinite loop. The guard breaks this: if the word is already set to the only matching hint, there's nothing to autofill, so don't schedule the callback.

- Controller text in didUpdateWidget instead of build()

_controller.text = widget.word was being called on every build(). TextEditingController's text setter calls notifyListeners(), which tells Android the text input has changed. With 12 word fields all doing this during a single rebuild, Android tries to reinitialize the IME 12 times - roughly 80ms each - adding up to 1 second of lag. Moving it to didUpdateWidget with a change check means the controller only notifies when the word actually changes for ex via autofill, not on every rebuild.

### DIFF
--- a/lib/core/widgets/mnemonic_widget.dart
+++ b/lib/core/widgets/mnemonic_widget.dart
@@ -189,7 +189,7 @@ class MnemonicWord extends StatefulWidget {
 }
 
 class MnemonicWordState extends State<MnemonicWord> {
-  final _controller = TextEditingController();
+  late final _controller = TextEditingController(text: widget.word);
 
   String get displayIndex {
     final displayIndex = widget.index + 1;
@@ -197,9 +197,22 @@ class MnemonicWordState extends State<MnemonicWord> {
   }
 
   @override
+  void didUpdateWidget(MnemonicWord oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    if (oldWidget.word != widget.word && _controller.text != widget.word) {
+      _controller.text = widget.word;
+    }
+  }
+
+  @override
+  void dispose() {
+    _controller.dispose();
+    super.dispose();
+  }
+
+  @override
   Widget build(BuildContext context) {
     final isValidWord = widget.language.isValid(widget.word);
-    _controller.text = widget.word;
 
     return Container(
       padding: const EdgeInsets.all(3),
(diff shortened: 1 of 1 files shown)


lib/core/widgets/mnemonic_widget.dart
```diff
@@ -189,17 +189,30 @@ class MnemonicWord extends StatefulWidget {
 }
 
 class MnemonicWordState extends State<MnemonicWord> {
-  final _controller = TextEditingController();
+  late final _controller = TextEditingController(text: widget.word);
```
ethicnology: `late` doesn't seems required if you can instantiate during declaration.

If not strictly needed we should remove it to avoid late initialization errors.

ethicnology (changes requested):
is `late` required ?

lib/core/widgets/mnemonic_widget.dart
basantagoswami: late isn't strictly needed. did this this instead.

-  late final _controller = TextEditingController(text: widget.word);
+  final _controller = TextEditingController();
+
+  @override
+  void initState() {
+    super.initState();
+    _controller.text = widget.word;
+  }

ethicnology: cool!

ethicnology (approved):

kumulynja (approved):

i5hi (approved):

i5hi (approved):

PR #1941, opened 2026-03-15
fix: prevent same wallet selection in transfer dropdowns and add swap/flip button

### LINKED ISSUE #1939: Bug: Wallet Origin Change in Swap Doesn't Switch to Liquid
🔗 [View Slack thread](https://slack.com/app_redirect?channel=C09899B4C9Z&message_ts=1768060738.025269)

👤 Asked by: [Basanta Goswami](https://satoshi-portal.slack.com/team/U09E0QJSRH7)

Changing the wallet origin in a swap/transfer by selecting something else in the from wallet dropdown doesn't automatically change the to wallet drodpdown option

basantagoswami:
Fixes #1939 

This PR

- disables and greys out the opposite wallet in from/to dropdowns
- sort disabled items to the bottom of each dropdown
- add swap/switch button between the from and to dropdowns
- fix wallet swap state corruption by detecting direct wallet swap in _onWalletsChanged

The transfer screen has a few pre-existing issues and a couple of new ones introduced by my changes

Pre-existing

- The Internal/External Transfer toggle also toggles the toggle below (Subtract fees from amount / Receive exact amount)
- Hardware wallets are hidden from the "from" dropdown. Probably done purposefully but the transfer page works identical to a send flow with/without swaps, so I don't understand why it was done this way
- I don't really get why External Transfer as it's implemented today. the "to" external address can be a Liquid address if the "from" is a Bitcoin wallet and vice versa, but "to" can be Secure Bitcoin and "from" can be another imported Bitcoin wallet when it's an internal transfer, so why disallow it in the external transfer flow?

New issues I introduced in this PR

- Instant Payments wallet is selected in "from" by default which means Secure Bitcoin is selected by default in "to" and it's greyed out in the "from" dropdown list. If a user wants to do the opposite, they'll have to click the swap/flip button, which is not very intuitive. We should instead probably keep the "to" option empty initially, and display all the from dropdown options, and once one is chosen, collapse it and give the option to choose the "to"

Issue my code introduced, but I haven't addressed yet

- Toggling it to be an external transfer still keeps some wallets greyed out from the "from" dropdown menu. The state should be cleared

### DIFF
--- a/lib/core/widgets/dropdown/bb_dropdown.dart
+++ b/lib/core/widgets/dropdown/bb_dropdown.dart
@@ -43,6 +43,7 @@ class BBDropdown<T> extends StatelessWidget {
             items: items.map((item) {
               return DropdownMenuItem<T>(
                 value: item.value,
+                enabled: item.enabled,
                 child: Container(
                   width: double.infinity,
                   alignment: Alignment.centerLeft,
--- a/lib/features/swap/presentation/transfer_bloc.dart
+++ b/lib/features/swap/presentation/transfer_bloc.dart
@@ -242,23 +242,29 @@ class TransferBloc extends Bloc<TransferEvent, TransferState> {
       return;
     }
 
-    if (newFromWallet.isLiquid == newToWallet.isLiquid) {
-      if (newFromWallet.isLiquid) {
-        return;
-      }
-    } else {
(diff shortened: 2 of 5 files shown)


i5hi:
This was fixed in another PR https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/1943

Issue #1390, opened 2025-10-29
Bull Bitcoin Mobile v6.2.3 - Reproducible Build Issue


xrviv:
Hi Bull Bitcoin team! 👋

We've been working on verifying the reproducibility of Bull Bitcoin Mobile v6.2.3 for WalletScrutiny, and we wanted to share our findings with you. The good news is that all native Rust libraries (libbdk_flutter, libark_wallet, libboltz, etc.) compile identically from source - great work on that front! However, we discovered that **R8 obfuscation** is running with non-deterministic settings, which prevents the APK from being reproducible.

We found that R8 v8.6.17 is generating unique obfuscation mappings on each build, with a different `pg_map_hash` every time (we extracted the 20MB proguard.map file from the AAB's BUNDLE-METADATA). This results in 466,604 lines of bytecode differences across 1,128 files, with different class names (e.g., A1/A.smali vs A2/e.smali) and optimization passes between builds. While R8 is enabled automatically by Flutter's release build process, the randomized obfuscation seed makes it impossible for third parties to verify that the published binary matches the public source code.

Here are a few options:

**Option 1: Disable R8 obfuscation for release builds**
```gradle
buildTypes {
    release {
        minifyEnabled false
    }
}
```

**Option 2: Use fixed obfuscation dictionary (keeps obfuscation but makes it deterministic)**
```gradle
buildTypes {
    release {
        minifyEnabled true
        proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
    }
}
```

In `android/app/proguard-rules.pro`:
```
-obfuscationdictionary dictionary.txt
-classobfuscationdictionary dictionary.txt
-packageobfuscationdictionary dictionary.txt
```

**Option 3: Publish mapping files with releases** (allows verification via comparison)

We also noticed the `.env` file has production values in the official APK but uses template placeholders in the repository. Documenting the exact build-time configuration would help with reproducibility verification as well.

We're happy to collaborate!

Kind Regards,
Danny Garcia
**WalletScrutiny.com**

i5hi:
Hey thanks for following up! We have prioritized this for the upcoming spint.

We will make the updates requested and follow up.

xrviv:
# Update for 6.5.2

Hello, thank you for your attention to this matter.

## GitHub release APK v6.5.2 is not publicly reproducible from the repo

[...]

PR #1966, opened 2026-03-20
add .env validation on app startup and pre-build check tool

### LINKED ISSUE #1908: Setup .env check before building the app
Anything we add in the .env file will end up being bundled into the final release APK. While extracting an APK to get it's content, I have seen some variables in the .env file that we don't use in the project, probably added temporarily to it in the machine that built the release APK, which then didn't get removed

We need to either add this check to this repo itself or to mobile-operator

basantagoswami:


thibistaken:
Can you review this @i5hi?

ethicnology:
@thibistaken 

I'm closing this one since we do not have `.env` anymore with the work done on reproducible builds

related to https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1390#issuecomment-4161474944

cc: @i5hi

PR #1970, opened 2026-03-24
Support screen features 

BullishNode:
Redesigned the support login screen to preview the chat UI with a login overlay, replacing the static logo/text layout. Replaced the Telegram link in Settings with a GitHub link, and added a "Get Help" button that routes to the support chat (or login screen if not logged in), Removed the iOS platform/superuser gate on the support login route so that it is now harmonized.  Changed pushNamed to goNamed for support login navigation from exchange home and wallet app bar, so the support flow opens as a standalone screen rather than pushing onto the current stack, Added settingsGetHelpLabel and exchangeSupportLoginChatRequired translations for all 25 languages, Removed unused exchangeSupportLoginSubtitle, exchangeSupportLoginAccessMessage, and exchangeSupportLoginAccountInfo keys from all language files. Used theme color for the overlay instead of hardcoded Colors.black


i5hi (commented):
tACK

basantagoswami (approved):

ethicnology (approved):

PR #1971, opened 2026-03-24
Update support features

i5hi:
- [x] Remove Telegram link
- [x] Update message to login for support chat
- [x] Bump build number
- [x] Allow account deletion
- [x] Update boltz dependency to include BULL referals on chain swaps

### DIFF
--- a/lib/features/exchange/ui/screens/exchange_home_screen.dart
+++ b/lib/features/exchange/ui/screens/exchange_home_screen.dart
@@ -147,7 +147,7 @@ class ExchangeHomeScreen extends StatelessWidget {
                                               .state
                                               .notLoggedIn;
                                           if (notLoggedIn) {
-                                            context.goNamed(
+                                            context.pushNamed(
                                               ExchangeRoute
                                                   .exchangeLoginForSupport
                                                   .name,
--- a/lib/features/exchange/ui/screens/exchange_support_login_screen.dart
+++ b/lib/features/exchange/ui/screens/exchange_support_login_screen.dart
@@ -4,7 +4,6 @@ import 'package:bb_mobile/core/widgets/buttons/button.dart';
 import 'package:bb_mobile/core/widgets/inputs/text_input.dart';
 import 'package:bb_mobile/core/widgets/text/text.dart';
 import 'package:bb_mobile/features/exchange/ui/exchange_router.dart';
-import 'package:bb_mobile/features/wallet/ui/wallet_router.dart';
 import 'package:flutter/material.dart';
 import 'package:gap/gap.dart';
 import 'package:go_router/go_router.dart';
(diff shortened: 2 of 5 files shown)


lib/features/exchange_support_chat/ui/screens/exchange_support_chat_screen.dart
```diff
@@ -36,6 +41,18 @@ class ExchangeSupportChatScreen extends StatelessWidget {
...
+          leading: BackButton(
+            color: context.appColors.onSurface,
+            onPressed: () {
+              if (context.canPop()) {
+                context.pop();
+              } else if (fromExchange) {
+                context.goNamed(ExchangeRoute.exchangeHome.name);
+              } else {
+                context.goNamed(WalletRoute.walletHome.name);
+              }
```
kumulynja, line 53: Just wondering if it is really needed to use `goNamed` when navigating to the exchange support chat screen? If you have a back button here anyways, why not just use pushNamed anywhere and let the default leading back button work to go back to where the user came from and so remove this BackButton here.
Of course if we really want that the user after support chat only can go  to exchange home or wallet home, then we need it like this. But I feel like if I was in settings and used the "Get Help" button there, I should also just go back to settings instead of to home.

lib/features/settings/ui/screens/all_settings_screen.dart
```diff
@@ -192,18 +202,28 @@ class _AllSettingsScreenState extends State<AllSettingsScreen> {
...
-                        );
-                        launchUrl(url, mode: LaunchMode.externalApplication);
+                        final notLoggedIn =
+                            context.read<ExchangeCubit>().state.notLoggedIn;
+                        if (notLoggedIn) {
+                          context.goNamed(
+                            ExchangeRoute.exchangeLoginForSupport.name,
+                          );
+                        } else {
+                          context.goNamed(
```
kumulynja: See other comment regarding goNamed instead of just pushNamed

lib/features/wallet/ui/widgets/wallet_home_app_bar.dart
```diff
@@ -98,7 +98,7 @@ class _WalletHomeAppBarState extends State<WalletHomeAppBar> {
                             .state
                             .notLoggedIn;
                         if (notLoggedIn) {
-                          context.pushNamed(
+                          context.goNamed(
```
kumulynja: See other comment regarding goNamed instead of just pushNamed

lib/features/exchange/ui/screens/exchange_home_screen.dart
```diff
@@ -147,7 +147,7 @@ class ExchangeHomeScreen extends StatelessWidget {
                                               .state
                                               .notLoggedIn;
                                           if (notLoggedIn) {
-                                            context.pushNamed(
+                                            context.goNamed(
```
kumulynja: See other comment regarding goNamed instead of just pushNamed

lib/features/exchange/ui/screens/exchange_support_login_screen.dart
```diff
@@ -28,99 +27,247 @@ class ExchangeSupportLoginScreen extends StatelessWidget {
             color: context.appColors.onSurface,
             onPressed: () => context.goNamed(WalletRoute.walletHome.name),
           ),
```
kumulynja: If we have a BackButton anyways here too, why not use the default/native back behavior? Why not use push to get here instead of go and then there is no need to overwrite the leading widget...

PR #1977, opened 2026-03-24
Fix Show Address Only switch overflow in receive

### LINKED ISSUE #1913: Certain lanugages overflow in recieve making it impossible to click Show Address Only
## Description
When French (possible other too) is selected as the language, on the recieve screen - the button to `Copy only address` overflows out of the screen and makes it impossible to toggle.

### Reproduce 
1. Go to settings - app settings - language - french
2. Goto the recieve page
3. At the bottom notice the copy only address toggle

### Expected behavior
The text and button should be in separate columns and the toggle button should be locked in place and the text column should allow wrapping across multiple lines

### Demo

Image

kumulynja:
- Use Expanded for the Text to take up as much space as possible without pushing the Switch out of the boundary
- Use Column so the Text can wrap to the next line if not enough space
- Add a Gap between the Text and Switch so the text is never completely pegged against the switch

### DIFF
--- a/lib/features/receive/ui/screens/receive_qr_screen.dart
+++ b/lib/features/receive/ui/screens/receive_qr_screen.dart
@@ -619,11 +619,18 @@ class ReceiveCopyAddress extends StatelessWidget {
       padding: const EdgeInsets.symmetric(horizontal: 16 + 12),
       child: Row(
         children: [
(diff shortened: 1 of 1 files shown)


i5hi:
tACK

PR #1978, opened 2026-03-24
Create automated script for testing bull bitcoin mobile against payjoin-cli

benalleng:
This creates a script to automate the integration test between the payjoin-cli and bull bitcoin. This first test is using bull bitcoin as the sender and payjoin-cli as the receiver.

Claude Opus had a heavy hand in porting over the payjoin_test.dart to work with this new format

Blocked until #2041 is merged. Though this is also working with the current payjoin-flutter sdk I think it is probably more useful to wait until it is merged.


benalleng:
A note about the cli integration: As it is using testnet the funding wallets need to have enough utxos to do both tests at the same time since there is no guarantee that the first tx will confirm before the next test starts. perhaps there is some conditional logic to choose unconfirmed utxos but it is simple enough to just get some more funds in the relevant wallets

PR #1982, opened 2026-03-25
Use pushNamed for support chat navigation to preserve back stack

i5hi:


lib/features/exchange/ui/screens/exchange_support_login_screen.dart
```diff
@@ -14,137 +13,130 @@ class ExchangeSupportLoginScreen extends StatelessWidget {
...
-            context.loc.exchangeSupportChatTitle,
-            style: context.font.headlineMedium,
-          ),
-          backgroundColor: context.appColors.background,
+    return Scaffold(
+      backgroundColor: context.appColors.background,
+      appBar: AppBar(
+        leading: BackButton(
+          color: context.appColors.onSurface,
+          onPressed: () => context.pop(),
```
kumulynja: Is this done to overwrite the color specifically for this screen? Or can we just not even specify the leading widget so it just uses the default leading widget, which is a back button if there are screens on the navigation stack?

kumulynja (approved):
👍

Issue #1983, opened 2026-03-25
Use push instead of go for routing in support chat feature


i5hi:
Use push and pushReplacement instead of go for all routing for exchange support chat.

PR #1985, opened 2026-03-26
Fix routing to support chat. Update delete account content.

i5hi:
- [ ] Fixes https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1983
- [ ] Fixes broken navigation to Get Help button in settings
- [x] Update content for delete account


lib/features/exchange/ui/screens/exchange_home_screen.dart
```diff
@@ -153,10 +156,26 @@ class ExchangeHomeScreen extends StatelessWidget {
...
                                               ExchangeSupportChatRoute
                                                   .supportChat
                                                   .name,
+                                              queryParameters:
+                                                  isIOSNonSuperuser
+                                                      ? {
+                                                          'backToWalletHome':
+                                                              'true',
+                                                        }
+                                                      : {},
```
kumulynja, line 178: Is this really necessarry? The user will never get on the exchange home screen if non superuser in iOS, so the pushNamed to support chat will never be done from here, only from the wallet home or settings.

lib/features/exchange/ui/exchange_router.dart
```diff
@@ -107,10 +106,11 @@ class ExchangeRouter {
...
                 final isIOSNonSuperuser = Platform.isIOS &&
                     !(context.read<SettingsCubit>().state.isSuperuser ?? false);
-                context.goNamed(
+                context.pushNamed(
                   ExchangeSupportChatRoute.supportChat.name,
-                  queryParameters:
-                      isIOSNonSuperuser ? {} : {'from': 'exchange'},
+                  queryParameters: isIOSNonSuperuser
+                      ? {'backToWalletHome': 'true'}
+                      : {'from': 'exchange'},
```
kumulynja, line 113: Also here, why is this `fromSupport` param needed and the from exchange and backToWalletHome parameters? If you just push from screens the user was already allowed to go in, by using normal back behavior, he can never end up on screens he was not allowed to be on. That can only happen if goNamed is used somewhere, but in all of these cases regarding the support login and support chat, it should not be the case. So I feel this is all still too complicated and much of these conditions and queryParameters can just be removed with proper push and back behavior everywhere.

lib/features/exchange_support_chat/ui/screens/exchange_support_chat_screen.dart
```diff
@@ -24,9 +24,14 @@ import 'package:gap/gap.dart';
...
 class ExchangeSupportChatScreen extends StatelessWidget {
-  const ExchangeSupportChatScreen({super.key, this.fromExchange = false});
+  const ExchangeSupportChatScreen({
+    super.key,
+    this.fromExchange = false,
+    this.backToWalletHome = false,
+  });
 
   final bool fromExchange;
+  final bool backToWalletHome;
```
kumulynja, line 34: Same comment, this all feel like hacks and shouldn't be used if goNamed is not used. Since with push you can not go back to somewhere you don't come from.

kumulynja (commented):
I still see a lot of query parameters use and imo opinion unneccessary fields and conditions to manage the navigation. I think the whole support chat navigation, including the support chat login, can be managed without all of this by just using push instead of go and then you don't need to explicitly navigate back to somewhere, just use default `pop` back behavior to go back where the user came from.

PR #1988, opened 2026-03-26
fix(send): persist transaction labels and auto-populate from invoice labels (#1173)

### LINKED ISSUE #1173: LNURL payment: added notes are not saved in the transaction history
When paying via LNURL, I can set a custom amount and add a note in the wallet, but the note will not appear in the transaction history.

Image
Image

ethicnology:
**Fixes #1173**

## What was broken
When sending a transaction, any note you added in the send flow but was never actually written to the database. Same story for invoices that already carry a label (Bolt11 `description` field, BIP21 `label` param) 

## Tested
- Bitcoin
- Liquid
- Lightning invoice from Wallet of Satoshis

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to Bull Bitcoin Mobile will be documented in this file.
 
+## [X.X.X] - To be released
+
+### Bug Fixes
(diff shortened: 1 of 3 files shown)


i5hi (approved):

i5hi:
tACK

PR #1991, opened 2026-03-28
fix: Handle back button emit error state

### LINKED ISSUE #1979: Unhandled Error  Bad state: Cannot emit new states after calling close
## Description
We need to handle these cases properly in cubits/blocs where the cubit is closed before an async operation completes.
```
 SEVERE  Global Unhandled Error  Bad state: Cannot emit new states after calling close  #0      BlocBase.emit (package:bloc/src/bloc_base.dart:100:9)
```

### Reproduce 
1. Go to any feature, for example service status
2. Immediately exit
3. Observe logs

### Expected behavior
The cubit should handle early exists

kevkevinpal:
## Summary
This closes https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1979

This introduces `SafeCubit`, which extends `Cubit`. What I found was that if we wanted to avoid getting a `Severe` error when hitting the back button we would want to check to see if `isClosed` is set to true. If its False we would want to avoid calling `emit`.

We can see this already being done in `lib/features/transactions/presentation/blocs/transactions_cubit.dart`

---

Let me know if this is the wrong approach or missed anything this is my first PR in this repo. I have tested locally, but not sure if I missed any.


kumulynja:
Thanks a lot for contributing!

I think this is a good and convenient guardrail and we should definitely not use log severe for these Bloc StateErrors, but the fact that a state is still being emitted after the cubit is closed, might indicate an issue to solve differently than just checking `isClosed`. One reason such error could happen is that we might have forgotten to cancel a subscription in the `close` handler of the cubit. If we just "swallow" the error with the `SafeCubit` we are not notified about this to fix it.
So to not lose this info that something might not be done well in the cubit, it would be better if the `SafeCubit` still logs some warning at least. Not a severe error, but at least a warning if the `isClosed` is true.

Previously the Bloc library itself was swallowing this error and they changed that behavior specifically for this reason:
https://bloclibrary.dev/migration/#-emit-throws-stateerror-if-bloc-is-closed, so just swallowing them ourselves now doesn't seem like something we should do.

kevkevinpal:
Thank you for the review @kumulynja

> So to not lose this info that something might not be done well in the cubit, it would be better if the SafeCubit still logs some warning at least. Not a severe error, but at least a warning if the isClosed is true.

Sounds good I pushed [9e3448d](https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/1991/commits/9e3448d3b6d6e8c1ff0c7c0037d63043874af5ea) and it now logs if `isClosed` is true

---

The one situation I tested was 
Settings -> Services Status -> quickly press back -> Observe logs

If other situations might still give a severe warning, it would be useful to test those and see if we can handle it in `SafeCubit`

I'm just unsure of which other situations those might be

i5hi:
@kevkevinpal Thanks for the PR! There was a minor merge conflict that I attempted to fix but it has led to warnings `unused imports` could you please fix it on your branch

Note since this touches a lot of files, we may face a lot of conflicts and we need to prioritize issues that are part of our sprint. I will come back to check in on this PR after this sprint is over and we have published the next release. So, next Monday I will come back to this.

i5hi:
@kevkevinpal All good! I was able to edit it from this PR directly.

PR #1996, opened 2026-03-30
feat: added skeleton loader for home screen fiat price

### LINKED ISSUE #1409: Buggy pricer
Very often, fiat conversations do not appear on home even if pricer status is online

kevkevinpal:
## Summary
Might close https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1409

Motivated by https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1409#issuecomment-3971622177

Sometimes, when we try to fetch the price data, it either fails or takes a significant amount of time. Right now we display a 0 value if the fetch fails, which is inaccurate.

It makes sense to instead show a skeleton loader while we don't have the price data.

## Video of Fix
https://github.com/user-attachments/assets/917937e3-1b00-40a9-afcf-361ffed5c87a

### DIFF
--- a/lib/features/wallet/ui/widgets/home_fiat_balance.dart
+++ b/lib/features/wallet/ui/widgets/home_fiat_balance.dart
@@ -1,7 +1,6 @@
 import 'package:bb_mobile/core/exchange/domain/usecases/get_available_currencies_usecase.dart';
 import 'package:bb_mobile/core/widgets/bottom_sheet/x.dart';
 import 'package:bb_mobile/core/themes/app_theme.dart';
-import 'package:bb_mobile/core/utils/build_context_x.dart';
 import 'package:bb_mobile/core/widgets/price_input/price_input.dart';
 import 'package:bb_mobile/features/bitcoin_price/presentation/bloc/bitcoin_price_bloc.dart';
 import 'package:bb_mobile/features/bitcoin_price/ui/currency_text.dart';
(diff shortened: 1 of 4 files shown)


ethicnology:
Thanks for your contribution, definitely a UX improvement.

Note (out of scope): The pricer API is called on every user interaction with no caching. To reduce load and scale with our user base, exchange rates should be cached per currency and refreshed only when the cached value is older than 1 minute. Let's not make the pricer pricy.

cc @i5hi @kumulynja

ethicnology:
Hey @kevkevinpal  thanks for your contribution to this project!

@Claude do you mind giving your opinion on this PR?

kevkevinpal:
> Note (out of scope): The pricer API is called on every user interaction with no caching. To reduce load and scale with our user base, exchange rates should be cached per currency and refreshed only when the cached value is older than 1 minute. Let's not make the pricer pricy.

I agree that would make a good follow-up

kevkevinpal:
@claude do you mind giving your opinion on this PR?

lib/features/wallet/ui/widgets/home_fiat_balance.dart
```diff
@@ -55,50 +62,23 @@ class HomeFiatBalance extends StatelessWidget {
...
+    // [startupFailed], or a successful currency change still shows skeleton.
+    final showSkeleton =
+        showBalanceLoading || loadingPrice || !hasValid;
 
-    if (startupFailed) {
+    if (showSkeleton) {
       return GestureDetector(
         behavior: HitTestBehavior.opaque,
-        onTap: () {
-          context.read<BitcoinPriceBloc>().add(const BitcoinPriceStarted());
```
i5hi, line 72: This behaviour is important to handle a failed failed to load scenario. Without it a user will see a perpetual loading indicator with no clear way to retry. There is no pull to sync logic on price fetch, nor do we poll.

i5hi: Making an issue for this and merging this PR as is. The existing work is sound.
Tracking here: https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2208

PR #1998, opened 2026-03-30
fix: improve service status and indirectly electrum servers

### LINKED ISSUE #1923: Server Status should make realistic calls
## Description
The electrum server status shows online when infact the server is unable to handle requests like balance and broadcast. We should not use just ping for service statuses, instead we should make realistic calls. For eelectrum specifically we should link the status to last sync.

### LINKED ISSUE #1999: Health check should validate actual operations (sync, broadcast, fees), not just ping

ethicnology:
## Fixes

- **BUG**: Electrum "online" was based on the last sync timestamp, not actual connectivity
- **BUG**: Custom Electrum servers were saved even if the daemon was unreachable — the socket being open was not enough, the server must also respond to the Electrum protocol handshake
- **BUG**: Mempool always showed "online" even in airplane mode because the Liquid network fees check always returns hardcoded values without any network call
- **UX**: SSL toggle was re-enabled automatically when typing a clearnet `host:port`, overriding an explicit user choice
- **UX**: The service status page showed no progress indicator on the first load

## Refactor

- Merged `ServerStatusDatasource` into `ServerStatusAdapter` (eliminated unnecessary layer)


ethicnology:
Hey @Claude please review my PR!

lib/core/electrum/domain/ports/server_status_port.dart
```diff
@@ -1,10 +1,19 @@
...
     int? timeout,
     bool useTorProxy = false,
     int torProxyPort = 9050,
   });
+
+  /// Verifies the server speaks the Electrum protocol by sending a
+  /// `server.version` request. Works on all implementations (ElectrumX,
+  /// Fulcrum, electrs). Returns [ElectrumServerStatus.online] if the server
+  /// responds with a valid result, [ElectrumServerStatus.offline] otherwise.
+  Future<ElectrumServerStatus> checkProtocol({
```
i5hi: Although better than a simple ping, this also may not be realistic enough. I was thinking we query a known txid. I could be wrong on this, have to test but if for example the chain is not in sync or its not properly connected to bitcoind, we may still get back a version but our actual queries fail.

lib/core/electrum/interface_adapters/adapters/server_status_adapter.dart
```diff
@@ -1,64 +1,256 @@
...
+
+  /// Sends a `server.version` JSON-RPC request and returns the raw response
+  /// line. [SecureSocket] extends [Socket], so both branches share the same
+  /// write/read logic after construction.
+  Future<String> _sendVersionRequest({
+    required Uri uri,
+    required int timeoutSeconds,
+  }) async {
+    const request =
+        '{"id":1,"method":"server.version","params":["bb-mobile","1.4"]}\n';
```
i5hi: Related to the last comment - we should probably call a more realistic method that is used in the app like blockchain.transaction.get

lib/core/electrum/domain/ports/server_status_port.dart
ethicnology: Before we were checking only if the server was up.
Now we are sure that we can communicate with the Electrum deamon.
I can switch to a transaction get, but technically, in both it's a proof of communication with the electrum daemon

ethicnology: The main advantage of not specifying querying a tx is to check if the electrum daemon is working despite the blockchain (liquid/bitcoin)

i5hi: but the issue that this PR addresses is making `realistic` checks - so if a user has an electrum failure like a sync and they see service status as online - it leaves us with the same problem

lib/core/electrum/interface_adapters/adapters/server_status_adapter.dart
ethicnology: 760a65fc202995e5e2c29c0cb8b3ccfbe0da360b

lib/core/electrum/domain/ports/server_status_port.dart
ethicnology: 760a65fc202995e5e2c29c0cb8b3ccfbe0da360b

ethicnology:
@i5hi rebased and amended

Bitcoin mainnet check for pizza day txid
Liquid mainnet check for first quantum proof txid

Testnet are falling back on electrum version

i5hi:
Thanks @ethicnology

PR #2002, opened 2026-03-31
fix: exclude payjoins from swaps filter and swaps from payjoin filter

### LINKED ISSUE #1961: Swap shown under Payjoin filter
## Description
Swap tx is shown under the payjoin filter in transaction list

Image

ethicnology:
It appears that some swaps appear in payjoin filter, this PR should fix it.

Not tested, please if you experience such weird swaps give me the feedback


i5hi:
The actual issue here was that ongoing transfers shows on all filters. Fixed in the next commit.

PR #2005, opened 2026-04-03
feat: SamRock Protocol support for BTCPay Server

RenAndKiwi:
## Summary
- Add SamRock Protocol integration allowing BTCPay Server to generate payment addresses from the user's Bull Bitcoin wallet descriptor
- Accept both BTCPay Server (`btc-chain`, `liquid-chain`, `btc-ln`) and Aqua (`btc`, `lbtc`, `btcln`) URL formats
- Add SamRock entry in Settings screen
- Add paste input alongside QR scanner for URL entry

## Test plan
- [x] Tested with live BTCPay Server at btcpay.bitcoinbutlers.com
- [x] Verified invoice payment address matches Bull Bitcoin wallet receive address
- [x] All 16 unit tests pass

### DIFF
--- a/lib/features/samrock/data/datasources/samrock_api_datasource.dart
+++ b/lib/features/samrock/data/datasources/samrock_api_datasource.dart
@@ -1,0 +1,72 @@
+import 'dart:convert';
+
+import 'package:bb_mobile/features/samrock/domain/entities/samrock_setup.dart';
+import 'package:http/http.dart' as http;
+
+class SamrockApiDatasource {
+  final http.Client _client;
+
+  SamrockApiDatasource({http.Client? client})
+      : _client = client ?? http.Client();
+
+  Future<SamrockSetupResponse> submitSetup({
+    required SamrockSetupRequest request,
+    required Map<String, dynamic> descriptorPayload,
+  }) async {
+    final url = Uri.parse(request.setupUrl);
+
+    if (url.scheme != 'https') {
+      throw Exception('SamRock setup requires HTTPS');
+    }
+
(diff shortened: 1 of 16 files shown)


BullishNode:
The samrock protocol normally specifies to create a seperate Liquid Wallet other than the main one (in our case the instant payment wallet). This should be considered as part of our multi-wallet feature.

RenAndKiwi:
Thanks for the feedback @BullishNode.

We looked into creating a separate Liquid wallet for SamRock. The issue is that `lwk.Descriptor.newConfidential()` only takes mnemonic + network — it always derives account index 0 (`m/84'/1776'/0'`). There's no way to create a second Liquid wallet at account index 1 through the current LWK API.

Two options we see:

1. **Derive a confidential descriptor at account index 1 outside LWK** (manual BIP32 + SLIP77 blinding key derivation), then import it as a watch-only wallet via `createPublicWallet`. Label it "BTCPay - {serverHost}".

2. **Wait for multi-wallet support** — if you already have a pattern planned for creating purpose-specific Liquid wallets with custom account indices, we'd rather align with that.

Does LWK support account index selection that we're missing, or do you have a preferred approach in the multi-wallet roadmap?

BullishNode:
We have a specific approach for multiwallet roadmap we are working on, we will keep you posted. Thanks for the PR! Let's consider that multiwallet feature is a prerequisite for Samrock

benjamin-jarvie:
@BullishNode should we close this now as this (https://github.com/rockstardev/SamRockProtocol/pull/10) got added and merged?

PR #2011, opened 2026-04-08
fix: recurrent no-api-key error spam when not logged in

ethicnology:
- Exchange repository read methods (getUserSummary, getOrders,
  getOrderByTxId, listAnnouncements) return empty results instead
  of throwing ApiKeyException when no API key is stored
- Simplified ExchangeCubit.fetchUserSummary() error handling
- Write/action methods (placeBuyOrder, saveUserPreference, etc.)
  still throw ApiKeyException as expected for login UI prompts

### DIFF
--- a/lib/features/exchange/presentation/exchange_cubit.dart
+++ b/lib/features/exchange/presentation/exchange_cubit.dart
@@ -72,9 +72,7 @@ class ExchangeCubit extends Cubit<ExchangeState> {
 
   Future<void> fetchUserSummary({bool force = false}) async {
     try {
-      emit(
-        state.copyWith(apiKeyException: null, getUserSummaryException: null),
-      );
+      emit(state.copyWith(getUserSummaryException: null));
 
       final userSummary = await _getExchangeUserSummaryUsecase.execute();
 
--- a/lib/features/exchange/presentation/exchange_cubit.dart
+++ b/lib/features/exchange/presentation/exchange_cubit.dart
@@ -1,6 +1,5 @@
 import 'dart:async';
 
-import 'package:bb_mobile/core/errors/exchange_errors.dart';
 import 'package:bb_mobile/core/exchange/domain/entity/notification_message.dart';
 import 'package:bb_mobile/core/exchange/domain/usecases/delete_exchange_api_key_usecase.dart';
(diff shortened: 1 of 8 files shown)


lib/features/exchange/presentation/exchange_cubit.dart
```diff
@@ -92,15 +91,10 @@ class ExchangeCubit extends Cubit<ExchangeState> {
...
 
       loadAnnouncements();
+    } on GetExchangeUserSummaryException catch (e) {
+      emit(state.copyWith(getUserSummaryException: e));
     } catch (e) {
       log.severe(error: e, trace: StackTrace.current);
-      if (e is ApiKeyException) {
-        emit(state.copyWith(apiKeyException: e));
-        // Disconnect WebSocket if API key is invalid
-        disconnectWebSocket();
```
i5hi, line 100: we need this to be triggered otherwise we get error logs from the websocket

lib/features/exchange/presentation/exchange_cubit.dart
```diff
@@ -92,15 +91,10 @@ class ExchangeCubit extends Cubit<ExchangeState> {
       }
 
       loadAnnouncements();
+    } on GetExchangeUserSummaryException catch (e) {
+      emit(state.copyWith(getUserSummaryException: e));
     } catch (e) {
       log.severe(error: e, trace: StackTrace.current);
-      if (e is ApiKeyException) {
-        emit(state.copyWith(apiKeyException: e));
```
i5hi, line 98: if this emit is removed then apiKeyExeption is not required in the state anymore. we also dont need to set it to null at the start of this method anymore.

lib/features/exchange/presentation/exchange_cubit.dart
i5hi: nevermind, if api key does not exist the web socket should not even connect

lib/features/exchange/presentation/exchange_cubit.dart
ethicnology: 4c0d88145db4586bcd1ef7bf2053ce179e92bd64

PR #2012, opened 2026-04-08
fix: label upsert UNIQUE constraint failure on exchange orders

ethicnology:
## Summary
- `insertOnConflictUpdate` generates `ON CONFLICT(id) DO UPDATE`
  which only handles primary key collisions
- Since labels are inserted with auto-generated ids, there is
  never a conflict on `id` but the `(label, reference)` UNIQUE
  constraint can still be violated with no handler, causing
  SEVERE errors
- Changed to `insertOrIgnore` which covers all constraint
  violations and silently skips duplicates

### DIFF
--- a/lib/features/labels/adapters/labels_repository_adapter.dart
+++ b/lib/features/labels/adapters/labels_repository_adapter.dart
@@ -15,7 +15,10 @@ class DriftLabelsRepositoryAdapter implements LabelsRepositoryPort {
     final companion = LabelMapper.newLabelEntityToCompanion(newLabel);
     final id = await _database.into(_database.labels).insert(
       companion,
-      mode: InsertMode.insertOrIgnore,
+      onConflict: DoUpdate(
+        (old) => companion,
+        target: [_database.labels.label, _database.labels.reference],
+      ),
     );
 
     return LabelEntity(
--- a/lib/features/labels/adapters/labels_repository_adapter.dart
+++ b/lib/features/labels/adapters/labels_repository_adapter.dart
@@ -13,9 +13,10 @@ class DriftLabelsRepositoryAdapter implements LabelsRepositoryPort {
   @override
   Future<LabelEntity> store(NewLabel newLabel) async {
     final companion = LabelMapper.newLabelEntityToCompanion(newLabel);
-    final id = await _database
-        .into(_database.labels)
-        .insertOnConflictUpdate(companion);
+    final id = await _database.into(_database.labels).insert(
+      companion,
+      mode: InsertMode.insertOrIgnore,
+    );
 
     return LabelEntity(
       id: id,


i5hi:
# Claude Review

## Findings

### [IMPORTANT] I3: `insertOrIgnore` returns bogus id on duplicate, creating a phantom LabelEntity

**File**: `lib/features/labels/adapters/labels_repository_adapter.dart:16-26`

**What**: When `insertOrIgnore` skips a duplicate row (matching `(label, reference)` UNIQUE constraint), Drift returns `database.lastInsertRowId` — the id from whatever the *last successful* insert was on that connection, not the id of the skipped row. Per [SQLite docs](https://www.sqlite.org/c3ref/last_insert_rowid.html): *"An INSERT that fails due to a constraint violation is not a successful INSERT and does not change the value returned by this routine."* The code then constructs a `LabelEntity` with this stale/unrelated id — a phantom entity pointing to the wrong row or a row in a different table entirely.

**Why**: The `store()` method's contract (defined in `LabelsRepositoryPort`) promises to return a valid `LabelEntity`. `StoreLabelUsecase` returns this entity to callers as an `ApplicationLabel`. If any downstream code uses that garbage id for fetching (`fetchById`) or deletion (`trash`), it will operate on the wrong row or silently fail. This is worse than a crash — it's silent data corruption. The `ImportLabelsUsecase` discards the return value so it's safe, but `StoreLabelUsecase` is not.

**Fix**: Use Drift's `DoUpdate` with an explicit conflict target on the correct columns instead:

```dart
@override
Future<LabelEntity> store(NewLabel newLabel) async {
  final companion = LabelMapper.newLabelEntityToCompanion(newLabel);
  final id = await _database.into(_database.labels).insert(
    companion,
    onConflict: DoUpdate(
      (old) => companion,
      target: [_database.labels.label, _database.labels.reference],
    ),
  );

  return LabelEntity(
    id: id,
    type: newLabel.type,
    label: newLabel.label,
    reference: newLabel.reference,
    origin: newLabel.origin,
  );
}
```

This does `ON CONFLICT(label, reference) DO UPDATE` — which is what the original `insertOnConflictUpdate` *intended* but couldn't express (it defaulted to `ON CONFLICT(id)` which never fires on an auto-increment PK). This correctly handles the UNIQUE constraint AND returns a valid row id.

Alternatively, if true "skip duplicates" semantics are desired, `insertOrIgnore` is fine but then the method must fetch the existing row on conflict:

```dart
final id = await _database.into(_database.labels).insert(
  companion,
  mode: InsertMode.insertOrIgnore,
);
if (id == -1) {
  // Row already exists, fetch it
  final existing = await fetchByReference(newLabel.reference);
  return existing.firstWhere((e) => e.label == newLabel.label);
}
```

---

### [MINOR] M12: `ImportLabelsUsecase.call` should be `execute`

**File**: `lib/features/labels/application/usecases/import_labels_usecase.dart:17`

**What**: Method is named `call` instead of `execute`.

**Why**: Convention per codebase rules — all use case methods should be named `execute`.

**Fix**: Not in scope for this PR, noting for tracking.

---

## Verdict

TOTAL: 2 findings (0 critical, 1 important, 0 architecture, 1 minor)

**REQUEST CHANGES**

The root cause diagnosis is correct — `insertOnConflictUpdate` targets the PK `id` which never collides on an auto-increment column, so the `(label, reference)` UNIQUE constraint throws unhandled. However, the chosen fix (`insertOrIgnore`) introduces a regression: on duplicate, it returns `id: -1`, producing a phantom `LabelEntity` that `StoreLabelUsecase` hands back to callers.

The correct fix is `DoUpdate` with `target: [labels.label, labels.reference]` — this performs the upsert on the right columns and always returns a valid id. Alternatively, `insertReturningOrNull()` can be used if "skip duplicates" is the intent — it returns `null` when the row is ignored, allowing the code to fetch the existing row instead.

ethicnology:
@claude i've refactored according to your comment ab9b66d010a0924fdd8c6c8ffde0dbc6a3eeda3e



cc: @i5hi

PR #2018, opened 2026-04-12
more reproducibility related changes on top of the recent refactor

basantagoswami:
- Removed the Dockerfile with verification related tools, instead of having two containers we'll use the same one. Currently for GitHub releases, we build both the AAB and APK separately, don't need to do that if we use bundletool. Also because we now have two different containers, one just containing the tools, and the other for building, it made sense for all the tools to live inside the same container
- Upgraded the apktool and bundletool versions with integrity checks for the downloaded binaries
- Renamed Dockerfile.apk to Dockerfile.build. The .apk makes it seem like an apk file called Dockerfile. Also, the Dockerfile is used to build both APKs and AABs. Same logic for the `make apk` command, renamed it to `make build`.
- Cleaned up the `verify_build.sh` script to remove some magic strings


basantagoswami:
@ethicnology made some changes on top of your branch

PR #2021, opened 2026-04-13
Fss hybrid

i5hi:
Hybrid flutter secure storage initialization strategy. 
Attempt fss10 (migrateWithBackup: true) initialization and fallback to fss9 if migration fails.


ios/Podfile.lock
ethicnology, line 1: Since you have modified your `pubpec.yaml` you may want to regenerate this file

pubspec.yaml
```diff
@@ -21,10 +21,16 @@ dependencies:
...
+      url: https://github.com/SatoshiPortal/flutter_secure_storage
+      ref: 47ad06f
+      path: flutter_secure_storage
+  flutter_secure_storage_legacy:
+    git:
+      url: https://github.com/SatoshiPortal/flutter_secure_storage_legacy
+      ref: c32db29
       path: flutter_secure_storage
   path_provider: ^2.1.5
+  shared_preferences: ^2.3.0
```
ethicnology, line 33: I see it used only in test so could be added as a `dev_dependency` instead

lib/core/storage/migrations/schema_11_to_12.dart
```diff
@@ -69,14 +69,22 @@ class Schema11To12 {
...
 
-    // MempoolServers table: add enableSsl column
-    final mempoolServers = schema12.mempoolServers;
-    await m.addColumn(mempoolServers, mempoolServers.enableSsl);
+    // MempoolServers table: add enableSsl column (idempotent)
+    try {
+      final mempoolServers = schema12.mempoolServers;
+      await m.addColumn(mempoolServers, mempoolServers.enableSsl);
+    } catch (e) {
+      if (!e.toString().contains('duplicate column')) rethrow;
```
ethicnology, line 77: I don't get why we have such errors ?

lib/core/storage/storage_locator.dart
ethicnology, line 1: Let's talk about this file tomorrow just to make sure I got the whole thing

lib/core/storage/migrations/schema_11_to_12.dart
i5hi: I think this was happening just with internal testers, will make an issue so we review it later

PR #2024, opened 2026-04-14
fix: clean lock files from bdk_flutter

ethicnology:


ethicnology:
it's a transitive dependency from satoshifier

PR #2029, opened 2026-04-15
fix: payjoin repository url

ethicnology:


ethicnology:
@i5hi this should change your pubspec.lock when you pub get

i5hi (approved):

PR #2017, opened 2026-04-11
refactor: reproducible builds from devcontainers

ethicnology:
- **Reproducible builds**: two-stage Docker build root `Dockerfile` (toolchain only my `devcontainers` works)
- **new `Dockerfile.apk`** (crafted from @basantagoswami 's `reproducibility` branch)
- **Verification tooling**: `reproducibility/verify_build.sh` downloads an official APK (GitHub or local), rebuilds from source, decodes both with apktool, and diffs pre-signature contents (from @basantagoswami 's `reproducibility` branch)
- **Makefile targets**: `make apk [release|debug]`, `make verify VERSION=x.y.z [APK=path]`

PR #2041, opened 2026-04-22
Upgrade payjoin

spacebear21:
Working branch for updating to payjoin 0.25 with native Dart bindings

### DIFF
--- a/integration_test/payjoin_test.dart
+++ b/integration_test/payjoin_test.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:bb_mobile/core/fees/domain/fees_entity.dart';
+import 'package:bb_mobile/core/payjoin/data/datasources/local_payjoin_datasource.dart';
 import 'package:bb_mobile/core/payjoin/domain/entity/payjoin.dart';
 import 'package:bb_mobile/core/payjoin/domain/repositories/payjoin_repository.dart';
 import 'package:bb_mobile/core/payjoin/domain/usecases/receive_with_payjoin_usecase.dart';
@@ -33,6 +34,7 @@ Future<void> main({bool isInitialized = false}) async {
   final addressRepository = locator<WalletAddressRepository>();
   final utxoRepository = locator<WalletUtxoRepository>();
   final payjoinRepository = locator<PayjoinRepository>();
+  final localPayjoinDatasource = locator<LocalPayjoinDatasource>();
   final receiveWithPayjoinUsecase = locator<ReceiveWithPayjoinUsecase>();
   final sendWithPayjoinUsecase = locator<SendWithPayjoinUsecase>();
   final prepareBitcoinSendUsecase = locator<PrepareBitcoinSendUsecase>();
@@ -49,6 +51,34 @@ Future<void> main({bool isInitialized = false}) async {
 
   setUpAll(() async {
     await locator<SetEnvironmentUsecase>().execute(Environment.testnet);
+
+    // Drain any persisted payjoin state so the test starts clean. Ongoing
+    // payjoins left behind by a previous (possibly crashed) run keep their
+    // inputs frozen via getUtxosFrozenByOngoingPayjoins(), which would starve
+    // the sender wallet. _resumePayjoins in PayjoinRepositoryImpl's
+    // constructor runs unawaited and writes its own updates concurrently, so
+    // we expire + recheck until the ongoing set stays empty for several polls.
+    const pollInterval = Duration(milliseconds: 500);
+    const requiredStableChecks = 3;
+    const maxIterations = 40;
+    var stableChecks = 0;
+    for (var i = 0; i < maxIterations; i++) {
+      final ongoing = await localPayjoinDatasource.fetchAll(
+        onlyUnfinished: true,
+      );
+      if (ongoing.isEmpty) {
+        stableChecks++;
+        if (stableChecks >= requiredStableChecks) break;
+      } else {
+        stableChecks = 0;
+        for (final payjoin in ongoing) {
+          await localPayjoinDatasource.update(
+            payjoin.copyWith(isExpired: true),
+          );
+        }
+      }
+      await Future.delayed(pollInterval);
+    }
 
     final receiverSeed = await seedRepository.createFromMnemonic(
       mnemonicWords: receiverMnemonic.split(' '),
--- a/integration_test/payjoin_test.dart
+++ b/integration_test/payjoin_test.dart
@@ -5,7 +5,7 @@ import 'package:bb_mobile/core/payjoin/domain/entity/payjoin.dart';
 import 'package:bb_mobile/core/payjoin/domain/repositories/payjoin_repository.dart';
 import 'package:bb_mobile/core/payjoin/domain/usecases/receive_with_payjoin_usecase.dart';
 import 'package:bb_mobile/core/payjoin/domain/usecases/send_with_payjoin_usecase.dart';
-import 'package:bb_mobile/core/seed/data/models/seed_model.dart';
+import 'package:bb_mobile/core/seed/data/repository/seed_repository.dart';
 import 'package:bb_mobile/core/settings/domain/settings_entity.dart';
 import 'package:bb_mobile/core/utils/constants.dart';
 import 'package:bb_mobile/core/wallet/data/repositories/wallet_address_repository.dart';
@@ -16,7 +16,6 @@ import 'package:bb_mobile/features/send/domain/usecases/prepare_bitcoin_send_use
 import 'package:bb_mobile/features/settings/domain/usecases/set_environment_usecase.dart';
 import 'package:bb_mobile/locator.dart';
 import 'package:bb_mobile/main.dart';
-import 'dart:io' show Platform;
 
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart' show TestWidgetsFlutterBinding;
@@ -30,6 +29,7 @@ Future<void> main({bool isInitialized = false}) async {
   late Wallet senderWallet;
 
   final walletRepository = locator<WalletRepository>();
+  final seedRepository = locator<SeedRepository>();
   final addressRepository = locator<WalletAddressRepository>();
   final utxoRepository = locator<WalletUtxoRepository>();
   final payjoinRepository = locator<PayjoinRepository>();
@@ -37,32 +37,32 @@ Future<void> main({bool isInitialized = false}) async {
   final sendWithPayjoinUsecase = locator<SendWithPayjoinUsecase>();
   final prepareBitcoinSendUsecase = locator<PrepareBitcoinSendUsecase>();
 
-  final receiverMnemonic = Platform.environment['TEST_ALICE_MNEMONIC'];
-  final senderMnemonic = Platform.environment['TEST_BOB_MNEMONIC'];
+  const receiverMnemonic = String.fromEnvironment('TEST_ALICE_MNEMONIC');
+  const senderMnemonic = String.fromEnvironment('TEST_BOB_MNEMONIC');
 
-  if (receiverMnemonic == null || receiverMnemonic.isEmpty) {
+  if (receiverMnemonic.isEmpty) {
     throw Exception('TEST_ALICE_MNEMONIC environment variable is not set');
   }
-  if (senderMnemonic == null || senderMnemonic.isEmpty) {
+  if (senderMnemonic.isEmpty) {
     throw Exception('TEST_BOB_MNEMONIC environment variable is not set');
   }
 
   setUpAll(() async {
     await locator<SetEnvironmentUsecase>().execute(Environment.testnet);
 
-    final receiverSeedModel = SeedModel.mnemonic(
+    final receiverSeed = await seedRepository.createFromMnemonic(
       mnemonicWords: receiverMnemonic.split(' '),
     );
-    final senderSeedModel = SeedModel.mnemonic(
+    final senderSeed = await seedRepository.createFromMnemonic(
       mnemonicWords: senderMnemonic.split(' '),
     );
     receiverWallet = await walletRepository.createWallet(
-      seed: receiverSeedModel.toEntity(),
+      seed: receiverSeed,
       network: Network.bitcoinTestnet,
       scriptType: ScriptType.bip84,
     );
     senderWallet = await walletRepository.createWallet(
-      seed: senderSeedModel.toEntity(),
+      seed: senderSeed,
       network: Network.bitcoinTestnet,
       scriptType: ScriptType.bip84,
     );
--- a/lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
+++ b/lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
@@ -636,28 +636,34 @@ class PdkPayjoinDatasource {
       final receiverModel = PayjoinReceiverModel.fromJson(
         data as Map<String, dynamic>,
       );
-      final receiver = Receiver.fromJson(json: receiverModel.receiver);
 
       // Start checking for a payjoin request from the sender periodically
       Timer.periodic(const Duration(seconds: PayjoinConstants.directoryPollingInterval), (
         Timer timer,
       ) async {
-        log(
-          '[Receivers Isolate] Checking for request in receivers isolate for '
-          '${receiver.id()}',
-        );
+        log('[Receivers Isolate] Checking for request for ${receiverModel.id}');
         try {
-          final request = await getRequest(receiver: receiver, dio: dio);
-          if (request != null) {
-            requests.putIfAbsent(receiver.id(), () async {
-              log(
-                '[Receivers Isolate] Request found in receivers isolate for '
-                '${receiver.id()}',
-              );
+          final persister = InMemoryJsonReceiverSessionPersister.fromJson(
+            receiverModel.receiver,
+          );
+          final state = pj.replayReceiverEventLog(persister: persister).state();
+          if (state is! pj.InitializedReceiveSession) return;
+
+          final unchecked = await _getUncheckedOriginalPayload(
+            state.inner,
+            persister,
+            dio,
+          );
+          if (unchecked != null) {
+            requests.putIfAbsent(receiverModel.id, () async {
+              log('[Receivers Isolate] Request found for ${receiverModel.id}');
+              final maybeInputsOwned = unchecked
+                  .assumeInteractiveReceiver()
+                  .save(persister: persister);
               // The original tx bytes are needed in the main isolate for
-              //  further processing so extract them here and pass them through
-              //  the model
-              final originalTxBytes = await request
+              //  further processing so extract them here and pass them
+              //  through the model
+              final originalTxBytes = maybeInputsOwned
                   .extractTxToScheduleBroadcast();
               final originalTx = await BitcoinTx.fromBytes(originalTxBytes);
               final originalTxId = originalTx.txid;
@@ -667,12 +673,12 @@ class PdkPayjoinDatasource {
               );
               log(
                 '[Receivers Isolate] Request original Tx ID: $originalTxId and amount: $amountSat for '
-                '${receiver.id()}',
+                '${receiverModel.id}',
               );
               final updatedModel = receiverModel.copyWith(
-                receiver: receiver.toJson(),
+                receiver: persister.toJson(),
                 originalTxBytes: originalTxBytes,
-                originalTxId: originalTxId,
+                originalTxId: originalTx.txid,
                 amountSat: amountSat,
               );
 
@@ -681,23 +687,22 @@ class PdkPayjoinDatasource {
 
               // Cancel the timer since the request has been received
               log(
-                '[Receivers Isolate] cancelling timer in receivers isolate for ${receiver.id()}',
+                '[Receivers Isolate] cancelling timer in receivers isolate for ${receiverModel.id}',
               );
               timer.cancel();
               log(
-                '[Receivers Isolate] timer cancelled in receivers isolate for ${receiver.id()}',
+                '[Receivers Isolate] timer cancelled in receivers isolate for ${receiverModel.id}',
               );
             });
           } else {
             log(
               '[Receivers Isolate] No valid request found in receivers isolate for '
-              '${receiver.id()}',
+              '${receiverModel.id}',
             );
           }
         } catch (e) {
           log(
-            '[Receivers Isolate] periodic timer get request exception: $e for '
-            '${receiver.id()}',
+            '[Receivers Isolate] periodic timer get request exception: $e for ${receiverModel.id}',
           );
           if (e is PayjoinExpiredException) {
             // If the request returns an expiry error, mark the receiver as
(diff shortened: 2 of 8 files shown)


BullishNode:
Is payjoin 0.25 the new release number for what was 1.0.0 before?

pubspec.lock
```diff
@@ -1488,15 +1472,13 @@ packages:
...
-      path: "."
-      ref: "913812dba120703ce9738fa330c5203e1d11f2fa"
-      resolved-ref: "913812dba120703ce9738fa330c5203e1d11f2fa"
-      url: "https://github.com/SatoshiPortal/payjoin-dart"
-    source: git
-    version: "0.23.0"
+      path: "../rust-payjoin/payjoin-ffi/dart"
+      relative: true
+    source: path
+    version: "0.1.0"
```
DanGould: rust-payjoin@993f3b8b240e29d46c5ecc1eb9d11958b3b2002a: sources

DanGould:
We've got 1.0 rc's but still have some TODOs on [the milestone](https://github.com/payjoin/rust-payjoin/milestone/1). The rcs aren't discoverable on docs.rs, so we decided to cut another minor release and continue to do so until we're way more certain 1.0 is right around the corner (like 1 week away).

ethicnology:
Hello guys, thanks for the contribution 

I've had to fork [bdk-dart](https://github.com/ethicnology/bdk-dart/tree/fix/reproducible-build) to target my own fork of uniffi that include [the fix for reproducible builds](https://github.com/mozilla/uniffi-rs/pull/2873), this fix is merged upstream.

Anytime soon we should get a new tag/release from uniffi that include it and will be required for our native assets dependencies

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -12,6 +12,7 @@ import 'package:bb_mobile/core/utils/constants.dart';
 import 'package:bb_mobile/core/utils/logger.dart' as logger;
 import 'package:crypto/crypto.dart';
 import 'package:dio/dio.dart';
+import 'package:payjoin/http.dart' show fetchOhttpKeys;
```
DanGould, line 15: This is OK. But if it's possible to do with a dart client, that PR could land independently and delete some code here.

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -87,42 +87,32 @@ class PdkPayjoinDatasource {
...
 
-      final receiver = await Receiver.load(
-        token: token,
-        persister: noOpPersister,
-      );
+      final persister = InMemoryJsonReceiverSessionPersister();
+      final initialized = receiverBuilder.build().save(persister: persister);
+      final pjUri = initialized.pjUri().asString();
+      // Derive the receiver ID from pjUri
+      final id = sha256.convert(utf8.encode(pjUri)).toString().substring(0, 16);
```
DanGould, line 94: This seems like it may be both a regression and divergence from the state of the art in our reference implementation. There we're generating session IDs from random because it's possible that multiple attempts could be made with the same URL. At least I think that's what the rationale is here: 

https://github.com/payjoin/rust-payjoin/blob/993f3b8b240e29d46c5ecc1eb9d11958b3b2002a/payjoin-cli/src/db/v2.rs#L108-L120

I *think* if the ID were to come from the persister implementation after the `.save` call that might match, but I do understand that it's also imperative messiness to expect that state after `save` rather than return it from `.save()`. I question why the model needs this to begin with.

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -87,42 +87,32 @@ class PdkPayjoinDatasource {
...
           PayjoinModel.receiver(
-                id: receiver.id(),
+                id: id,
                 address: address,
                 isTestnet: isTestnet,
-                receiver: receiver.toJson(),
+                receiver: persister.toJson(),
                 walletId: walletId,
-                pjUri: (await receiver.pjUri()).asString(),
+                pjUri: pjUri,
```
DanGould, line 103: I wonder why ~`receiver: persister` and why `pjUri` is saved rather than the whole `receiver`. This doesn't seem like a clean change, but it may just be an intermediate step still (if so, a FIXME comment in the in-between code that gets removed in the later commit in the same PR would make this easier to review).

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -133,43 +133,37 @@ class PdkPayjoinDatasource {
...
-    final token = await newSender.persist(persister: persister);
-    final sender = await Sender.load(token: token, persister: persister);
-    final senderJson = sender.toJson();
+    var sendBuilder = pj.SenderBuilder(psbt: originalPsbt, uri: pjUri);
+    final persister = InMemoryJsonSenderSessionPersister();
+    final minFeeRateSatPerKwu = (networkFeesSatPerVb * 250).round();
+    pj.WithReplyKey? withReplyKey;
+    try {
+      withReplyKey = sendBuilder
+          .buildRecommended(minFeeRate: minFeeRateSatPerKwu)
```
DanGould: nit: if this minFeeRate parameter is not strongly typed internalizing units then the unit belongs in the name of the parameter. I see this is not a regression but it is something for us to note before merge.

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -133,43 +133,37 @@ class PdkPayjoinDatasource {
...
     // Create and store the model with the data needed to keep track of the
-    //  payjoin session
+    // payjoin session
     final model =
         PayjoinModel.sender(
-              uri: uri.asString(),
+              uri: parsedUri.asString(),
               isTestnet: isTestnet,
-              sender: senderJson,
+              sender: persister.toJson(),
```
DanGould, line 159: same question as receive side. Why is `sender: persister` and if we're passing the real `sender` why even pass the `uri` as a separate field?

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -185,6 +179,39 @@ class PdkPayjoinDatasource {
...
+        log('sender v2 post via $relay failed: $e');
+        lastError = e;
+        continue;
+      }
+    }
+    if (!posted) {
+      throw SendCreationException(
+        'Failed to post original PSBT to any OHTTP relay: $lastError',
+      );
+    }
```
DanGould, line 205: I think this is a slight difference in functionality but it's not bad, just worth calling out. Only saying the final OHTTP relay errored might be a source of confusion.

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -214,105 +214,328 @@ class PdkPayjoinDatasource {
...
+    required bool Function(Uint8List) hasOwnedInputs,
+    required bool Function(Uint8List) hasReceiverOutput,
+    required List<PayjoinInputPairModel> inputPairs,
+    required PayjoinReceiverModel receiverModel,
+    required String Function(String) processPsbt,
+  }) async {
+    if (state is pj.InitializedReceiveSession) {
+      throw StateError('Original PSBT is retrieved in the receiver isolate');
+    } else if (state is pj.UncheckedOriginalPayloadReceiveSession) {
+      return _checkProposal(
```
DanGould: Doesn't dart have [`switch` statements as of 3.0](https://dart.dev/language/branches#switch-statements) to make this pretty?

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -200,107 +179,363 @@ class PdkPayjoinDatasource {
...
+    PayjoinReceiverModel receiverModel,
+    String Function(String) processPsbt,
+  ) async {
+    final candidates = inputPairs.map(_buildInputPair).toList();
+    pj.InputPair? chosen;
+    try {
+      chosen = inner.tryPreservingPrivacy(candidateInputs: candidates);
+    } catch (e) {
+      throw StateError('No inputs available to contribute to payjoin');
+    }
```
DanGould, line 455: This is slightly different behavior than the branch you're building off of. BBMobile Asked us to use `candidateInputs.first` on this error condition rather than throw an error for the time being.

see https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2041/changes#diff-6d93fc7fa95c1bb7e88cbe0548c88acbe007bbaad6aa51a867df5619613158baL261-L263

separating behavior changes from the upgrade

pubspec.yaml
```diff
@@ -31,10 +31,8 @@ dependencies:
       path: flutter_secure_storage
   path_provider: ^2.1.5
   shared_preferences: ^2.3.0
-  payjoin_flutter:
-    git:
-      url: https://github.com/SatoshiPortal/payjoin-dart
-      ref: 913812dba120703ce9738fa330c5203e1d11f2fa
+  payjoin:
+    path: ../rust-payjoin/payjoin-ffi/dart
```
DanGould:
```suggestion
payjoin:
  git:
    url: https://github.com/payjoin/rust-payjoin
    ref: 993f3b8b240e29d46c5ecc1eb9d11958b3b2002a
    path: payjoin-ffi/dart
```

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -50,25 +49,21 @@ class PdkPayjoinDatasource {
...
-          payjoinDirectory: payjoinDirectory,
+        final ohttpKeys = await fetchOhttpKeys(
+          ohttpRelayUrl: ohttpRelayUrl,
+          directoryUrl: payjoinDirectory,
         );
-        ohttpRelay = relay;
-        break;
+        return (ohttpKeys, ohttpRelayUrl);
       } catch (e) {
         continue;
```
DanGould, line 56: prevent swallowing the errors? `log('fetchOhttpKeys via $ohttpRelayUrl failed: $e')` I don't think this is the only place we're losing error info, but it's not a regression so it could be info overload.

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
```diff
@@ -549,271 +788,215 @@ class PdkPayjoinDatasource {
...
-      data: req.body,
-      options: Options(
-        headers: {'Content-Type': req.contentType},
-        responseType: ResponseType.bytes,
-      ),
-    );
-    log('[Senders Isolate] Received response from ${req.url.asString()}');
+class OhttpRelaysUnavailableException extends BullException {
+  OhttpRelaysUnavailableException(super.message);
+}
```
DanGould, line 780: nit: ever thrown?

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
spacebear21: fetchOhttpKeys is implemented with the native dart:io http client https://github.com/payjoin/rust-payjoin/pull/1495/changes fwiw. I'm not sure I understand what code would be delete-able, wouldn't it be a net add to implement fetchOhttpKeys from scratch?

Separately, do you think [this undertaking](https://github.com/payjoin/rust-payjoin/issues/1362) to expose fetchOhttpKeys in each supported downstream language is misguided? My concern with encouraging bespoke implementations is that it introduces privacy footguns, e.g. implementers might just fetch from the directory directly without proxying through the relay.

DanGould: i forgot this had been implemented in dart and wasn't using the rust version. What do you have right here Seems like the best of both worlds not misguided .

DanGould:
What flutter version are you on? I'm on 
```dan@Excavator bullbitcoin-mobile % fvm flutter --version
Flutter 3.38.5 • channel stable • https://github.com/flutter/flutter.git
Framework • revision f6ff1529fd (4 months ago) • 2025-12-11 11:50:07 -0500
Engine • hash c108a94d7a8273e112339e6c6833daa06e723a54 (revision 1527ae0ec5) (4 months ago) • 2025-12-11 15:04:31.000Z
Tools • Dart 3.10.4 • DevTools 2.51.1
```
but getting what seem to be flutter version-related failures when I build. I think bbmobile pins 3.38.5

trying the android emulator locally and hitting exhaustive-switch errors in recoverbull/recipients/key_server_status_widget (amongst many other issues) against Flutter 3.38.5.

ethicnology:
@DanGould `3.38.5` is the correct version, what kind of error are you experiencing?

You may experience a global unhandled error related to Boltz testnet deprecated [that will be fix](https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2042)

DanGould:
First off, @ethicnology I was trying to build on 2025 `.freezed` files which caused those errors. My bad! not a problem.

Second
I ran integration_test/payjoin_test.dart on Android (emulator-5554, API 35 arm64) against d8a9e849: 3 pass (error paths), 2 fail (happy path). Robot suggests the happy path is blocked by a bdk_dart/bdk-ffi rustls init issue, not payjoin-ffi.

these passed afaict:

02:07 +1 -2: Payjoin Integration Tests with one receive and one send should fail if the receiver does not have enough
02:14 +2 -2: Payjoin Integration Tests with one receive and one send should fail if the sender does not have enough 
02:21 +3 -2: Payjoin Integration Tests with one receive and one send should expire if time to wait for a request is over

These are the problems I faced running the test:

Android blocker No. 1: ElectrumClient::new fails with Failed to install CryptoProvider — rustls/TLS init issue in bdk_dart/bdk-ffi v2.3.1, sync never succeeds.

<details>
<summary>Log</summary>

```console
2026-04-24T14:30:44.991269      SEVERE  sync wallet error with electrum server: ssl://blockstream.info:993      CouldNotCreateConnectionElectrumException(unexpected error: Faile
d to install CryptoProvider)    #0      checkCallStatus (package:bdk_dart/bdk.dart:24814:5)
#1      rustCall (package:bdk_dart/bdk.dart:24837:5)
#2      new ElectrumClient (package:bdk_dart/bdk.dart:20680:14)
#3      _performFullScan (package:bb_mobile/core/wallet/data/datasources/bdk_wallet_datasource.dart:717:26)
<asynchronous suspension>
#4      _RemoteRunner._run (dart:isolate:1118:18)
<asynchronous suspension>
```

</details>

Android blocker No. 2: first happy-path test throws SeedNotFoundException('Seed not found for fingerprint: 858ab5df') from send_with_payjoin_usecase.dart:45, after 4 null reads over ~4.5s from secure storage. Same fingerprint reads fine in subsequent tests. Mechanism unclear — may be a persistence gap in walletRepository.createWallet for the second wallet, or a storage timing issue/race?


<details>
<summary>Log</summary>

```console
2026-04-24T14:31:00.931014      FINE    Seed read returned null for fingerprint 858ab5df on attempt 1, retrying in 300ms
2026-04-24T14:31:01.248030      FINE    Seed read returned null for fingerprint 858ab5df on attempt 2, retrying in 600ms
2026-04-24T14:31:01.852582      FINE    Seed read returned null for fingerprint 858ab5df on attempt 3, retrying in 1200ms
2026-04-24T14:31:03.067069      FINE    Seed read returned null for fingerprint 858ab5df on attempt 4, retrying in 2400ms
01:54 +0 -2: Payjoin Integration Tests with one receive and one send should work with one receiver and one sender [E]
  Seed not found for fingerprint: 858ab5df
  package:bb_mobile/core/payjoin/domain/usecases/send_with_payjoin_usecase.dart 45:7                                     SendWithPayjoinUsecase.execute
  ===== asynchronous gap ===========================
  package:stream_channel                                                                                                 _GuaranteeSink.add
  var/folders/f8/bx_b4w_90hl17m6ynl85nc0r0000gn/T/flutter_tools.5pAZVh/flutter_test_listener.45KaLB/listener.dart 56:22  main.<fn>
```

</details>

ethicnology:
@DanGould

Can you attempt a `make setup` this should clear old files and prepare your local repo to compile for android.

Also, on develop we have `devcontainer` now, so the whole repo tooling is containerized and workable from this container if you want a setup without polluting your OS. It does work on amd64, if you work on arm64 it will be super slow.

integration_test/payjoin_test.dart
benalleng: Nit: Can be removed now that we are using `FromEnvironment`

benalleng (commented):
My test is hanging at the process proposal step but I cannot really find a good way to expose the error that is happening as all we get is _did not complete_
```
2026-04-29T09:37:18.846543      INFO    Processing payjoin request: 4a25f4cb99a44221
01:19 +1: Payjoin Integration Tests with one receive and one send should work with one receiver and 
one sender
Payjoin event for 4a25f4cb99a44221: PayjoinStatus.requested
01:20 +1: Payjoin Integration Tests with one receive and one send should work with one receiver and 
one sender - did not complete [E]
```

~To get to this point I also had to completely remove the proxy behavior in the http.dart on the rust-payjoin side and just get the keys from the directory directly.~ I missed that is the point of https://github.com/payjoin/rust-payjoin/pull/1500

DanGould:
> Can you attempt a make setup this should clear old files and prepare your local repo to compile for android.

I manually got rid of freezed files which did it but will do this in the future

> Also, on develop we have devcontainer now, so the whole repo tooling is containerized and workable from this container if you want a setup without polluting your OS. It does work on amd64, if you work on arm64 it will be super slow.

arm64 Apple Silicon baby

got tests running based on your spec @spacebear21 but need human-in-the-loop to get into this PR https://github.com/DanGould/bullbitcoin-mobile/commit/72b0b65ec603057ae0facf1b8392d0c99662751b

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
spacebear21: `try_preserving_privacy` internally falls back to the first candidate input for convenience if no UIH-defeating input is found, so the behavior is the same regardless. This may be a footgun for implementers though, not sure that we have the right trade-off here.

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
spacebear21: `pjUri` is a legacy field I didn't want to touch here yet so yeah I consider this an intermediate step until the Receiver/Sender Models are refactored for proper session event log support.

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
spacebear21: same as https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2041#discussion_r3184774282

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
spacebear21: also adding here as a reminder: there are a few places in the FFI API with awkward naming, particularly the InputPair-related structs: `PlainTxIn`, `PlainOutPoint`, `PlainPsbtInput`, `PlainTxOut`. These are primitive representations of bitcoin types which are then validated in payjoin-ffi, but those names are confusing from an outsider perspective. I think we could just drop the `Plain` prefix and make it clear they expect primitive types in the respective fields.

lib/core/payjoin/data/datasources/pdk_payjoin_datasource.dart
spacebear21: The model itself probably doesn't need it, seems like it could just be a DB concern to auto-generate primary keys for pj sessions. I think the same answer as https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2041#discussion_r3184774282 applies here.

spacebear21:
> got tests running based on your spec @spacebear21 but need human-in-the-loop to get into this PR https://github.com/DanGould/bullbitcoin-mobile/commit/72b0b65ec603057ae0facf1b8392d0c99662751b

cherry-picked into this PR and confirmed all payjoin tests are passing locally

DanGould (commented):
As far as I can tell the review back-and-forth between @spacebear21 and I is satisfied and we need feedback from Bull Bitcoin @ethicnology. While this passes tests and would *work*, we probably want to follow this up with a polish to the sender/receiver models, as noted by @spacebear21, and are ultimately blocked from hitting the merge button on this by an upstream PR https://github.com/payjoin/rust-payjoin/pull/1500.

spacebear21:
Updated the payjoin dependency to use the published package at https://pub.dev/packages/payjoin instead of a local path

integration_test/payjoin_test.dart
benalleng, line 193: We really should reconsider this as the feerate. whether here or as a followup in #1978

BullishNode:
Only issues worth mentioning found the BULL review skill 

 1. **Fix Payjoin polling concurrency/recovery**

     `Timer.periodic` can overlap async polls, and polling is cancelled before
  the updated model is safely handed off. Add per-session in-flight guards or
  use a self-scheduling loop, and avoid silently stranding sessions on local
  processing errors.

  2. **Handle old persisted Payjoin sessions explicitly**

     The persisted session format changes, but unknown/legacy JSON currently
  decodes to an empty event list. Either migrate old unfinished sessions,
  deliberately expire them, or otherwise implement an explicit upgrade policy.

  3. **Fix the Payjoin integration test config**

     The test now uses `String.fromEnvironment`, while the repo setup still
  provides runtime env vars / `.env`. Restore `Platform.environment` or update
  the runner/docs to pass `--dart-define`.

benalleng:
> Fix the Payjoin integration test config

> The test now uses String.fromEnvironment, while the repo setup still
provides runtime env vars / .env. Restore Platform.environment or update
the runner/docs to pass --dart-define.

The commit de44c1b3100cb6125cd37f9e0d0ea3caf8dae316 from #2017 removed the use of env vars but the payjoin integration test relies on these to complete tests, is there a flutter native way to add env values to the local testing emulator as it seems that `Platform.environment` does not properly get loaded in the runner where using the env + --dart-define still works.

benalleng:
> Handle old persisted Payjoin sessions explicitly

> The persisted session format changes, but unknown/legacy JSON currently
decodes to an empty event list. Either migrate old unfinished sessions,
deliberately expire them, or otherwise implement an explicit upgrade policy.

I assume this is referring to existing payjoins during the upgrade? @spacebear21 do you have any opinions on handling these upgrades? Is it expected that this may cause some error within the wallet or will these silently be ignored?

ethicnology:
Sorry guys, absorbed by others topics these days.

Regarding integrations test, you cannot run them in the CI as of now since they have to spin in an emulator BUT, last weekend I've worked BULL support on Linux desktop with that need in mind. Soon we will have a CI with integrations tests running.

We have some mnemonic stored as secrets in the runner that should hold some testnet coins, that we were using for the payjoin integration tests.

Could we prioritize this PR in the next sprint @thibistaken ?

spacebear21:
> I assume this is referring to existing payjoins during the upgrade? @spacebear21 do you have any opinions on handling these upgrades? Is it expected that this may cause some error within the wallet or will these silently be ignored?

Yes, legacy sessions would currently fail silently upon attempting and failing to replay the session event log. The suggestion to "deliberately expire them" sounds reasonable but I'm not sure it's actually worth the added complexity since the hardcoded expiration is only 24 hours. Pre-upgrade sessions would simply remain in a un-resumable state until they naturally expire within that time window (if they haven't expired already).

ethicnology:
Integration tests running in the CI are now merged on develop.

I've commented out the old payjoin tests, but rebasing on develop could let you try this new pipeline

ethicnology:
reminder for me we should target https://github.com/mozilla/uniffi-rs/releases/tag/v0.31.2

ethicnology:
PR created https://github.com/Uniffi-Dart/uniffi-dart/pull/146

Then we will need to update the dependency in payjoin and bdk (out of the scope of this PR)

cc: @i5hi

ethicnology:
Thanks to @spacebear21 https://github.com/Uniffi-Dart/uniffi-dart/pull/146 got merged, when a new version get released we should PR bdk and payjoin to update the dependency

ethicnology:
Hey @spacebear21 I was wondering what's left ?

ethicnology:
Creating a new branch so collaborators can work directly on it

spacebear21:
> Hey @spacebear21 I was wondering what's left ?

Sorry for the late response, last week I was focused on getting https://crates.io/crates/payjoin/1.0.0-rc.4 ready and published. We've already upgraded payjoin-ffi to use the latest uniffi-dart version with uniffi 0.31.2 for the bindings generator. The last thing to do on the payjoin side is to actually publish these bindings to pub.dev so that they can be consumed here - I'll aim to get that done early this week. The changes from payjoin 0.25 are minimal, mostly renames and more comprehensive error handling, so I don't expect this integration to need many more changes.

PR #2044, opened 2026-04-23
add --ulimit nofile=65536:65536 to makefile

basantagoswami:


ethicnology (approved):
should avoid failure for ubuntu hosts

basantagoswami:
Ubuntu systems fail with a "too many open files" error without this option

PR #2054, opened 2026-04-29
fss-10 migrateWithAlgorithmChange: false - with fallback to legacy

i5hi:


ios/Podfile.lock
ethicnology, line 1: shouldn't be deleted

pubspec.lock
ethicnology, line 1: should be discarded

lib/core/storage/storage_locator.dart
ethicnology, line 1: This simplify drastically the migration.

Do we still need log.shout ?

ethicnology (commented):
This simplify drastically the migration.

Do we still need log.shout ?

i5hi:
We now skip migrations entirely, for those users who are on fss10 this is fine. For those who are on fss9 esp: true, we fallback to our legacy library and show them the warning.

i5hi (commented):
tACK

PR #2077, opened 2026-05-02
fix: reintroduce timeout and retry for electrum client using updated bdk-dart

i5hi:
Introduces timeout and retry values in electrum client. Sets batchSize value based on stopGap.

### DIFF
--- a/lib/core/blockchain/data/datasources/bdk_bitcoin_blockchain_datasource.dart
+++ b/lib/core/blockchain/data/datasources/bdk_bitcoin_blockchain_datasource.dart
@@ -32,12 +32,13 @@ class BdkBitcoinBlockchainDatasource {
   ) async {
     final blockchain = bdk.ElectrumClient(
       url: electrumServer.url,
-      // Only set the socks5 if it's not empty,
-      //  otherwise bdk will throw an error
-      // TODO: this was in bdk_flutter, check if it's still needed in bdk_dart
       socks5: electrumServer.socks5?.isNotEmpty == true
           ? electrumServer.socks5
           : null,
+      // electrum-client caps timeout/retry at u8 (255s, 255 retries).
+      timeout: electrumServer.timeout.clamp(0, 255),
+      retry: electrumServer.retry.clamp(0, 255),
+      validateDomain: electrumServer.validateDomain,
     );
 
     return blockchain;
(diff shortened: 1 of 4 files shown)


ethicnology (changes requested):
Periodic tasks must align with the AppLifecycle

When the user leaves the app without killing it (home button, app switcher, lock screen), the OS suspends the Dart isolate within seconds. Consequences:

- `Timer.periodic` callbacks keep firing on a frozen isolate → Electrum sockets throw (`SocketException`, `TimeoutException`) as soon as the app resumes, or worse, accumulate zombie connections.
- `Stream.periodic` and `await Future.delayed` loops have the same issue: the Dart scheduler isn't running.
- Without explicit handling, every resume stacks new Electrum connections → resource leak, server-side rate limiting, stale UI on return.

The simple fix: **cancel the timer on `paused`, recreate it on `resumed`**, and trigger an immediate re-sync on return. Don't fight the OS, align with it.

## Example

```dart
class WalletSyncService with WidgetsBindingObserver {
  WalletSyncService({required this.onSync});
  final Future<void> Function() onSync;

  Timer? _timer;
  DateTime? _lastSync;

  void start() {
    WidgetsBinding.instance.addObserver(this);
    _schedule();
    _syncNow();
  }

  void dispose() {
    WidgetsBinding.instance.removeObserver(this);
    _timer?.cancel();
  }

  @override
  void didChangeAppLifecycleState(AppLifecycleState state) {
    if (state == AppLifecycleState.resumed) {
      _schedule();   // recreate the timer
      _syncNow();    // refresh immediately
    } else {
      _timer?.cancel(); // paused / inactive / hidden / detached
      _timer = null;
    }
  }

  void _schedule() {
    _timer?.cancel();
    _timer = Timer.periodic(const Duration(seconds: 60), (_) => _syncNow());
  }

  Future<void> _syncNow() async {
    // Debounce: avoid hammering Electrum on rapid resumes
    if (_lastSync != null &&
        DateTime.now().difference(_lastSync!) < const Duration(seconds: 30)) {
      return;
    }
    try {
      await onSync();
      _lastSync = DateTime.now();
    } catch (e) {
      debugPrint('Sync failed: $e');
    }
  }
}
```

**Key points:** cancelling on `paused` prevents exceptions on return, the debounce avoids a connection storm on rapid resumes, and the `try/catch` keeps a network error from crashing the app. If you actually need polling *while* backgrounded (e.g. notifying on tx confirmation), you'll need `flutter_background_service` or FCM/APNs — `AppLifecycleState` alone isn't enough for that case.

i5hi:
https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2081

@ethicnology Issue created; will address this in a separate PR

ethicnology (approved):

PR #2085, opened 2026-05-05
Electrum bdk dart

i5hi:
Max Stop Gap: 5000
Upper limit for batchSize: 1000

Warning to keep app open in foreground for large gaps.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -346,10 +346,10 @@ packages:
     dependency: transitive
     description:
       name: characters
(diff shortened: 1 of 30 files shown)


pubspec.lock
ethicnology, line 1: this should be discarded

ethicnology (changes requested):
discard the lock file and we are good to go

pubspec.lock
i5hi: done

ethicnology (approved):

PR #2093, opened 2026-05-07
refactor(pull-to-refresh): standardize via BBPullableBody, fix small-screen reach

ethicnology:
## Summary

Pull-to-refresh was inconsistent across the app and unreliable on small screens. The shared `BBRefreshIndicator` was correct in isolation, but each screen wired its own scroll widget, physics, and bottom-buttons layout, so the gesture only worked in part of the screen, and small phones required an awkwardly long pull. Some screens were missing pull-to-refresh entirely.

This PR introduces a single `BBPullableBody` widget that enforces the three invariants pull-to-refresh actually needs:

1. A single scrollable that fills the viewport (`CustomScrollView` + `SliverFillRemaining`).
2. `AlwaysScrollableScrollPhysics` so short content can still overscroll and trigger refresh.
3. No nested scrollables (lists inside the body opt into a `sliver: true` mode).

`BBPullableBody` accepts a list of `slivers` and an optional `bottomChild` that auto-pins to the bottom of the viewport via `SliverFillRemaining(hasScrollBody: false)`. Future screens get correct gesture coverage by default.

`BBRefreshIndicator` now centers the spinner around 40% of screen height (via `MediaQuery.sizeOf`) so it lands in a consistent, reachable spot on every device.

## Test plan

Pull from the **middle of the screen** (and verify spinner lands ~40% from top) on:

- [x] Wallet home — pull, verify wallets refresh; bottom action buttons stay pinned
- [x] Wallet detail — open any wallet, pull from over the tx list area (was the worst case before)
- [x] Transactions — empty filter result, populated list, with active filter; pull from filter row, middle, bottom
- [x] Service status — initial spinner appears on open; subsequent pull triggers re-check
- [x] Exchange home (logged in) — pull, verify user summary refresh; bottom buttons stay pinned
- [x] Electrum settings — pull, spinner persists until load completes (no more instant dismiss)
- [ ] ~~Ark wallet detail — pull, verify ark cubit reloads; bottom buttons stay pinned~~
- [x] Mempool settings — pull, verify both default and custom server status indicators flicker to "checking" then update (same effect as the small refresh icon next to each server)

Verify on a small and big device that the pull distance feels comfortable.

### DIFF
--- a/lib/core/widgets/bb_pullable_body.dart
+++ b/lib/core/widgets/bb_pullable_body.dart
@@ -1,0 +1,55 @@
+import 'package:bb_mobile/core/widgets/bb_refresh_indicator.dart';
+import 'package:flutter/material.dart';
+
+/// Standard pull-to-refresh body. Use this for any screen that supports
+/// pull-to-refresh — it enforces the three invariants that make the gesture
+/// reliable across the whole screen:
+///
+/// 1. Single scrollable (CustomScrollView) so notifications never fight with
+///    nested ListView/SingleChildScrollView.
+/// 2. `AlwaysScrollableScrollPhysics` so short content can still overscroll
+///    and trigger refresh.
+/// 3. `SliverFillRemaining(hasScrollBody: false)` appended so the scrollable
+///    fills the viewport — the pull gesture works from anywhere on screen,
+///    including any [bottomChild] that gets pinned to the bottom.
+///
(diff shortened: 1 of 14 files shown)


ethicnology:
@kiranmetri @wired-pasteque 

Need your feedback on this one

i5hi (approved):

PR #2097, opened 2026-05-08
Remove recoverbull sync page

i5hi:


i5hi:
Part of #2100 
This PR removes the old hacky scan screen in favor of being replaced by a properly designed flow for this usecase.

Issue #2000, opened 2026-03-31
Fallback to alternative electrum servers when primary fails functionally


ethicnology:

ethicnology:
need your opinion on this one

thibistaken:
Can you pls test this @ethicnology ?

ethicnology:
@thibistaken 

Tested ✔️ 
- Broke the **1st** server (fulcrum) → wallet synced via the **2nd** (wes) ✅
- Broke the **1st + 2nd** (fulcrum, wes) → synced via the **3rd** (blockstream) ✅
- Broke **all 3** defaults → sync failed: *"All default servers failed (3 tried)"* ✅
- Added a **valid custom** server → synced through it, defaults ignored ✅
- Broke **only the custom** (defaults left healthy) → sync failed: *"All custom servers failed (1 tried)"* — **refused to fall back to defaults** (privacy rule) ✅

PR #2139, opened 2026-05-19
Swap electrum url

### LINKED ISSUE #1937: Boltz swap claim/refund ignores user-configured electrum servers
All swap claims and refunds use Electrum URLs baked into the swap object at creation time (hardcoded to BB or public servers) instead of the user's configured Electrum servers. Users with custom Electrum servers have no privacy guarantee for swap claim/refund transactions.

i5hi:
Fixes https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1937. 
Includes usage of new hex architecture with Port & Facade. Requires review @kumulynja and @ethicnology


i5hi:
Fixes https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1937

ethicnology:
@Claude preliminary review please

BullishNode:
Ran the six-agent review pass and reconciled the findings against `upstream/develop`. I’m intentionally not including the “testnet repository removed” finding because that came from comparing against a stale local `develop`; PR #2139 itself is the 10-file diff shown by GitHub.

Findings:

1. **Custom TCP Electrum servers are forced to TLS**

   `lib/core/electrum/electrum_facade.dart:54`

   `ElectrumFacade.getPreferredServer()` always returns `tls: true`, but Bitcoin custom servers can be stored as `tcp://...` when SSL is disabled. Swap claim/refund can therefore use the wrong transport and fail for users who deliberately configured non-TLS Electrum.

   Fix: derive `tls` from the selected server URL/protocol and normalize the URL for `bull_sdk.ElectrumSettings`.

2. **Electrum resolution silently bypasses user settings**

   `lib/core/swaps/data/services/swap_watcher.dart:41`

   `_resolveElectrumSettings()` catches every exception and returns `null`. That makes claim/refund fall back to SDK defaults on DB, DI, parsing, or settings errors, with no signal to the user. This is bad for reliability and privacy expectations.

   Fix: only return `null` for an explicit “no configured server” case. Log and fail/retry real resolution errors instead of silently changing networking policy.

3. **New Electrum selection policy duplicates the existing use case**

   `lib/core/electrum/electrum_facade.dart:37`

   The new facade reimplements server selection already handled by `GetElectrumServersToUseUsecase`, but drops app-level proxy fields such as Tor/proxy state. That creates a second source of truth for “which Electrum server should we use.”

   Fix: delegate through the existing Electrum application use case, or extract one SDK-ready Electrum resolver used by both the Electrum feature and swaps.

4. **Refund paths use current app environment for new Electrum resolution**

   `lib/core/swaps/data/services/swap_watcher.dart:427`

   Refund flows fetch current `settings.environment` and then resolve Electrum settings from that, while claim paths use `swap.environment.isTestnet`. If the user toggles environment after creating a swap, refund can choose the wrong network’s Electrum settings.

   Fix: use `swap.environment.isTestnet` for refund Electrum resolution too.

5. **Active Electrum settings are only wired into claim/refund, not swap creation**

   `lib/features/send/domain/usecases/create_send_swap_usecase.dart:67`

   Creation still passes hardcoded Electrum constants. If the PR goal is “swaps use the active Electrum URL,” this is incomplete: creation and claim/refund can use different Electrum policy.

   Fix: either wire active Electrum settings into creation too, or narrow the PR title/tests to claim/refund/broadcast only.

6. **New iOS lockfile is unrelated scope unless intentionally changing policy**

   `ios/Podfile.lock:1`

   Against `upstream/develop`, this PR adds a 297-line `ios/Podfile.lock` without `pubspec.yaml` or iOS dependency changes in the exact PR diff.

   Fix: remove it from this PR, or split the iOS lockfile policy change into its own PR.

7. **AGENTS.md naming compliance**

   `lib/core/swaps/interface_adapters/electrum_settings_adapter.dart:1`

   Under PR #2141’s AGENTS.md rules, new adapter code should use `adapters/`, not `interface_adapters/`.

   Fix: if AGENTS.md is enforced for this PR, move the new adapter to `lib/core/swaps/adapters/`.

I did not run analyzer or tests; this was a static six-agent review plus local reconciliation.

i5hi:
All issues addressed except Tor. 

Library support for tor is not ready yet, as a result this PR will not work for users who have custom electrum servers that are hidden services.

ethicnology:
rebased on develop

ethicnology:
@Claude review potential conflicts between #2139 and https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2201

#2000's design (from `electrum_servers_port.dart`, `electrum_fallback_runner.dart`, `electrum_servers_adapter.dart`) is a strict chokepoint:

- One port, `ElectrumServersPort.runWithFallback`, is the only way to use Electrum.
- The doc-comment on the port is explicit: *"The resolved server list is deliberately NOT exposed: callers cannot grab it and re-implement (or skip) the selection/fallback rule."*
- Privacy rule R2: a failing custom server is **never** replaced by a default — `fetchActiveServers` returns either all-custom or all-default, never mixed.

#2139 does exactly what #2000's port forbids:

- `ElectrumFacade.getPreferredServer` returns a single resolved server's `(url, tls, validateDomain, timeout)` so the swap path can hand it to the `boltz` SDK as raw connection args.
- Swap creation/claim/refund then use that tuple once; no fallback loop, the boltz SDK owns the socket from then on.
- It re-implements `fetchActiveServers` body inline (custom-filter + priority-sort) rather than going through any port.

Where they collide:

1. **Selection logic triplication.** Once both land you'll have the rule encoded in `DriftElectrumServerRepository.fetchActiveServers` (`drift_electrum_server_repository.dart:67`), in #2000's `ElectrumServersAdapter.runWithFallback`, and in #2139's `ElectrumFacade.getPreferredServer`. That is precisely the duplication #2000 was eliminating.
2. **Boltz is a hard bypass.** Even with #2000 merged, the boltz SDK makes its own Electrum connection internally. You can't pipe that through `runWithFallback` without either re-implementing claim/refund client-side over the chokepoint port or getting the SDK to accept a connection callback. #2139 takes neither path — it just hands boltz a URL and steps back. So swaps will still be the documented "remaining money-path/swap bypass" your memory note flags.
3. **No swap-side fallback at all.** #2139 picks server-with-priority-0 and stops. If that server is down, claim/refund fails — and there's no second attempt, even though #2000 builds the machinery for exactly that case for the rest of the app. So #2139 *partially* satisfies R2 (no custom→default leak, because no fallback) but loses R1 (try the next server on transient failure).

**My recommendation:** land #2000 first, then rebase #2139 onto it and rework swaps to go through `ElectrumServersPort` — adding a thin `resolveActiveServer({network})` (or `runWithFallback` adapted for "give me each server in turn, I'll hand it to boltz") on the same port so there's still one source of truth for selection. That also gives swaps the same fallback behaviour everything else is getting in #2000, and lets you delete `ElectrumFacade` entirely. Doing it the other way around (merging #2139 first) means #2000 lands with an exempt sibling already in place and a known bypass to clean up afterwards — which is what your memory note describes today.

If you do merge #2139 first because #2000 isn't ready, the prior-review fix #1 (delegate to `fetchActiveServers`) becomes load-bearing — it reduces the triplication to duplication and makes the eventual #2000 rebase a much smaller diff.

PR #2148, opened 2026-05-20
fix: prevent amount mismatch during send-max via swap

### LINKED ISSUE #1735: On 'v6.4.3+133' observing 'Build failed' (amount mismatch) error while trying to send MAX sats for an 'On-chain Swap' (L-BTC to BTC).
https://github.com/user-attachments/assets/c54d2cf4-8cdb-473d-bf60-96de9ce60972

i5hi:
Fixes https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1735

### DIFF
--- a/lib/features/send/presentation/bloc/send_cubit.dart
+++ b/lib/features/send/presentation/bloc/send_cubit.dart
@@ -300,8 +300,15 @@ class SendCubit extends Cubit<SendState> {
         }
       }
 
-      // Use the preselected wallet passed in the constructor if available,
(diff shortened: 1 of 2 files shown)


i5hi:
I will remove the comments and squash before merging - just have them in there to help track the lifecycle of chain swaps in the send_cubit (which is an extremely large file which covers many flows)

ethicnology (approved):
we may have conflict with https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2149

PR #2149, opened 2026-05-20
fix(send): respect manually selected wallet on amount entry

### LINKED ISSUE #1918: The cold wallet selected when sending funds does not remain selected. The Send reverts to the hot wallet without any notification.
## Description
After the user selects to send funds from their cold wallet, the UI reverts back to the 'Secure Bitcoin" wallet, without informing the user.

### Reproduce 
1. Import a cold wallet into the bull-bitcoin wallet app. (in my case,Jade HWW ver1.0.38)
2. In the Bull bitcoin wallet app home screen - click on Send 
3. Paste your on-chain destination address and then press continue.
4. Select your cold wallet.
5. Click on the amount entry (the currently zero Sats box), type ay amount in sats. eg 2000
6. Now look at the wallet name below and the wallet dropdown above: They have both reverted back to 'secure bitcoin' , not the cold wallet that the user previously selected. 
7. If the user does not notice this change the funds will be signed and sent from the secure wallet accidently.  

### Expected behavior
The cold wallet should remain selected, since it was selected earlier in the flow. (this Issue is also mentioned in #PR1917). The wallet pre-selected by the user should not be changed inadvertently by the app.

Additionally, the continue button on the 'secure bitcoin' (aka the hot wallet) sending, should read Confirm send (or similar) , instead of Continue as it is the final screen.  
If spending from the cold wallet then 'Continue' is the correct button label, as it not the final step.

## Context
- `BULL`: release `6.5.2` 
- `OS`: Android `16.1.163`
- `Device`: Google Pixel 9
- `Extra`: `GrapheneOS` with Google Services

### Platforms
- [x] Android
- [ ] iOS

ethicnology:


i5hi (approved):

i5hi:
Added some logic to ensure that swaps via send cannot be done via hardware wallets

i5hi:
Still needs a little more extensive testing

PR #2154, opened 2026-05-21
build tools container before devcontainer builds

basantagoswami:
A small 2 line change

Added `make container-tools` to `initializeCommand` in `devcontainer.json` otherwise you manually have to run this command before DC is setup else the CLI or VSCode extension fails. This now works with VSCode and the CLI. Zed still fails probably because they don't follow the spec completely

Also removed `--delete-conflicting-outputs` from `make build-runner` because it's been removed in [v2.7.0](https://pub.dev/packages/build_runner/changelog#270)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,7 +9,7 @@
     "remoteUser": "${localEnv:USER}",
     "updateRemoteUserUID": true,
     "shutdownAction": "stopContainer",
-    "initializeCommand": ".devcontainer/init-ssh-agent.sh",
+    "initializeCommand": ".devcontainer/init-ssh-agent.sh && make container-tools",
     "runArgs": [
         "--name",
         "${localWorkspaceFolderBasename}",
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,31 +1,31 @@
 {
-    "name": "${localWorkspaceFolderBasename}",
-    "build": {
-        "dockerfile": "../Containerfile.app",
-        "args": {
-            "USERNAME": "${localEnv:USER}"
-        }
+  "name": "${localWorkspaceFolderBasename}",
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,7 @@
     "remoteUser": "${localEnv:USER}",
     "updateRemoteUserUID": true,
     "shutdownAction": "stopContainer",
-    "initializeCommand": ".devcontainer/init-ssh-agent.sh",
+    "initializeCommand": ".devcontainer/init-ssh-agent.sh && make container-tools",
```
ethicnology: since the build section use the `Containerfile.app` which depends on `Containerfile.tools` it should already be built.

ethicnology (changes requested):
initializeCommand shouldn't make the container-tools since it's already built by the devcontainer

.devcontainer/devcontainer.json
basantagoswami: on a fresh machine, it failed to build when i tried with zed, vscode, and the devcontainer cli. i had to build the tools image. else all three of these environments tried to fetch it from the registry and failed to do so

basantagoswami: can you try to reproduce it by running `podman system reset` or something equivalent and then try opening the repo in vscode and select "reopen in devcontainer". it fails on my system

maybe a multi stage Containerfile can also work here, instead of what i am doing

ethicnology: Maybe we can put Container.tools instead of .app for the devcontainer?

ethicnology: I've made the replacement in 34f5bb8ada858baa800285200eecb983ad489e9f

ethicnology (approved):
I've edited in 34f5bb8ada858baa800285200eecb983ad489e9f

PR #2163, opened 2026-05-22
fix: remove manual retry swap claim button

### LINKED ISSUE #1912: Retry Swap Claim doesn't work for me
## Description
I tried to swap some sats and still not claimed. I did it on 23/1 and almost every day I tried Retry Swap Claim, but did not work. 

I use current Bull Wallet on current iOS on Apple iPhone Air. 

Image

Thanks for help. 

Jikoun

### LINKED ISSUE #2134: Remove "Retry Swap Claim" button and automate swap claim retries
The "Retry Swap Claim" button should be removed from the UI. Swap claim retries should be handled automatically by the app without requiring manual user intervention.

## User stories

- As a user, I should not have to manually retry claiming a swap — the app should handle retries automatically in the background
- As a user, I should not see a confusing "Retry Swap Claim" button that may not work reliably

## Context

Users have reported that the "Retry Swap Claim" button does not work reliably, leading to stuck swaps with no recourse. Rather than fixing the button, the claim retry logic should be automated so users never need to intervene manually.

- Related: #1912 (user report of Retry Swap Claim not working)

## Acceptance criteria

- [ ] Remove the "Retry Swap Claim" button from the transaction detail UI
- [ ] Implement automatic swap claim retries in the background (e.g. on app launch, on sync, on resume)
- [ ] Ensure pending/claimable swaps are retried without user action
- [ ] If a swap remains unclaimable after retries, surface a clear status to the user (not a manual retry button)

## Test cases

- [ ] Swap in claimable state is automatically retried on app launch
- [ ] Swap in claimable state is automatically retried on wallet sync
- [ ] No "Retry Swap Claim" button visible in the UI
- [ ] Swap that succeeds on automatic retry updates status correctly
- [ ] Swap that fails all automatic retries shows appropriate status message

wired-pasteque:
Fix: #2134, #1912

Simulator Screenshot - iPhone 16e - 2026-05-22 at 10 48 07


BullishNode:
Why are we removing the retry claim button? It simply wasn't working? @i5hi

i5hi:
Its running the same logic as in the swap watcher. We had it there earlier because a mismatch of the status sometimes never triggered the right method in the watcher, so we let you manually trigger it - those statuses are all now solid and covered.

A pull to sync will trigger the right claim/refund method. Most of the time now it just causes more frustration because it seems like something that can help you but it never does because the problem is not the swap watcher not catching it, its in the execution of the claim/refund logic. 

You could argue there is no harm in having it there, just in case ...

i5hi (dismissed):

lib/features/transactions/ui/screens/transaction_details_screen.dart
```diff
@@ -173,19 +174,20 @@ class TransactionDetailsScreen extends StatelessWidget {
...
-                        : context.loc.transactionDetailRetrySwap(swapAction),
-                    onPressed: () async {
-                      await context.read<TransactionDetailsCubit>().processSwap(
-                        swap,
-                      );
-                    },
-                    bgColor: context.appColors.primary,
-                    textColor: context.appColors.onPrimary,
-                  ),
+                  if (isChainSwap)
```
i5hi: Looks like we are still showing it

i5hi (changes requested):

i5hi:
@BullishNode Its also safer not to allow it to prevent race conditions if i swap is being processed and a user tries to trigger it while one is already in progress. In most cases, it should not affect anything and just fail if the existing attempt worked.
If we really want to keep it, we need to introduce a storage flag to track if a swap is being currently processed and only if it is not then allow a manual intervention.

wired-pasteque:
@i5hi, I’ve removed the retry swap action button

ethicnology (approved):

ethicnology:
@i5hi need your approval & merge here

i5hi (approved):

PR #2164, opened 2026-05-22
Fix/zh localization wallet translations

lwhile:
- Kept hardware wallet brand names in their original English form.
- Batch translated English with the same content.


BullishNode:
Appreciate it. We will review soon!

wired-pasteque (approved):
Reviewed, translations look correct and consistent

PR #2166, opened 2026-05-22
refactor(devcontainer): mount ssh socket under $HOME so it's visible in podman VM

ethicnology:
The ssh-agent socket source was hard-coded to /tmp/ssh-agent.sock. On the podman machine VM, /tmp is its own tmpfs (only /Users and /var/folders are virtiofs-mounted from the host), so the bind mount failed with "statfs /tmp/ssh-agent.sock: no such file or directory". Move the path to $HOME/.ssh-agent-devcontainer.sock so it lives under /Users and is visible inside the VM on macOS. Linux is unaffected (native podman + real /tmp).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 {
   "name": "${localWorkspaceFolderBasename}",
   "build": {
-    "dockerfile": "../Containerfile.tools",
+    "dockerfile": "../Containerfile.app",
     "args": {
       "USERNAME": "${localEnv:USER}",
     },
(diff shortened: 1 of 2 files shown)


ethicnology:
@basantagoswami you can try this

use `make devcontainer`

.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "${localWorkspaceFolderBasename}",
   "build": {
-    "dockerfile": "../Containerfile.tools",
+    "dockerfile": "../Containerfile.app",
```
basantagoswami, line 4: Let's keep it Containerfile.tools itself. The tools container is enough to be used as the project devcontainer, while the app container works for other stuff like reproducibility testing and building. Changing it back to `Containerfile.app` will cause issues with the devcontainer CLI again

PR #2168, opened 2026-05-24
feat(linux): BULL now runs on desktop

ethicnology:
First desktop port of the app.

Builds and runs end-to-end on Linux x86_64. Everything works except background tasks, `workmanager` only supports Android and iOS, so we skip its init on desktop.

macOS and Windows are next, and most of the heavy lifting in this PR carries over: the `bull_sdk` FFI is now proven on a desktop target, the desktop platform guards (workmanager, future plugins) are in place, and `BBRefreshIndicator` already handles pointer-based drag. The remaining work per platform is mostly scaffolding (`macos/`, `windows/`) and a couple of plugin bumps.

**Big benefit:** integration tests can now run on a GitHub Actions Linux runner, no Android emulator in CI.

**Caveat:** the UI is built for mobile and looks rough on desktop. Still usable enough for debugging though, which means contributors without high-end machines no longer need to fire up an Android emulator just to poke at the app, RAM is expensive these days.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ docs/zola/public/
 
 # # Untrack unused platform files
 macos/
-linux/
 windows/
 
 # Files and directories created by pub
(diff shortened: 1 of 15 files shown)


ethicnology:
No regressions on mobile expected

PR #2169, opened 2026-05-24
update devcontainer readme to fix adb-over-host setup for usb debugging

basantagoswami:


basantagoswami:
Maybe we should just update the code to handle this automatically, instead of just updating the README. I don't know if we automatically set `ADB_SERVER_SOCKET`, but user is not running the server in the host, that'll throw an error or if we can make the container adb setup try to connect via `ADB_SERVER_SOCKET`, when failed spins a server locally (to work using an emulator inside the container itself for example)

ethicnology:
@basantagoswami on develop is ADB now working for you

basantagoswami:
@ethicnology ADB works with the updated setup, closing this PR

PR #2172, opened 2026-05-25
dedup no_screenshot kotlin source dirs for K2

basantagoswami:
on a fresh setup/build ran into this error:

```
e: file:///home/human/.pub-cache/hosted/pub.dev/no_screenshot-1.1.0/android/src/main/kotlin/com/flutterplaza/no_screenshot/NoScree
  nshotPlugin.kt:39:11 Conflicting declarations:
  val SCREENSHOT_ON_CONST: String
  e: .../NoScreenshotPlugin.kt:42:11 Conflicting declarations:
  val PREF_NAME: String
  ...
  e: .../NoScreenshotPlugin.kt:72:38 Overload resolution ambiguity between candidates:
  val PREF_NAME: String
  val PREF_NAME: String
  e: .../NoScreenshotPlugin.kt:97:65 Overload resolution ambiguity between candidates:
  val SCREENSHOT_METHOD_CHANNEL: String
  val SCREENSHOT_METHOD_CHANNEL: String
  ...

  FAILURE: Build failed with an exception.

  * What went wrong:
  Execution failed for task ':no_screenshot:compileDebugKotlin'.
  > A failure occurred while executing
  org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
     > Compilation error. See log for more details

  BUILD FAILED in 41s
  Error: Gradle task assembleDebug failed with exit code 1
```

the full log repeats this for every top level constant

we probably need to consider eventually mvoing to a more maintained screenshot-prevention plugin

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -43,6 +43,21 @@ subprojects {
     }
 }
 
+// Workaround: no_screenshot 1.1.0 lists `src/main/kotlin` twice, which the K2
+// frontend (Kotlin 2.2.10) rejects as "Conflicting declarations". Dedup its
+// Kotlin source dirs by canonical path.
+// TODO: consider moving to a more maintained screenshot-prevention plugin.
+subprojects {
+    afterEvaluate { project ->
+        if (project.name == 'no_screenshot') {
+            def kotlinExt = project.extensions.findByName('kotlin')
+            kotlinExt?.sourceSets?.configureEach { ss ->
+                ss.kotlin.setSrcDirs(ss.kotlin.srcDirs.collect { it.canonicalFile }.unique())
+            }
+        }
+    }
+}
+
 tasks.register("clean", Delete) {
     delete rootProject.buildDir
 }


ethicnology:
At this point I would even suggest that we remove no_screenshot from the codebase

What you think @i5hi @thibistaken

ethicnology:
BTW seems maintained latest 2 months ago, PR and Issues processed

https://github.com/FlutterPlaza/no_screenshot

basantagoswami:
> BTW seems maintained

yup that fucker claude lied to me

i agree we should drop no_screenshot from the repo. let me check if it's relatively easy to viveslop something that only does screenshot blocking for android and ios based on that package

basantagoswami:
vivesloping done. if this works then it makes sense for us to maintain it instead of relying on another dependency. i'll test this patch later

```
diff --git a/android/app/src/main/kotlin/com/bullbitcoin/bb_mobile/MainActivity.kt b/android/app/src/main/kotlin/com/bullbitcoin/bb_mobile/MainActivity.kt
index 65d0e5ad3..5531523a5 100644
--- a/android/app/src/main/kotlin/com/bullbitcoin/bb_mobile/MainActivity.kt
+++ b/android/app/src/main/kotlin/com/bullbitcoin/bb_mobile/MainActivity.kt
@@ -1,5 +1,31 @@
 package com.bullbitcoin.mobile
 
+import android.view.WindowManager
 import io.flutter.embedding.android.FlutterActivity
+import io.flutter.embedding.engine.FlutterEngine
+import io.flutter.plugin.common.MethodChannel
 
-class MainActivity: FlutterActivity(){}
+class MainActivity : FlutterActivity() {
+    private val privacyScreenChannel = "com.bullbitcoin.mobile/privacy_screen"
+
+    override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
+        super.configureFlutterEngine(flutterEngine)
+
+        MethodChannel(flutterEngine.dartExecutor.binaryMessenger, privacyScreenChannel)
+            .setMethodCallHandler { call, result ->
+                when (call.method) {
+                    // FLAG_SECURE blocks screenshots and screen recording and
+                    // blanks the app-switcher thumbnail.
+                    "enable" -> {
+                        window.addFlags(WindowManager.LayoutParams.FLAG_SECURE)
+                        result.success(null)
+                    }
+                    "disable" -> {
+                        window.clearFlags(WindowManager.LayoutParams.FLAG_SECURE)
+                        result.success(null)
+                    }
+                    else -> result.notImplemented()
+                }
+            }
+    }
+}
diff --git a/ios/Runner/AppDelegate.swift b/ios/Runner/AppDelegate.swift
index bc109bc67..3890e19dc 100644
--- a/ios/Runner/AppDelegate.swift
+++ b/ios/Runner/AppDelegate.swift
@@ -4,11 +4,32 @@ import workmanager_apple
 
 @main
 @objc class AppDelegate: FlutterAppDelegate {
+  private let privacyScreen = PrivacyScreen()
+
   override func application(
     _ application: UIApplication,
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
   ) -> Bool {
     GeneratedPluginRegistrant.register(with: self)
+
+    if let controller = window?.rootViewController as? FlutterViewController {
+      let channel = FlutterMethodChannel(
+        name: "com.bullbitcoin.mobile/privacy_screen",
+        binaryMessenger: controller.binaryMessenger
+      )
+      channel.setMethodCallHandler { [weak self] call, result in
+        switch call.method {
+        case "enable":
+          self?.privacyScreen.enable()
+          result(nil)
+        case "disable":
+          self?.privacyScreen.disable()
+          result(nil)
+        default:
+          result(FlutterMethodNotImplemented)
+        }
+      }
+    }
     WorkmanagerPlugin.registerPeriodicTask(
       withIdentifier: "com.bullbitcoin.mobile.bitcoin-sync-id",
       frequency: NSNumber(value: 20 * 60)
@@ -29,3 +50,76 @@ import workmanager_apple
     return super.application(application, didFinishLaunchingWithOptions: launchOptions)
   }
 }
+
+// iOS has no public API to block screenshots. The supported trick is to host
+// the app's content inside a UITextField whose `isSecureTextEntry` layer is
+// excluded from screenshots and screen recordings, so the captured frame comes
+// out blank. We reparent the key window's layer under the secure field's layer
+// to extend that protection to the whole screen.
+final class PrivacyScreen {
+  private var screenPrevent = UITextField()
+  private weak var attachedWindow: UIWindow?
+
+  func enable() {
+    attachWindowIfNeeded()
+    screenPrevent.isSecureTextEntry = true
+  }
+
+  func disable() {
+    screenPrevent.isSecureTextEntry = false
+  }
+
+  private func attachWindowIfNeeded() {
+    guard let window = activeKeyWindow() else { return }
+    if window === attachedWindow { return }
+
+    // Undo previous reparenting before attaching to a different window.
+    if let oldWindow = attachedWindow,
+       let rootLayer = screenPrevent.layer.superlayer {
+      rootLayer.addSublayer(oldWindow.layer)
+      screenPrevent.layer.removeFromSuperlayer()
+    }
+
+    // Fresh field avoids stale layer state.
+    screenPrevent = UITextField()
+    configure(window: window)
+    attachedWindow = window
+  }
+
+  private func configure(window: UIWindow) {
+    guard let rootLayer = window.layer.superlayer else { return }
+    guard screenPrevent.layer.superlayer == nil else { return }
+
+    screenPrevent.semanticContentAttribute = .forceLeftToRight  // RTL fix
+    screenPrevent.textAlignment = .left                         // RTL fix
+
+    // Briefly add to the window so UIKit builds the text field's internal
+    // sublayer hierarchy, then force a layout pass and remove it so it is not
+    // an actual subview. This avoids a circular view hierarchy that causes
+    // EXC_BAD_ACCESS on iOS 26+.
+    window.addSubview(screenPrevent)
+    screenPrevent.layoutIfNeeded()
+    screenPrevent.removeFromSuperview()
+
+    // Keep the layer at the origin so reparenting does not shift content.
+    screenPrevent.layer.frame = .zero
+
+    rootLayer.addSublayer(screenPrevent.layer)
+    if #available(iOS 17.0, *) {
+      screenPrevent.layer.sublayers?.last?.addSublayer(window.layer)
+    } else {
+      screenPrevent.layer.sublayers?.first?.addSublayer(window.layer)
+    }
+  }
+
+  private func activeKeyWindow() -> UIWindow? {
+    guard let scene = UIApplication.shared.connectedScenes
+      .first(where: { $0.activationState == .foregroundActive }) as? UIWindowScene
+    else { return nil }
+
+    if #available(iOS 15.0, *) {
+      return scene.keyWindow
+    }
+    return scene.windows.first(where: { $0.isKeyWindow })
+  }
+}
diff --git a/lib/core/mixins/privacy_screen.dart b/lib/core/mixins/privacy_screen.dart
index 468bb551c..248661429 100644
--- a/lib/core/mixins/privacy_screen.dart
+++ b/lib/core/mixins/privacy_screen.dart
@@ -1,9 +1,13 @@
-import 'package:no_screenshot/no_screenshot.dart';
+import 'package:flutter/services.dart';
 
 mixin PrivacyScreen {
+  static const MethodChannel _channel = MethodChannel(
+    'com.bullbitcoin.mobile/privacy_screen',
+  );
+
   Future<void> enableScreenPrivacy() async =>
-      await NoScreenshot.instance.screenshotOff();
+      _channel.invokeMethod<void>('enable');
 
   Future<void> disableScreenPrivacy() async =>
-      await NoScreenshot.instance.screenshotOn();
+      _channel.invokeMethod<void>('disable');
 }
diff --git a/pubspec.lock b/pubspec.lock
index 4336a83d1..21d99f5f2 100644
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1357,14 +1357,6 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "1.0.0"
-  no_screenshot:
-    dependency: "direct main"
-    description:
-      name: no_screenshot
-      sha256: "373cc7d9ec62f2fe0898318bfa38ad8f717e5b9729be78cd00112d3ab278cb18"
-      url: "https://pub.dev"
-    source: hosted
-    version: "1.1.0"
   node_preamble:
     dependency: transitive
     description:
diff --git a/pubspec.yaml b/pubspec.yaml
index 6082d3af1..1e923e4b1 100644
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -59,7 +59,6 @@ dependencies:
       path: packages/bitbox-transport
   hex: ^0.2.0
   auto_size_text: ^3.0.0
-  no_screenshot: ^1.1.0
   freezed_annotation: ^3.1.0
   convert: ^3.1.2
   crypto: ^3.0.7
```

ethicnology:
@basantagoswami do you still run in that error ?

basantagoswami:
> @basantagoswami do you still run in that error ?

I have setup a fresh container a few times, but only ran the Linux version of the app, will need to check by running the Android app once

basantagoswami:
@ethicnology I no longer run into this error, probably resolved in one of your recent merged PR with version bump or something. Did you explicitly do something to fix this? Either way, we can close this PR I think

Issue #1992, opened 2026-03-30
Electrum health check misleading — ping succeeds but operations fail


thibistaken:
Settings shows electrum servers as "up" (ping-only check) while homepage shows server failure. Broadcast, sync, and fees (via mempool) all non-functional for both Bitcoin and Liquid. Causes stuck swaps.

**Fix needed:**

ethicnology:
@thibistaken even if #1999 make it to this sprint there is no work for #2000 yet

thibistaken:
## Test Case

1. Open app and check server status in Settings
2. Compare with homepage server indicator
3. **Verify**: both show consistent status
4. If server is degraded (ping works but operations fail), verify status shows error/offline
5. **Verify**: broadcast, sync, and fee estimation all reflect correct server health

thibistaken:
Likely to move to next release because #2000 has not been worked on yet.

PR #2201, opened 2026-05-27
refactor(electrum): route all server use through one fallback chokepoint

### LINKED ISSUE #2000: Fallback to alternative electrum servers when primary fails functionally

ethicnology:
# Route all Electrum interactions through one fallback chokepoint

Closes #2000 (sub-issue of #1992).

Electrum server selection + fallback was hand-rolled in ~9 separate loops across broadcast, sync, tx-fetch, dry-scan and payjoin, and the custom-vs-default rule was duplicated in 3 places. This PR centralises both in one auditable, unbypassable seam.

### The rule (now enforced in one place, by construction)
- **R1** — no custom server → use defaults, fall back across defaults.
- **R2** — custom server(s) set → use only custom, never defaults.
- **R2a** (privacy) — custom fails → do **not** fall back to a default; fail loudly. Routing a privacy user's traffic to Bull's servers is exactly what they opted out of. Canonical Electrum-wallet privacy stance.

### What changed

`ElectrumServersPort.runWithFallback` is the **only** public surface for Electrum traffic out of the app. Its operation callback receives a fully-resolved `ElectrumConnection` (url + retry + timeout + stopGap + validateDomain + isCustom + socks5), produced by the adapter from three sources merged exactly once per call: active servers, persisted electrum settings, and the current Tor preference. Tor proxy applies to Bitcoin only, never Liquid.

`runElectrumFallback` itself is now internal to the adapter — no consumer outside `core/electrum` imports it. Failures emerge as one of two typed exceptions: `NoElectrumServersConfiguredException` (set was empty, nothing tried) or `AllElectrumServersFailedException` (every transient attempt failed, with the per-server `attempts` list intact). Programming bugs in the operation — `TypeError`, `StateError`, failed assertions — now propagate immediately instead of being retried against every server and masked as "all servers failed": the default `isTransient` is `e is Exception`, not bare `catch`. An empty list throws `ArgumentError` in release builds instead of silently falling through `assert`.

The two legacy `ElectrumServerPort.getElectrumServers` ports (`core/blockchain/`, `core/wallet/`) and their duplicate adapters are deleted. Payjoin's hand-rolled `_getSortedElectrumServersForNetwork` selector is deleted. Broadcast use cases no longer fetch lists — `BitcoinBlockchainRepository` / `LiquidBlockchainRepositoryImpl` take the port and the network. `_syncWallet`, `_getPublicWalletModels`, `tryBroadcastOriginalTransaction`, and `_broadcastPsbt` all route through `runWithFallback`. BDK/LWK datasources accept `ElectrumConnection` at the boundary, collapsing the two duplicate `ElectrumServer` data classes that were declared inside the legacy ports.

`_syncWallet` and `CheckWalletStatusUsecase` (renamed from `TheDirtyUsecase` per AGENTS.md naming) now rethrow the typed `ElectrumFallbackException` instead of flattening it into a string — the per-server `attempts` survive upstream for diagnostics. Empty-servers behaviour between the two sync paths is now uniform (both throw via the port — the previous silent-skip in `WalletTransactionRepositoryImpl` is gone).

**Net diff vs `develop`: +504 / −577 → −73 LOC** for substantially more behaviour: the chokepoint is enforced by construction, no module can fetch a server list and skip the rule.

### Tests

`test/core_test/electrum/` — 17 tests, all green.
- Runner: first-success, transient advance, permanent rethrow, all-failed exception shape with attempts, **default `isTransient` does NOT catch `Error`**, **`ArgumentError` on empty list**.
- Adapter: full fallback semantics + **electrum-settings merge into every connection**, **Tor disabled → persisted socks5 (which may be null)**, **Tor enabled on Bitcoin → `127.0.0.1:<port>`**, **persisted socks5 takes precedence over Tor default**, **Tor disabled on Liquid even when enabled globally**, **R2a — custom-only set + all fail never reaches defaults** (`verifyNoMoreInteractions(repo)`).

### Deliberately NOT in this PR — scope, for reviewers

- **Swaps still bypass the rule (biggest remaining gap).** Five swap use cases hardcode `bbElectrumUrl` / `bbLiquidElectrumUrlPath` to Boltz, ignoring a configured custom server (violates R2). Boltz takes one URL and owns its connection, so it's a *selection* fix (no fallback possible there, and per R2a no silent reachability-fallback to a default). Cross-module + needs a real testnet swap to verify. PR #2139 (`Swap electrum url`) addresses this in parallel — it introduces an `ElectrumFacade.getPreferredServer()` that threads a single URL into Boltz claim/refund/broadcast. The two PRs merge mechanically (only `core_locator.dart` and `electrum_locator.dart` overlap, additive). Flagged for #2139's review that its facade re-implements R1/R2 by hand and should ideally consume `ElectrumServersPort` once the Boltz SDK supports a per-attempt callback. cc @i5hi
- **Broadcast idempotency**: a verified `txn-already-in-mempool` reply should count as success by passing an `isTransient` classifier to `runWithFallback` in `bitcoin_blockchain_repository.broadcastPsbt`. Deferred until the exact BDK/LWK error surface is confirmed — guessing it risks masking a real broadcast failure.


ethicnology:
Hey @Claude what is your review on this PR

i5hi (approved):

PR #2203, opened 2026-05-27
fix: hinglish now resolves to hi_Latn instead of hindi

basantagoswami:


basantagoswami:
Saw this bug where if you choose Hinglish as the language in app settings, it showed Hindi instead

lib/core/settings/domain/settings_entity.dart
```diff
@@ -42,35 +42,41 @@ enum BitcoinUnit {
...
+  italian('it', label: 'Italiano'),
+  korean('ko', label: '한국어'),
+  portuguese('pt', label: 'Português'),
+  brazilianPortuguese('pt', label: 'Português (Brasil)', countryCode: 'BR'),
+  russian('ru', label: 'Русский'),
+  thai('th', label: 'ภาษาไทย'),
+  turkish('tr', label: 'Türkçe'),
+  ukrainian('uk', label: 'Українська'),
+  simplifiedChinese('zh', label: '简体中文'),
+  assamese('as', label: 'অসমীয়া');
```
ethicnology, line 67: What is this new label ?

ethicnology (commented):
What is this new label:

lib/core/settings/domain/settings_entity.dart
basantagoswami: label is the existing third positional arg (English, etc) to Language, just moved to a named parameter because countryCode and scriptCode are now optional

ethicnology: but label was nullable and now it's a hardcoded `label:` this may have an impact on the UI displaying empty labels for most languages?

basantagoswami: label was not nullable, it was and is `final String label`,  nulls you see disappearing from the old diff lines are the countryCode, not the label. ui won't change

basantagoswami: i can add back the countryCode actually, it was an unnecessary change

basantagoswami: added them back. diff is cleaner now

ethicnology (approved):

PR #2210, opened 2026-05-28
feat: export txn history as csv

### LINKED ISSUE #1363: CSV Export for All Transactions
Let's add this feature where a user can export all his wallet transactions, including btc, lbtc, lightning, and swaps. See: https://x.com/ricd40/status/1979857122621104335

basantagoswami:
Fixes #1363


ethicnology:
Hey @Claude can you review this PR according to our AGENTS.md and ARCHITECTURE.md rules and the expected feature detailed in #1363

ethicnology:
@basantagoswami any answers regarding Claude review ?

BullishNode:
@ethicnology comparing Claude with my codex bb-review skill

  1. The CSV exports sensitive Lightning fields without explicit disclosure or
  approval. `invoice` and `preimage` are included in the export, including
  `LnSendSwap.preimage`, while the UI describes this as a generic transaction
  export. Please remove those fields by default or add an explicit opt-in/
  disclosure with tests.

  2. Chain-swap rows can report the wrong `network`. The generic `network`
  column always uses the swap send network, even when the row is the receive-
  side wallet transaction. Prefer `walletTransaction.network` when present, and
  keep `send_network` / `receive_network` as swap-leg metadata.

  3. Date filtering has bad edge cases: reversed ranges produce a misleading “no
  transactions” result, and the end date uses `23:59:59`, excluding sub-second
  transactions in the final second. Validate picker ranges and use a half-open
  end bound: `< endDate + 1 day`.

  4. `backup_settings` imports `transactions/ui/transactions_router.dart`,
  bypassing the facade-only feature boundary, and `FEATURES.md` is not updated.
  Either remove the backup-settings entry and keep export discoverability in
  Transaction History, or expose a public transactions navigation contract and
  update the graph.

  5. `ExportTransactionsCubit` owns file saving (`file_picker`), filename
  construction, byte encoding, and raw `e.toString()` error display. Move save
  orchestration behind an application port/use case and map failures to
  localized presentation states.

  6. The export screen uses a fixed `Column` with long localized copy, date
  fields, `Spacer`, and a bottom CTA. This can overflow on small/large-text
  layouts. Make the body scrollable while keeping the CTA reachable.

  7. The formatter port/DI layer and single-file `application/usecases/` folder
  are premature for one CSV-only export. Simplify the formatter path and place
  the single use case directly under `application/` per AGENTS.md.

ethicnology:
debug apk available here https://github.com/SatoshiPortal/bullbitcoin-mobile/actions/runs/26586256138

basantagoswami:
@claude

- utc bug: tx.timestamp is already local time, because it's built via DateTime.fromMillisecondsSinceEpoch without isUtc: true
- swap status precedence: fix, swap status is checked before wallet tx status now
- lightning network column: lightning now has 'lightning' for network and nothing for address
- NoTransactionsToExportError: moved to application/transactions_application_error.dart along with InvalidDateRangeError
- cubit e.toString(): now emits error with no message, ui shows localized error message
- filename issue in cubit: TransactionExportSaver port + CsvTransactionExportSaver adapter, and filename now uses _ as separator
- single file in application/usecases/ breaking the arch doc rule: fixed by moving GetTransactionsUsecase and GetTransactionsByTxIdUsecase also
- formatter in registerUsecases() moved to registerAdapters()
- screen overflow: fixed using scrollable body + pinned CTA
- date locale: fixed using MaterialLocalizations.formatShortDate
- sort order: fixed to be newest first
- date range edge cases: InvalidDateRangeError on reversed range, half-open end bound (< endDate + 1 day)
- test gaps: added new tests, updated to reflect code changes
- bolt11 invoice: removed column completely, only keeping preimage

review again

basantagoswami:
- error file renamed to application_errors.dart
- chain-swap network column now shows receive network
- facade rejecte, every other cross-feature consumer of TransactionsRoute (wallet, buy, sell, send, pay, receive, withdraw) imports transactions_router.dart directly; the facade pattern is labels-only and not replicated elsewhere, so skipping it for now

ethicnology:
rebased on develop

PR #2218, opened 2026-05-29
fix(send): support LNURL pay limits

### LINKED ISSUE #1415: LNURL with and without amount
## Description
I tested several types of static QR codes for LNURL with different rules:
- 1000 sats minimum (no maximum)
- 5000 sats minimum (no maximum)
- No minimum (custom amount)
- 5000 sats fixed amount

For some reason, **BULL Wallet** has trouble reading these codes:

⚠️ **1000 sats minimum**: Can read it, but doesn't enforce the minimum — allows customization from 1 sat  
❌ **5000 sats minimum**: Cannot read it at all  
✅ **No minimum**: Can read it  
❌ **5000 sats fixed amount**: Cannot read it

After discussing with @ethicnology, his initial hypothesis was:  
> "Yeah, the wallet is expecting the LNURL payload directly (not embedded in `LIGHTNING:`)"

However, **all** these QR codes embed the LNURL payload in `LIGHTNING:`, and some still work.

### Reproduce
Scan the following QR codes and observe the behavior:  
QR code examples

### Expected Behavior
- **1000 sats minimum QR code**: The wallet should allow customization only for amounts ≥ 1000 sats  
- **5000 sats minimum QR code**: The wallet should allow customization only for amounts ≥ 5000 sats  
- **No minimum QR code**: The wallet should allow sending any amount  
- **5000 sats fixed amount QR code**: The wallet should **not** allow customization: user should only verify and pay

### Platforms
- [x] Android
- [x] iOS

BullishNode:
## Summary
- resolve LNURL-pay metadata before the send amount step and apply min/max/fixed amount limits
- remove the 1000 sat dummy invoice probe for LNURL-pay validation
- keep fixed LNURL amounts read-only, clear stale amount state on new recipients, and reselect the best wallet with the fixed amount
- make LIGHTNING:/mixed-case lightning: LNURL parsing strip the scheme consistently
- use the narrow `bech32` package directly for LNURL decoding instead of making `blockchain_utils` a direct dependency

## Alternatives considered
- Keep the old dummy invoice probe: smallest change, but it is the root cause of the bug because a 1000 sat probe rejects valid LNURL-pay requests whose `minSendable` or fixed amount is higher.
- Use `satoshifier`: not sufficient here because the send flow needs LNURL-pay metadata (`minSendable`, `maxSendable`, fixed amount), not just a normalized payment string.
- Add `blockchain_utils`: worked technically for Bech32 decoding, but was too broad for this PR. The final version uses the narrower `bech32` package already present in the lockfile.
- Only support Lightning addresses: simpler, but would not fix scanned `LIGHTNING:LNURL...` QRs, which are part of the reported failure mode.
- Reuse swap limit types: would reduce one small domain type, but muddles LNURL-pay metadata with swap limits and required generated-state changes while build runner is currently failing under this toolchain.
- Put all LNURL handling in `SendCubit`: fewer files, but bloats the Cubit with protocol decoding and metadata validation. The PR keeps decoding/fetching/parsing in a small usecase and leaves the Cubit to coordinate send state.
- Add a larger LNURL service layer: cleaner if LNURL support expands later, but overbuilt for this bug fix.

## Verification
- .fvm/flutter_sdk/bin/flutter test test/features/send/lnurl_pay_limits_test.dart test/features/send/presentation/send_state_lnurl_limits_test.dart --reporter=compact
- .fvm/flutter_sdk/bin/flutter analyze
- .fvm/flutter_sdk/bin/dart fix --dry-run
- git diff --check HEAD
- .fvm/flutter_sdk/bin/flutter test test/ --reporter=compact

Fixes #1415


BullishNode:
@claude review this PR

BullishNode:
Claude review comments addressed

BullishNode:
Not tested yet

PR #2223, opened 2026-05-31
feat: add Trezor hardware wallet support

### LINKED ISSUE #2294: Review/Merge hardware wallet PRs
https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2224
https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2223

anipy1:
This PR adds Trezor as a hardware wallet option in the Import Wallet screen. Users can import a Trezor wallet, verify receive addresses on the device, and sign + broadcast transactions, all routed through Trezor Suite via [Trezor Connect](https://github.com/cake-tech/trezor_connect).

Trezor connect is like a [deeplink](https://connect.trezor.io/9/details/deeplinking/) protocol with no USB or BLE stack needed in the app. Bull opens an HTTPS URL that Trezor Suite intercepts. Suite handles the protocol with the device. When done, Suite redirects back to Bull via a custom URL scheme (`bullbitcoin://trezor-callback`). An `app_links` listener catches the callback and resolves the Future the app is awaiting.

The new pieces:

- New feature slice at `lib/features/trezor/`
- `bullbitcoin` URL scheme registered on [iOS](https://github.com/anipy1/bullbitcoin-mobile/blob/trezor-connect/ios/Runner/Info.plist#L35) + [Android](https://github.com/anipy1/bullbitcoin-mobile/blob/trezor-connect/android/app/src/main/AndroidManifest.xml#L84).
- [Deeplink listener](https://github.com/anipy1/bullbitcoin-mobile/blob/trezor-connect/lib/main.dart#L290) wrapping `MaterialApp.router` in `main.dart`.
- GoRouter `onEnter` guard that blocks the callback URI from replacing the nav stack.
- `SignerDeviceEntity` enum extended with `trezor`.
- Send screen + Receive screen extended with the Trezor-specific buttons (same shape Ledger and BitBox use).

I used [Cake Wallet](https://github.com/cake-tech/cake_wallet) as my reference implementation. The Dart package this depends on was written by the Cake team here: https://github.com/cake-tech/trezor_connect — it wraps the Trezor Connect deeplink protocol documented at https://connect.trezor.io/9/details/deeplinking/.

I tested it with old Trezor Model One but the deeplink protocol is the same across all Trezor models (One, T, Safe 3, Safe 5), so the newer ones should work without code changes, but I haven't been able to test them.

A place where Cake offers more: After importing, Cake shows a list of accounts and lets the user pick (screenshot below). I kept the flow simple by always importing only account 0. Most users only use account 0 anyway, and the use case already supports paging (`startIndex` + `count`) so we can add a picker as a follow-up if you want it.

<details>
<summary>Cake's account picker for reference</summary>

image

</details>

**No localization yet.** All new Trezor strings are hardcoded English for now. I'll sweep them into the ARB files afterwards.


BullishNode (changes requested):
This was created by an automated review skill and should not be considered a real review by repository maintainers.

Requesting changes based on an automated review pass. The main blockers are around address verification, fingerprint ownership, signing metadata, callback failure handling, device compatibility, feature boundaries, and missing focused tests.

## High severity

### 1. Address verification does not verify the expected address

`verifyAddress` accepts the address to verify, but never passes it to Trezor Connect and never compares the returned address.

```dart
Future<bool> verifyAddress({
  required String address,
  required String derivationPath,
  required ScriptType scriptType,
}) async {
  final result = await _connect.getAddress(
    normalizedPath,
    showOnTrezor: true,
    coin: 'btc',
    scriptType: _inputScriptTypeFor(scriptType),
  );

  return result != null;
}
```

The pinned `trezor_connect` package supports an `address` validation parameter. Without using it, the app can show success for whatever address Trezor derived from the path, not necessarily the receive address displayed in Bull Bitcoin.

Fix: pass `address: address` to `getAddress`, and also compare `result.address == address` before returning success.

### 2. Master fingerprint cache can corrupt imports across devices

The repository stores one global cached fingerprint:

```dart
String? _cachedMasterFingerprint;
```

It is populated from the first account bundle and later consumed during import:

```dart
final masterFingerprint = await _trezorRepository.getMasterFingerprint();
```

If the user imports Trezor A, then Trezor B in the same app session, B's xpub can be paired with A's fingerprint. That produces a wrong descriptor origin.

Fix: return the master fingerprint with the same `TrezorAccount` result used for import, or return a single import-ready value object containing account + fingerprint. Do not keep an unkeyed repository-level cache.

### 3. BIP49 is exposed but change outputs use the wrong script type

The UI offers Nested SegWit/BIP49. Inputs are mapped correctly:

```dart
ScriptType.bip49 => 'SPENDP2SHWITNESS',
```

But P2SH outputs are always mapped as generic P2SH:

```dart
return 'PAYTOSCRIPTHASH';
```

For wallet-owned change outputs in a BIP49 wallet, Trezor should receive the wrapped SegWit change-output type, not generic P2SH.

Fix: classify change outputs using the wallet/account script type. For BIP49 change, emit `PAYTOP2SHWITNESS`. If BIP49 is not fully supported, remove it from this PR's import UI.

### 4. Trezor callback failures can hang until timeout

The pinned package callback handlers parse every callback as success:

```dart
completer.complete(TrezorAddress.fromJson(response["payload"]));
```

They do not check `success: false` / `payload.error`. If Trezor Suite returns a rejection, validation error, or other failure, `handleCallback` can throw while the original future remains pending. The app then waits for the long timeout instead of showing the actual error.

Fix: wrap or fork the package integration so callback payloads are parsed explicitly. On `success == false`, complete the operation with a typed `TrezorApplicationError`, remove the callback, and surface the real error immediately.

### 5. Model One import can fail after user approval

The import path requires parsing the master fingerprint from `raw.first.descriptor`:

```dart
final fp = _extractMasterFingerprint(raw.first.descriptor);
```

But the pinned model documents `descriptor` as not available for Model One. The UI advertises generic Trezor support, so Model One users can approve xpub export and then hit an unknown fingerprint/import failure.

Fix: add a tested fallback path for fingerprint retrieval, or explicitly gate supported models/platforms and explain the limitation in localized UI.

### 6. Trezor depends directly on watch-only internals

Trezor imports and returns watch-only feature internals:

```dart
import 'package:bb_mobile/features/import_watch_only_wallet/watch_only_wallet_entity.dart';
```

That creates a cross-feature dependency on implementation details instead of a public contract.

Fix: expose a watch-only public import/finalize API, or move shared descriptor/import DTOs into a neutral module. Trezor should not import another feature's internal entity/router files.

### 7. Screens bypass presentation boundaries and call use cases directly

Widgets resolve use cases from the service locator:

```dart
locator<GetTrezorAccountsUsecase>().execute(...)
locator<PrepareTrezorImportUsecase>().execute(...)
locator<SignPsbtTrezorUsecase>().execute(...)
locator<VerifyAddressTrezorUsecase>().execute(...)
```

This puts flow orchestration in UI code and bypasses the repo's UI -> presentation -> application structure.

Fix: add typed presentation controllers/Cubits for import, sign, and verify. Widgets should dispatch intent and render typed state, not call application use cases directly.

### 8. Missing focused tests for hardware-wallet signing and verification

The PR adds no focused tests for the risky paths. This needs coverage before merge because the PR adds new behavior around receive-address verification, PSBT signing, fingerprint derivation, deeplink callbacks, BIP49 handling, testnet handling, and route lifecycle.

Fix: add tests for address mismatch, expected address passed to Trezor Connect, fingerprint ownership, Model One missing descriptor, callback `success:false`, launch failure / timeout mapping, BIP49 change output script type, testnet coin/path behavior, late callback after route disposal, and missing route params.

## Medium severity

### 9. Testnet path/coin handling is hardcoded to mainnet

Import uses `m/<purpose>'/0'/...`, and sign/verify use `coin: 'btc'`. `isTestnet` is only used for local BDK address conversion.

Fix: thread network through import, verify, and sign. Map coin type and Trezor coin label consistently, or explicitly block Trezor on testnet.

### 10. UI imports low-level framework datasource

The deeplink listener reaches directly into the framework adapter:

```dart
locator<TrezorConnectDatasource>().connect.handleCallback(uri);
```

Fix: expose callback handling through a Trezor application/public service and remove the public raw `connect` getter.

### 11. Back/cancel during Suite round trip is not handled

The Cubit awaits an external app round trip and emits afterward without `isClosed` checks.

Fix: guard post-await emits with `if (isClosed) return;`, and either disable back during pending operations or explicitly cancel/ignore stale callbacks.

### 12. No-Suite / no-response failures wait too long

Import/sign can wait 5 minutes, verify waits 60 seconds. The package returns `launchUrl` success/failure but the Bitcoin helpers ignore it.

Fix: detect launch failure where possible, distinguish "could not launch" from "Suite/device did not answer," and show localized recovery copy quickly.

### 13. Route params are fragile because they depend on nullable `GoRouter.extra`

On route restore, direct navigation, or process death while Suite is foregrounded, the screen can rebuild without required params and only fail after the user taps.

Fix: redirect before constructing the page if params are missing, or encode restorable state in path/query/persistent pending-operation state.

### 14. Broad `bullbitcoin://*` scheme registration is too wide

Android registers the entire `bullbitcoin` scheme with no host restriction. Dart filters `trezor-callback`, but the platform-level ownership is still broader than this feature needs.

Fix: restrict Android by host/path. On iOS, consider a dedicated Trezor callback scheme if host-level narrowing is not available.

### 15. Generic Trezor capability overclaims model/platform support

The app stores a generic `trezor` signer and then enables import/sign/verify off that marker.

Fix: persist enough capability metadata to know which model/platform/account type supports which operation, or gate unsupported actions at import/action entry points.

### 16. Hardcoded English copy throughout the new UX

The new Trezor screens and send/receive entry points add user-facing English strings directly in Dart.

Fix: move user-facing strings into ARB localization files and regenerate l10n.

### 17. Verify-address copy points users at Suite instead of the device

The trusted comparison should be the hardware display, not the companion app.

Fix: change copy to compare against the address shown on the Trezor device.

### 18. Feature graph was not updated

The PR adds a new `trezor` feature and wires it into import/send/receive/router, but does not update the feature graph.

Fix: update `FEATURES.md` after correcting the intended boundaries.

### 19. Missing Linux generated plugin files after adding `app_links`

Adding `app_links` changes tracked Linux generated plugin registration files, but they are not in the PR diff.

Fix: commit `linux/flutter/generated_plugin_registrant.cc` and `linux/flutter/generated_plugins.cmake`, or gate/remove Linux plugin expectations.

## Low severity / cleanup

### 20. Verify screen copies a spaced address

The verify screen renders the selectable address with spaces inserted every four characters. Copying this text can place an invalid spaced address on the clipboard.

Fix: visually group the address without changing selectable/copy value, or provide a raw-address copy button.

### 21. Dead `processing` state

`processing` is rendered by screens but never emitted by the Cubit.

Fix: emit it for a real phase or delete the state and branches.

### 22. Generic `dynamic` Cubit result

`TrezorOperationCubit` uses `Future<dynamic> Function()` and stores `dynamic result`, forcing runtime type checks in screens.

Fix: use typed flow-specific Cubits or make the operation state generic.

### 23. Unused signing `derivationPath` route param

The sign route requires `derivationPath`, but signing never uses it.

Fix: remove the param and the guard, unless signing validation starts using it.

### 24. `TrezorSignedPsbt` DTO adds little value

The caller immediately projects to `serializedTxHex`.

Fix: return the tx hex string directly, or keep the DTO only if txid/metadata is actually needed.

### 25. Premature multi-account API

The API supports `startIndex/count`, but the UI always requests `count: 1` and uses `accounts.first`.

Fix: simplify to `getDefaultAccount(scriptType)` or add real account selection.

## Dropped / merged finding

I would not keep "dependency behavior not verified enough" as a standalone review comment. It is true as a theme, but the actionable dependency problems are already covered by the concrete issues above: address validation, callback error parsing, BIP49 script types, launch failure behavior, and Model One descriptor behavior.

BullishNode:
@anipy1 thanks for the updates. Please let me know when you are done and want the second review!

anipy1:
@BullishNode still working on it, will let you know when I'm ready for the second-review

anipy1:
@BullishNode this is ready for the second review.

For the "Feature graph was not updated" review 18, I checked, the graph treats hardware-wallet integrations as a single `HW_WALLETS` umbrella node; Ledger and BitBox aren't broken out as individual nodes either, and per-feature call sites into the umbrella (`RECEIVE → HW_WALLETS` for verify-on-receive, `SEND → HW_WALLETS` for sign-on-send) aren't currently tracked. Adding only Trezor breakout would be inconsistent with how Ledger and BitBox have been represented since they were introduced. Let me know if you'd prefer me to update the graph anyway.

Related to the localization review 16, I migrated all UI-layer strings into `app_en.arb`, but the error strings inside `TrezorOperationBaseCubit._messageFor` deliberately stay hardcoded English since the cubit doesn't have `BuildContext` access; localizing them properly requires a bit of state-shape restructuring. Let me know if you'd prefer that refactor

BullishNode (changes requested):
Requesting a small follow-up pass after the second review.

The major first-review Trezor issues look materially improved: address verification now passes and compares the expected address, fingerprint cache pollution is gone, BIP49/testnet handling is better, callback failures are handled in the fork, route guards exist, and the focused Trezor tests pass locally.

**Must fix before merge**

1. **Resume fallback can emit error, then later success**
   `TrezorOperationBaseCubit._onResumeGraceExpired` emits `suiteUnresponsive`, but the original awaited operation can still complete later and emit success. Please add an operation id/cancel/stale-completion guard, plus a test for "operation resolves after fallback error."

2. **Lockfile has unrelated dependency churn**
   `pubspec.yaml` adds `app_links` and `trezor_connect`, but `pubspec.lock` also bumps unrelated packages such as `bloc`, `flutter_svg`, `go_router`, `sentry_flutter`, `url_launcher_*`, and `vector_graphics`. Please revert unrelated lockfile changes unless they are required by the solver, or split/justify them.

3. **A couple Trezor user-facing strings still bypass l10n**
   `send_screen.dart` hardcodes `Transaction signed by Trezor`; `receive_qr_screen.dart` hardcodes the Trezor missing-wallet/address error. Please move these to ARB keys.

**Clean up or explicitly defer**

4. **Unused public facade/scaffolding**
   `TrezorFacade` is registered but unused, while callers import/use the Trezor router directly. It also exposes another feature's `WatchOnlyDescriptorEntity`. Please delete it for now, route callers through it, or explicitly defer public-boundary cleanup.

5. **Trezor folder shape vs updated architecture docs**
   The new Trezor slice uses `application/`, `adapters`, and `frameworks`, while the updated docs push new code toward `domain/usecases`, `domain/repositories`, `data`, and `data/datasources`. Please either align this slice or explicitly mark it as a follow-up/exception.

Accepted for this beta scope:
- Trezor import remaining behind the superuser gate.
- Full Trezor error localization/state refactor deferred to a later cleanup.
- No `FEATURES.md` graph update for this PR.

Verification I ran locally:
- `dart pub get --enforce-lockfile`
- `dart run build_runner build`
- `flutter gen-l10n`
- `flutter analyze --fatal-warnings --fatal-infos`
- `dart fix --dry-run` -> `Nothing to fix!`
- `flutter test test/features/trezor` -> 27 tests passed.

anipy1:
Items 1-4 of the second review is fixed and pushed. For the 5th (folder shape update following the updated architecture docs) I need some time to go through the architecture docs carefully before I work on the migration.

anipy1:
@BullishNode updated it following the updated architecture.

ethicnology:
@anipy1 thanks for your contribution!

I'm merging it on a local branch so we can test and finalize easily the integration.

@wired-pasteque can you give it a try please ?

Issue #2132, opened 2026-05-18
Improve UX for automatic wallet sync on app refresh


thibistaken:
Fixes https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2090

Improve the user experience when the app automatically syncs wallet data with the network, ensuring the latest transaction data is always visible after key user actions and app lifecycle events.

## User stories

- As a user, I want to see a visual sync indicator when I open the app, so I know my wallet data is being updated
- As a user, I want my wallet to automatically sync when I return to the app after it was backgrounded, so I always see the latest transactions
- As a user, after completing a transaction (send, receive, buy, swap, RBF), I want to see the wallet syncing when I return to the home screen, so I can confirm the transaction is reflected
- As a user, I don't want redundant syncs running simultaneously, so the app stays performant

## Scope

- Centralized sync orchestration with per-kind dedup (no concurrent duplicate syncs) and sequential FIFO queue (bitcoin → liquid → swaps)
- Gate sync on app lifecycle: auto-trigger sync on `paused → resumed`, skip sync while backgrounded
- Show pull-to-refresh spinner on first arrival (cold boot, post-activity redirect, env change, PIN unlock)
- Post-activity flows (broadcast, RBF, swap, buy, receive, ARK setup, vault recovery, PIN unlock, recovery test) trigger a wallet refresh before navigating home

## Related

- PR: https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2090

## Test cases

- [x] Cold boot (no PIN) → spinner appears on wallet home, dismisses on completion
- [ ] Cold boot (PIN set) → unlock → spinner appears post-unlock
- [x] Pull-to-refresh manually → spinner shows, runs once
- [x] Pull during in-flight sync → no second sync triggered
- [ ] Background app >30s, foreground → silent auto-sync runs
- [x] Brief inactive (control-center swipe) → no auto-sync
- [x] Receive payment → return to wallet home → spinner shows
- [ ] Buy success → close → spinner shows
- [ ] Broadcast tx → done → spinner shows
- [ ] RBF → spinner shows on landing
- [x] Swap "go home" → spinner shows
- [x] Send cancel → return to wallet home → no spinner (correct)
- [ ] Sync failure (turn off wifi mid-pull) → spinner dismisses, retry works
- [ ] Env change in settings → pop back → spinner shows

kiranmetri:
[...]

PR #2228, opened 2026-06-01
refactor(sync): sync on return to home

### LINKED ISSUE #2222: sync on return to home
Regardless of where you are in the app when you return to home it should trigger a sync

ethicnology:
https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2132
https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2222

## Changes

**fix(wallet) — trigger when the router lands on home**
- Replace the shell-blind `RouteObserver`/`RouteAware` with a `GoRouter` location listener: refresh when `routerDelegate.currentConfiguration.uri.path` becomes the wallet-home path. Works across the shell boundary for pop-back, tab switch, and push.
- Keep an on-mount trigger for cold boot / tab remount.
- Remove the now-unused root `RouteObserver`.

**refactor(sync) — slim the coordinator + gate to resumed-only**
- Drop the unused state-broadcast surface (`SyncCoordinatorState`, `StreamController`, `stream`/`state`, `_emit`) — zero consumers; relocate `SyncCoordinatorException` into the coordinator.
- Gate syncs to `AppLifecycleState.resumed` only (was a paused/hidden denylist that also let `inactive`/`detached` through). On return to resumed, replay a catch-up sync so a gated request is never lost.
- Collapse ~15 fine-grained log lines + the per-kind severe into one summary per `sync()`: `fine` on success, `info` on a dropped sync (gated/throttled/pending), `severe` on failure — sanitized so only `kind: runtimeType` reaches Sentry.

Unchanged: per-kind dedup, FIFO ordering, 2s throttle, and the `WalletRefreshed` → bloc → coordinator funnel (`droppable()` still drops overlapping refreshes).


ethicnology:
Hey @claude can you make a pre-review / audit of this branch ?

i5hi (approved):

PR #2229, opened 2026-06-02
fix: display autoswap active status when enabled

### LINKED ISSUE #2196: Autoswap active warning no longer appears on home screen
The warning indicator that autoswap is active no longer appears on the home screen. Users have no visual cue that autoswap is enabled.

## Expected behavior

- Home screen displays a visible warning/indicator when autoswap is active

wired-pasteque:
Fix: #2196

Simulator Screenshot - iPhone 16e - 2026-06-01 at 17 42 50

### DIFF
--- a/lib/features/wallet/presentation/bloc/wallet_state.dart
+++ b/lib/features/wallet/presentation/bloc/wallet_state.dart
@@ -81,8 +81,6 @@ sealed class WalletState with _$WalletState {
   }
 
   bool showAutoSwapActiveStatus() {
-    return autoSwapSettings != null &&
-        autoSwapSettings!.enabled &&
-        !autoSwapSettings!.showWarning;
+    return autoSwapSettings != null && autoSwapSettings!.enabled;
   }
 }


lib/features/wallet/presentation/bloc/wallet_state.dart
```diff
@@ -81,8 +81,6 @@ sealed class WalletState with _$WalletState {
   }
 
   bool showAutoSwapActiveStatus() {
-    return autoSwapSettings != null &&
-        autoSwapSettings!.enabled &&
-        !autoSwapSettings!.showWarning;
+    return autoSwapSettings != null && autoSwapSettings!.enabled;
```
ethicnology, line 84: What is the use of  `showWarning` ?
Is it still used anywhere ?
If not, should we clean it ?

ethicnology (commented):
What is the use of  `showWarning` ?
Is it still used anywhere ?
If not, should we clean it ?

lib/features/wallet/presentation/bloc/wallet_state.dart
wired-pasteque: @ethicnology ,`showWarning` is still used in multiple places in autoswap logic, but in our case it should not be used for the home screen active indicator, since it incorrectly hides the enabled state.

ethicnology: Any inputs @i5hi ?

i5hi: yeah, showWarning is the warning shown the first time to a user to let them know autoswap is enabled - when they see the warning they have to deal with it via `I understand the risk` or `I accept`. If they just close it, showWarning stays true and they see it again when they restart the app - but autoswap will not actually trigger.

For autoswap to be actually trigger (or be truly enabled from user perspective) it has to be enabled && !showWarning

i5hi: its interesting that this fixes the issue because, for users who are not seeing it, somehow showWarning is not set to false and they are also not seeing the warning to be able to deal with it and make the flag become false

i5hi: if showWarning is true, then even though its enabled - autoswap will not trigger

i5hi (approved):

PR #2230, opened 2026-06-02
fix(broadcast_signed_tx): show broadcast failures and pin action buttons

### LINKED ISSUE #1247: More useful details on broadcast screen
When broadcasting a signed PSBT, we are seeing raw tx details that are not useful for the user. 

What we need to see is:

Receiving address + amount 
Change address + amount 
Network fee amount

ethicnology:
Broadcast failures were caught in the cubit but never rendered: the error
UI only existed in the paste/scan branch (transaction == null), so a rejected broadcast (already confirmed, fee too low, no servers reachable)
silently re-enabled the button with no feedback — which users read as "the
broadcast button doesn't work".

- Render a localized error in the review screen via a new typed BroadcastFailedError; the raw Electrum/node detail goes to log.warning (file + console, no Sentry) instead of leaking to the UI.
- Add broadcastSignedTxBroadcastError and translate it across all locales.
- Pin the action buttons to the bottom (bottomNavigationBar + SafeArea) and move the error into the scroll content, so the buttons no longer jump when the error appears/disappears.

### DIFF
--- a/lib/features/broadcast_signed_tx/errors.dart
+++ b/lib/features/broadcast_signed_tx/errors.dart
@@ -9,6 +9,10 @@ class InvalidTxError extends BroadcastSignedTxError {
     : super('Input is not a transaction using PSBT or HEX format');
 }
 
+class BroadcastFailedError extends BroadcastSignedTxError {
+  BroadcastFailedError() : super('Failed to broadcast the transaction');
+}
+
 class PushTxNoNdefRecordsError extends BroadcastSignedTxError {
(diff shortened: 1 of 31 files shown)


ethicnology:
@kiranmetri and @wired-pasteque you can give a try to this one.

If it fails to broadcast it should show an error, and the precise reason why it failed should appears in the logs.

ethicnology:
Any opinion @claude ?

ethicnology:
Are you satisfied now @claude ?

i5hi (approved):

PR #2232, opened 2026-06-02
feat(android): add installable beta tester channel

ethicnology:
Adds a **`beta`** Android product flavor: a release-mode build that installs
**alongside** the production app, for handing release candidates directly to
testers (no store). Replaces the practice of shipping **debug** APKs as candidates.

## Why

We were giving testers **debug** builds, which are too big and not
representative of what ships. Neither existing build fits a tester channel:

- **debug** — Dart runs in **JIT** (full VM + uncompiled kernel snapshots, no
  tree-shaking, VM service exposed), so the APK is ~2–3× a release build, slower,
  and not something you want in testers' hands for a wallet.
- **release** — shares the production `applicationId` and needs the **production
  signing key**, so it would *overwrite* a tester's real app and force the prod
  key into the build pipeline.

`beta` gives the small, optimized, AOT-compiled output of a release build, but as
a **separate, independently-signed app**.

## beta vs release vs debug

| | debug (old candidate) | release (official) | **beta (this PR)** |
|---|---|---|---|
| Dart compilation | JIT | AOT (optimized) | **AOT (optimized)** |
| APK size | ~2–3× larger | small | **small (≈ release)** |
| applicationId | `…mobile.debug` | `…mobile` | **`…mobile.beta`** |
| Installs next to production? | yes | **no — replaces it** | **yes** |
| Signing key | public debug key | production release key | **own dedicated beta key, no fallback** |
| In-app marker | DEBUG banner | none | **BETA banner** + "Bull Beta" label |
| Fit for a tester channel? | no | no | **yes** |

## How it works

- **`beta` flavor** (`android/app/build.gradle`): `.beta` applicationId suffix,
  `-beta` versionName, "Bull Beta" launcher label (`src/beta/`), and an in-app
  **BETA** corner banner (`appFlavor == 'beta'`) since release mode drops the
  Flutter DEBUG banner. The `.beta` id means its own sandbox/storage — a tester's
  real wallet is never touched.
- **Per-flavor signing**, isolated by design:
  - **release** → the production key (applied on the signing machine; unsigned
    elsewhere, e.g. in the reproducibility/verify flow).
  - **beta** → its **own dedicated key only**, provided by CI from secrets.
    There is **no debug-keystore fallback**: a beta built without its key is left
    **unsigned** (never signed with the public debug key), so a distributed beta
    can't be forged with the repo's public key.
  - **debug** → the committed public debug keystore.
  - The `release` build type sets no signingConfig on purpose — AGP resolves a
    build-type signingConfig ahead of a flavor's, so signing is set per-flavor.
- **Official build untouched**: the `production` flavor adds no applicationId or
  version overrides and keeps the release signing key, so reproducible-build
  verification and in-place updates of already-distributed releases are unaffected.
- ⚠️ **Reproducibility note**: selecting any flavor makes Flutter bake
  `FLUTTER_APP_FLAVOR` into `libapp.so`, so the production binary changes **once**
  at this commit. The build stays deterministic, so `make verify` (rebuilds the
  matching tag with the same flavor) still works — it's a one-time hash change,
  not a verification break.

## Usage

```bash
make android beta          # -> ./BULL-beta.apk  (release mode, .beta id)
```

With no local `android/key-beta.properties`, the beta build is unsigned and is
extracted as `./BULL-beta-unsigned.apk` to make that obvious. `make android
release|debug` are unchanged (the `apk` target was renamed to `android` and host
artifacts branded `BULL-<target>`).

In CI: **Actions → Build Android APK → mode `beta`** signs with the beta key from
repo secrets (`BETA_KEYSTORE_BASE64` / `BETA_KEY_ALIAS` / `BETA_KEY_PASSWORD` /
`BETA_STORE_PASSWORD`). If any are missing the build **fails loud** rather than
producing an unsigned/forgeable artifact. It uploads `BULL-beta.apk`; no release
publishing, no token — distribution is manual: attach to a GitHub **pre-release**
and testers auto-update via Obtainium.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -11,6 +11,8 @@ ios/
 .fvm/
 
 # Local APKs/AABs built previously (host-side artifacts)
+BULL-*.apk
+BULL-*.aab
 app-*.apk
 app-*.aab
 
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -5,7 +5,7 @@ on:
     inputs:
       mode:
         type: choice
-        options: [release, debug]
+        options: [release, debug, beta]
         default: release
 
 jobs:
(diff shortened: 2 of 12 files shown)


ethicnology:
Please @claude give your opinion on that

ethicnology:
@claude is it better now?

PR #2249, opened 2026-06-05
fix: transfer flow should allow receive exact amount option for liquid. better errors in form validation. escalation backup warning content as a bulleted list.

i5hi:
Fixes:
Transfer Flow:

- [x] Show the `Recieve Exact Amount` option when the from wallet is liquid 
- [x] Insufficient balance should be a form validation error
- [x] Continue should not be clickable if form is not validated
- [x] Split the single large warning into two small warnings after transfer is confirmed

Backup Content:
- [x] Display content as a bulleted list for easier reading

Simulator Screenshot - iPhone 17 Pro - 2026-06-05 at 17 19 07

### DIFF
--- a/lib/core/errors/send_errors.dart
+++ b/lib/core/errors/send_errors.dart
@@ -4,6 +4,10 @@ class SwapCreationException extends BullException {
   SwapCreationException(super.message);
 }
 
+class InsufficientFundsSwapException extends SwapCreationException {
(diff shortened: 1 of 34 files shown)


i5hi:
Fixes https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2264

PR #2259, opened 2026-06-08
fix(payjoin): correct malformed receive BIP21 URI

ethicnology:
### What was wrong
Our receive payment request was rebuilt through `dart:core Uri`, which
percent-encoded the pj endpoint's `://` (`HTTPS%3A%2F%2F…`) and dropped
`pjos`. The result was an out-of-spec, malformed BIP21 URI.

### The fix
We now take the PDK-produced URI (already valid) and append `amount`/`message`
verbatim instead of re-encoding it. The pj endpoint stays byte-identical and
last (BIP-77 SHOULD), and we advertise `pjos=0` when the PDK omits it — which
matches our receiver, since it never substitutes outputs. Covered by unit
tests: old-behavior characterization, new generator, and a decode round-trip.

- The over-encoding and missing `pjos` were **our** bug (BULL generation), not
  the `bip21_uri` package — the package's encoder keeps `://` clean.
- The package's *decoder* does mutate the pj (force-uppercase, `-`→`+`). That's
  a separate, known issue, documented as skipped tests.

### ⚠️ Still locked on payjoin 0.23.0
We have not changed the payjoin engine. 0.23.0 uses the **old `+` fragment
separator**; final BIP-77 uses `-`. So:
- This fix makes our URI clean and in-spec *for the 0.23 wire format* — it does
  **not** fix cross-version interop.
- The `bip21_uri` decoder's `-`→`+` rewrite is currently **load-bearing**: it's
  what lets us accept a final-BIP-77 (`-`) URI on the send path against the 0.23
  PDK. Removing it before upgrading would break sending.
- So the package cleanup and the PDK upgrade to a final-BIP-77 release **must
  land together** — not before.

### DIFF
--- a/lib/features/receive/presentation/bloc/receive_state.dart
+++ b/lib/features/receive/presentation/bloc/receive_state.dart
@@ -111,7 +111,7 @@ abstract class ReceiveState with _$ReceiveState {
           scheme: 'bitcoin',
           path: bitcoinAddress!.address,
           queryParameters: {
-            if (confirmedAmountBtc > 0) 'amount': confirmedAmountBtc.toString(),
+            if (confirmedAmountBtc > 0) 'amount': formatBtcAmount(confirmedAmountBtc),
             if (note.isNotEmpty) 'message': note,
           },
         );
@@ -128,7 +128,7 @@ abstract class ReceiveState with _$ReceiveState {
           scheme: 'liquidnetwork',
           path: liquidAddress!.address,
           queryParameters: {
-            if (confirmedAmountBtc > 0) 'amount': confirmedAmountBtc.toString(),
+            if (confirmedAmountBtc > 0) 'amount': formatBtcAmount(confirmedAmountBtc),
             if (note.isNotEmpty) 'message': note,
             'assetid':
                 wallet != null && wallet!.network == Network.liquidMainnet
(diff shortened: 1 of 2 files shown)


ethicnology:
Good catch @claude on the amount formatting. Added `formatBtcAmount` (toStringAsFixed(8) + trim) and applied it to all three receive paths (payjoin, bitcoin, liquid), so the pre-existing non-payjoin bug is covered too. New tests assert 1/10/100-sat amounts render as plain decimals.

Re: the base address — intended. The payjoin receiver is created with `bitcoinAddress.address` (`createReceiver(address: ...)`), so the PDK URI's base address is that same address. Taking it from the PDK URI just avoids re-assembling a string we already have in valid form.

Is it good enough now @claude ?

ethicnology:
Hey @DanGould can you review that temporary fix until we prioritize https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2041

DanGould:
FYI there shouldn't be any wire format changes from 0.23 to 1.0.0-rc3 or cross version interop problems *other than* the URI. Technically 0.23 is out of spec but 1.0.0-rc3 iirc does accept the old URI for compatibility's sake for the time being so we don't kill clients' whose newest versions still aren't up to 0.25+. This will not be forever.

> This fix makes our URI clean and in-spec for the 0.23 wire format — it does not fix cross-version interop.

There's no cross-version interop issues between 0.23,0.24,0.25, 1.0.0-rc3 afaiu beyond this issue if you consider it a wire format issue

DanGould:
> we advertise pjos=0 when the PDK omits it — which matches our receiver, since it never substitutes outputs

["Backwards-compatible receivers MUST disable output substitution by setting pjos=0 to prevent modification by a malicious directory](https://github.com/bitcoin/bips/blob/master/bip-0077.md)" this is NOT about BBMobile substituting outputs, actually pjos=0 + output substitution is OK for BIP 77. it's for the BIP 78 backwards compatibility, where a malicious directory/proxy in the middle could actually STEAL FUNDS by malleating the plaintext if this isn't set properly

PR #2261, opened 2026-06-08
Use env only for in debug and profile

i5hi:
Move all staging envs to a .env

Ensures env is not used for release builds and only for debug or profile builds


i5hi:
- [ ] --dart-define-from-file=.env flag should be implmented wherever we use flutter run - makefile, container...

i5hi:
Dart format went to town on almost every single file :/

PR #2263, opened 2026-06-09
26-06-09_Update Translation_de_DE

bsn21m:
Hello, 

after yesterday’s beta I made a few additional linguistic adjustments to the German localization. While reviewing it, I noticed that some UI wording was not ideal and that certain words or labels are too long for specific buttons, tabs, or views.

I also noticed that the month names do not yet seem to be fully covered by localized strings/keys. In addition, the dollar symbol does not feel appropriate in every language and currency context. From a European point of view, it might feel more natural either to use a neutral symbol, for example a coin, or to adapt the icon dynamically to the selected currency.

More generally, I also have the impression that some strings/keys are still missing in the current app structure.


bsn21m:
Hello,

I also noticed that app_de.arb still contains a few keys that are currently not present in app_en.arb, but they still exist in several other localization files as well. Because of that, they do not look like German-only leftovers, but rather like older or parallel localization entries that may still be relevant.

At the moment, this affects:

autoswapWarningTriggerAmount
autoswapWarningBaseBalance
homeLegacyStorageWithNoBackupTitle
homeLegacyStorageWithNoBackupDescription
homeLegacyStorageTitle
homeLegacyStorageDescription
legacyStorageWarningBackupNow
legacyStorageWarningLater
errorReportingMigrationSubtitle
transactionSwapOpenWithin24Hours

So I would not remove these keys without checking first. It may be worth confirming whether they are still intentionally used, or whether the English file should be cleaned up or aligned in that area.

ethicnology:
Thanks @bsn21m 

I've removed these keys in #2285 

You can rebase on develop to remove them from your file as well before we merge this PR ?

PR #2290, opened 2026-06-13
chore: clean up swaps

i5hi:
Swap rework: reliable status tracking, claims/refunds, and live fees

Reworks the swaps feature end-to-end to fix unreliable Boltz status updates and failing claims/refunds, and to make persisted fees reflect what was actually paid on-chain. Only includes logic improvements - no fundamental architectural changes.

All swap types need to be tested before merging.

- [x] Reverse swaps into LBTC and BTC from LN via Receive
- [x] Submarine swaps from LBTC and BTC to LN via Send
- [x] Chain swaps to and from LBTC - BTC via Transfer
- [x] Chain swaps to and from LBTC - BTC via Send


i5hi:
Note: production flavor is commented out for local ios simulator development. It must be removed before merging.

i5hi:
Note: debug logs must be cleaned before merging

PR #2307, opened 2026-06-17
refactor(devcontainer): split into per-OS linux/ and macos/ configs

ethicnology:
The single combined config couldn't start on macOS. Split .devcontainer into two self-contained subfolders that build the same Containerfile.tools image but differ only in host integration:

- macos/: drops what the applehv podman VM can't provide — --device=/dev/dri, X11/Wayland mounts+env, SSH-agent forwarding (impossible via virtiofs), and the keyring/dbus bootstrap (libsecret only runs in the Linux GUI app, which macOS can't display). Runs Rosetta setup in initializeCommand. Verified working on Apple Silicon (amd64 via Rosetta — pinned for reproducible builds).
- linux/: unchanged behavior (GPU + X11/Wayland + ssh-agent + keyring retained).

Each folder carries its own lifecycle scripts (not shared); macos uses a slim init-mounts.sh and its own macos-setup.sh.

make devcontainer auto-detects the host OS (override DEVCONTAINER_OS=), reuses an existing BULL container instead of recreating it, and replays the linux keyring on restart. Dropped the redundant container-tools prerequisite — the devcontainer CLI builds its own image (USERNAME=$USER cache-busts vs bull-tools).

Cleanups: removed no-op consistency=cached (ignored by podman), documented the adb -a LAN exposure and hardcoded DNS, and fixed the adb version comment.


basantagoswami:
@claude check now. the PR is updated

basantagoswami (approved):

basantagoswami:
Tested it on Linux, LGTM

PR #2310, opened 2026-06-17
store and read staging exchange basic auth creds from db

### LINKED ISSUE #2291: Staging exchange integration not working
Since we changed the basic auth credentials for the staging exchange environment, the exchange can no longer be accessed in testnet mode. The fix is to introduce a .env for the debug build, or to allow adding the creds for people under dev mode

basantagoswami:
Fixes #2291


basantagoswami:
@claude

- I am modifying the already created schema v13 because it's only in develop and one of the recently merged PR added it. That's how we do things typically. Else you'll end up with a bunch of new schemas for no reason
- The password field is not obscured deliberately. In this context the username and password should be thought of as configs instead of credentials
- The credentials can stay in git history, it's not an issue because they have been changes already. Also, it's deliberate for the staging exchange integration to not work unless someone has configured the basic auth username/password

i5hi (approved):

Issue #2243, opened 2026-06-04
Scrolling on home to view multiple wallets is not constrained to the wallet list section


i5hi:
## Description
When we scroll to view other wallets , all ui elements move up. 

### Expected behavior
Only the wallet list section should be scrollable to view wallets.

<img width="1206" height="2622" alt="Image" src="https://github.com/user-attachments/assets/daade781-0a05-4c6f-a832-c1b568399687" />

ethicnology:
They are floating over the list. We have a design issue there, because on small screen if we don't make the floating over the list, the default wallet themself will be truncated and the pull-to-refresh will be affected too.

Below a screenshot from a Samsumg S10e

Two improvements points:
1. We have the buttons Wallet / Exchange that are kind of useless, because as of today these two features are 100% coupled. You have the buttons Buy/Sell/Pay and you have exchange transactions in the list. 
2. The part of the screen with the background design from Madex take too much space.

Overall we need to redesign the wallet with this kind of details in mind.

For now, unless we achieve to decouple the exchange from the wallet, the button on the bottom shouldn't be on the home page.

<img width="1080" height="2280" alt="Image" src="https://github.com/user-attachments/assets/e630a8b2-5301-4123-808b-b6a220c23016" />


cc: @thibistaken @BullishNode

PR #2315, opened 2026-06-19
fix(wallet): keep the last wallet card visible on the home screen

### LINKED ISSUE #2314: Wallet home: last wallet card is cut off by Receive/Send buttons when 3+ wallets are present
## Description
On the Wallet home screen, when there are three or more wallet cards (e.g. Secure Bitcoin, Instant payments, Some Wallet), the last card cannot be fully scrolled into view, it stays hidden behind the floating Receive/Send button bar at the bottom of the screen.

### Reproduce 
1. Open the app on a wallet that has three or more cards on the home screen
2. Scroll the wallet card list to the bottom.
3. Observe that the last card is clipped by the Receive/Send button bar and cannot be fully revealed regardless of how far you pull.

### Expected behavior
The last wallet card should be fully visible and scrollable above the Receive/Send button bar

Image

anipy1:
Fixes #2314 

image

### DIFF
--- a/lib/features/wallet/ui/screens/wallet_home_screen.dart
+++ b/lib/features/wallet/ui/screens/wallet_home_screen.dart
@@ -118,47 +118,40 @@ class _WalletHomeScreenState extends State<WalletHomeScreen> {
       child: PopScope(
         canPop: false,
         onPopInvokedWithResult: (didPop, _) {},
-        child: Stack(
-          children: [
-            // Black background visible only during iOS top overscroll
-            Positioned(
-              top: 0,
-              left: 0,
-              right: 0,
-              child: ColoredBox(
-                color: AppColors.dark.background,
-                child: const SizedBox(height: 300),
-              ),
-            ),
-            BBPullableBody(
-              indicatorKey: _indicatorKey,
-              onRefresh: () async {
-                // User gesture — bypass the coordinator throttle.
-                final bloc = context.read<WalletBloc>();
-                bloc.add(const WalletRefreshed(force: true));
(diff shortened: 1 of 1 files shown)


basantagoswami:
LGTM

i5hi:
Tests pending

i5hi:
Thanks for the PR, which is at the core of this UI issue on the home page.

We have this older issue: https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2243

This PR does not really fix the core issue. If you are willing to take it up, that would be great.

Requirements:
- On home the entire page should not be scrollable, only the wallet section and it should always lock into only showing 2 wallets at a time
- The above may affect pull to sync behaviour. What we want is that the app should allow pull to sync anywhere on the screen - if the user is scolling up/pulling down to go up on their wallet list, then pull to sync should only trigger when they reach the top of the list and still pull down to sync

anipy1:
> Thanks for the PR, which is at the core of this UI issue on the home page.
> 
> We have this older issue: #2243
> 
> This PR does not really fix the core issue. If you are willing to take it up, that would be great.
> 
> Requirements:
> 
> * On home the entire page should not be scrollable, only the wallet section and it should always lock into only showing 2 wallets at a time
> * The above may affect pull to sync behaviour. What we want is that the app should allow pull to sync anywhere on the screen - if the user is scolling up/pulling down to go up on their wallet list, then pull to sync should only trigger when they reach the top of the list and still pull down to sync

Thanks for the feedback and for pointing #2243 I hadn't seen that issue when i opened this one.

Since #2243 already captures the proper scope. I'll close this PR (as #2314 as a duplicate of #2243)

Issue #2138, opened 2026-05-19
Bull Sdk Fixes


i5hi:
Fix several issues in the Bull SDK integration to improve build reliability and CI coverage.

## Acceptance criteria
- [ ] Pin \`bull_sdk\` git dependencies to commit SHAs (replace \`branch = "migrate-to-bull-sdk"\` with \`rev = "<sha>"\`)
- [ ] Pin the app's ref to a specific commit instead of \`main\`
- [ ] Add \`bull_sdk\` CI matrix: build each sub-crate standalone, build with all features, run \`fix_frb_generated.sh\` idempotency check
- [ ] Add mirror-type assertion test to CI (verify \`TxFee\` and \`ArkTransaction\` mirror declarations match upstream)
- [ ] Add cross-crate FRB version enforcement check in CI
- [ ] Plan for folding \`bdk_dart\` into bull_sdk (next step, not blocking this issue)

## Test cases
- [ ] CI builds pass with pinned commit refs
- [ ] Modify upstream \`TxFee\` struct and verify CI mirror assertion catches the drift
- [ ] Run \`fix_frb_generated.sh\` twice and verify idempotent output
- [ ] Verify each sub-crate builds independently in CI
- [ ] Verify cross-crate FRB version check fails when versions diverge

ethicnology:
@i5hi what is the new PR associated to this issue now?

i5hi:
https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2316

i5hi:
Lets keep this open to track publishing lwk and boltz to pub.dev

ethicnology:
https://github.com/bitcoindevkit/bdk-ffi/pull/1040

ethicnology:
https://github.com/bitcoindevkit/bdk-dart/pull/101

PR #2316, opened 2026-06-19
feat: swap mnemonic + restore endpoints

### LINKED ISSUE #2282: Swap mnemonic: Follow boltz new standard for swap mnemonic to enable recovery endpoints
Use the updated boltz-dart library which has a new flow for swap mnemonic. 

The wallet earlier derived the swap mnemonic from a bip44 special account path and required always passing the wallet mnemonic to the library to derive. 

Now, boltz-dart has a new helper method to derive the swap mnemonic which should be stored separately and subsequently for every new swap - only this mnemonic is passed into swap creation methods. 

What this enables:
- [ ] Recovery endpoints - using the recovery endpoints requires that swap keys are derived at a specific unhardened path so that boltz can derive all the pubkeys related to a user using the master xpub. This will enable full restoration of completed and in progress swaps on a fresh app installation
- [ ] Swap from hardware - previously, hardware wallets could only be on the receiving end of the swaps because all swap creation methods required a wallet api from which it derived the swap key. Since hardware wallets have no mnemonic, they could never be a sender. Using the swap mnemonic solves this problem. Swaps can now be created with the swap mnemonic and the hardware wallet just has to fund the lock up to be the sender. 

## User stories
- As a user who has swap history, I want to delete the app, restore from my seed and get back all my swap history
- As a user who has an incomplete swap and reinstalled the app, I want to be able to recover the swap/ complete it on my fresh installation

The UX does not change, only the internals of how swaps are handled changes. The user must not have to deal with the fact that they have a swap mnemonic - they must only be concerned with their main wallet mnemonic.

## Test cases
- [ ] Chain swaps via Transfer from BTC to LBTC
- [ ] Chain swaps via Transfer from LBTC to BTC
- [ ] Submarine swaps via Send from LBTC to LN
- [ ] Submarine swaps via Send from BTC to LN
- [ ] Reverse swaps via Receive from LN to LBTC
- [ ] Reverse swaps via Receive from LN to BTC
- [ ] Perform a chain swap, btc submarine swap and btc reverse swap and delete the app before they complete
- [ ] Reinstall and recover from the same wallet menmonic
- [ ] Entire swap history must be restored and ongoing swaps should be restored

i5hi:
Swaps now uses more deterministic secrets (preimage included) and are now derived from a standardized path that can be used to restore stuck swaps via the `restore` endpoint by sharing the master xpub with boltz. 

Fixes: https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2282 https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2138


i5hi:
Not yet ready for review. Just a first draft.

i5hi:
@claude Please run another review

ios/Podfile.lock
ethicnology, line 1: Isn't BLE required for Bitbox on iOS ?

lib/core/storage/tables/swaps_table.dart
```diff
@@ -30,6 +30,11 @@ class Swaps extends Table {
   IntColumn get serverNetworkFees => integer().nullable()();
   BoolColumn get wasDirectPayment =>
       boolean().withDefault(const Constant(false))();
+  // Reconstructed by the swap-restore/rescue flow rather than created in-app, so
+  // fields not derivable from the Boltz restore response + on-chain data are not
+  // trustworthy and are hidden in the UI (see RescueSwapUsecase / the swap
+  // entity's `recovered` doc for the exact list of uncertain fields).
+  BoolColumn get recovered => boolean().withDefault(const Constant(false))();
```
ethicnology, line 37: Is there any side effect of recovering a mnemonic without this metadata ?

lib/core/swaps/swaps_locator.dart
ethicnology, line 1: Is it normal to have this logic in the locator?

lib/core/storage/tables/swaps_table.dart
i5hi: kind of and this field is there to address that. the main side effect is that we dont have all the fields we had when we create the swap, for example:
- wallets involved
- boltz and miner fees at the time of the swap

In a later enhancement we can try to figure out which wallet was involved by scanning inputs and outputs of the lockup and claim but for now i left it out of the scope. 

The main feature affected by this is transaction details as well as the swap watcher itself. The presence of this field lets us know to handle things a bit different.

lib/core/swaps/swaps_locator.dart
i5hi: fixed.

Issue #1895, opened 2026-02-27
Sanitize All User-Facing Error Messages


thibistaken:
Audit all user-facing error messages and replace technical/cryptic strings with clear, actionable text. This is the parent tracking issue — #2140 covers API error sanitization specifically.

## User stories
- As a non-technical user, I want error messages to explain what went wrong and what I can do about it, so I'm not confused by technical jargon or class names

## Acceptance criteria
- [ ] Inventory of all error message sources (BDK, LWK, Boltz, API, network, local validation)
- [ ] Each error mapped to a user-friendly localized string
- [ ] No raw exception class names, stack traces, or technical identifiers shown to users
- [ ] Error messages include actionable guidance where possible (e.g. "Check your internet connection", "Try again later")
- [ ] Design review for error presentation (banners, dialogs, inline)

## Test cases
- [ ] Trigger a network timeout and verify the error message is user-friendly
- [ ] Trigger a Boltz swap failure and verify no technical details are exposed
- [ ] Trigger an insufficient funds error and verify the message includes the shortfall amount
- [ ] Verify all error strings are localized (at minimum English and French)
- [ ] Search codebase for raw \`.toString()\` error displays and verify none remain

## Context
- Related: #2140 (API error sanitization)

EDIT by @ethicnology :

## List

You can create a sub-issue and make a PR against this branch from it 

### Done (reference examples — do not redo)
- [x] `pin_code`
- [x] `app_unlock`
- [x] `import_watch_only_wallet`
- [x] `broadcast_signed_tx` (incl. `transaction_review`)
- [x] `labels`

### Tier A — has an existing error family (migrate to the pattern / fix the leaking catch-all)

- [ ] `transactions` — two error files (`transaction_error.dart`, `application_errors.dart`) → consolidate
- [ ] `wallet` — `wallet_error.dart` is sealed but has no `toTranslated`; add it + generic catch-all
- [x] `replace_by_fee` — abstract-class + `toTranslated` already; migrate to `sealed`, verify generic catch-all
- [x] `recoverbull` — `errors.dart` + `toTranslated`; verify catch-all doesn't leak
- [x] `recoverbull_google_drive` — `errors.dart` + `toTranslated`; verify catch-all
- [x] `bip85_entropy` — `errors.dart`, wraps `e.toString()`; convert + stop leak
- [x] `import_mnemonic` — `errors.dart`, raw strings to UI; convert + stop leak
[...]

PR #2330, opened 2026-06-22
refactor(recoverbull): sanitize errors with sealed failures and a Result boundary

ethicnology:
Replace raw exception text in the UI with typed, translated failures across both recoverbull features and their shared core layer.

core: add sealed RecoverBullCoreFailure; repositories become the data boundary (catch + log raw + map -> Result), use-cases propagate Result. Absorb decrypt deserialization, vault-file validation, Drive reduce-to-latest, key-server mapping, and vault assembly into the repositories. checkConnection stays a throwing bool probe so the shared status checker is untouched.

features: sealed RecoverBullFailure / RecoverBullGoogleDriveFailure in domain/, translation in presentation/*_failure_l10n.dart; blocs switch on Result and lift core failures to the feature failure; state holds the typed failure.

di: inject datasources into repositories; register PickVaultUsecase; blocs no longer self-instantiate use-cases.

Remove legacy errors.dart (feature + core); reparent TorNotStartedError to BullException. Fixes latent bugs surfaced by the boundary: empty-Drive reduce StateError, 429 cooldown NPE, and raw-string/generic Exception throws.


ethicnology:
@Claude audit that, it's the refactoring related to #1895 with the scope of recoverbull

ethicnology:
What about now @claude

ethicnology:
@Claude can you investigate any kind of regressions?

PR #2386, opened 2026-06-29
fix(sync): keep pull-to-refresh spinner until all chains sync

ethicnology:
Spinner released after bitcoin sync, not all kinds: the forced refresh could be dropped by droppable() while a throttled bitcoin-only background refresh ran, and sync() awaited the shared drain rather than its own requested kinds.

sync() now awaits its own kinds via per-kind completers (resolves once they settle regardless of which drain ran them; errors flow through the completers). Replaces the ambiguous `force` bool with SyncTrigger { user, automatic }. Adds WalletBloc.refresh(): awaits the coordinator directly so the spinner tracks real all-kind completion and bypasses the droppable lane; never throws. The 4 pull-to-refresh screens now await it. Adds SyncCoordinator unit tests incl. the join-in-flight regression.

   <table><tr>
   <td>

   https://github.com/user-attachments/assets/7b131754-1680-40cc-8c1c-0355f800fc36

   </td>
   <td>

   https://github.com/user-attachments/assets/665594ed-9906-429a-8a8a-8a32208afaa8

   </td>
   </tr></table>

### DIFF
--- a/lib/core/sync/sync_coordinator.dart
+++ b/lib/core/sync/sync_coordinator.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:collection';
 
 import 'package:bb_mobile/core/sync/sync_kind.dart';
+import 'package:bb_mobile/core/sync/sync_trigger.dart';
 import 'package:bb_mobile/core/swaps/domain/usecases/restart_swap_watcher_usecase.dart';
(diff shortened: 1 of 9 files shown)


i5hi (approved):

wired-pasteque:
@ethicnology, tested on my side - working as expected, spinner now stays until all chains finish syncing.

PR #2387, opened 2026-06-29
fix(wallet): make wallet-card sync bar visible in light theme

ethicnology:


i5hi (approved):

i5hi:
An image to see the final result would be good. We want a shade of grey that it subtle and not something too loud and obvious.

wired-pasteque:
@ethicnology, checked on my side - wallet-card sync bar is visible in light theme:

Screenshot_20260702-094244

PR #2392, opened 2026-07-02
ci(claude): use Fable 5 model and bump checkout to v7

ethicnology:
Pass `--model claude-fable-5` via claude_args so the review agent runs on Anthropic's most capable model (claude-code-action@v1 has no dedicated model input; the model is set through the CLI args). Also bump actions/checkout v6 → v7.


ethicnology:
@Claude review

PR #2421, opened 2026-07-08
add tool for working with loc (arb) files

basantagoswami:
When working on new features I realized that Claude always just adds the new key to `localization/app_en.arb` and doesn't touch the other files. Editing them via agents fails most of the time for some reason. Asking Claude to manually edit the file often lead to malformed JSON and consume a lot of context unnecessarily. So created this tool to make things easier


basantagoswami:
@claude check again

ethicnology (approved):

PR #2431, opened 2026-07-09
payjoin: exclude 0-conf inputs from receiver candidates

### LINKED ISSUE #2430: Payjoin receiver can contribute unconfirmed (0-conf) inputs
the payjoin candidate filter (`_filterAvailableUtxos`, `payjoin_repository_impl.dart`) removes only locked and non-Bitcoin UTXOs, never confirmation depth, so the receiver can contribute a 0-conf input.

this is a temporal fingerprint: an unconfirmed-ancestor spend is observable on-chain and partitions the payjoin's anonymity set by elimination (receivers that only contribute confirmed inputs are ruled out). A 0-conf parent can also be RBF'd out, invalidating the payjoin.

fix: exclude 0-conf from the payjoin candidates (`confirmations > 0`). the field already exists on `WalletUtxoModel`. Scoped to the payjoin receiver path; normal sends are unaffected.

ref: https://github.com/payjoin/rust-payjoin/issues/1597#issuecomment-4649767163

bc1cindy:
the payjoin candidate filter never checked confirmation depth, so the receiver could contribute a 0-conf input, a temporal fingerprint (an unconfirmed-ancestor spend is observable and partitions the anonymity set by elimination), and a 0-conf parent can be RBF'd out, invalidating the payjoin.

excludes 0-conf from the payjoin candidates (`confirmations > 0`). Receiver-scoped; normal sends unaffected.

ref: https://github.com/payjoin/rust-payjoin/issues/1597#issuecomment-4649767163

closes #2430

### DIFF
--- a/lib/core/payjoin/data/repository/payjoin_repository_impl.dart
+++ b/lib/core/payjoin/data/repository/payjoin_repository_impl.dart
@@ -446,6 +446,8 @@ class PayjoinRepositoryImpl implements PayjoinRepository {
   ) {
     return unspent
         .where((u) => !locked.any((l) => l.txId == u.txId && l.vout == u.vout))
+        // Skip 0-conf UTXOs: temporal-fingerprint leak + parent can be RBF'd out.
+        .where((u) => u.confirmations > 0)
         .whereType<BitcoinWalletUtxoModel>()
         .map((u) => PayjoinInputPairModel.fromWalletUtxoModel(u))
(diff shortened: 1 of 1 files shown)


BullishNode:
Concept nACK 

Unless my mind can be changed here: https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2430

PR #2436, opened 2026-07-10
fix(reproducibility): close build-reproducibility audit findings

ethicnology:
Closes findings from the 2026-07 reproducible-build audit. Each change removes a source of non-determinism from the release build, or hardens the verification tooling so a broken build fails **closed** (red) instead of silently passing.

## Build determinism

- **`android/app/build.gradle`** — disable two AGP-injected, source-independent metadata blobs:
  - `dependenciesInfo { includeInApk/​includeInBundle = false }` — the Play dependency-metadata snapshot (upload key / Play Console project). Invisible while release builds are unsigned, but becomes a non-reproducible blob the moment the production flavor is signed. *Tradeoff: also drops the metadata Play Console uses for SDK-issue warnings — intended.*
  - `vcsInfo { include = false }` — AGP's per-build git version-control snapshot. Currently falls back to a constant `NO_SUPPORTED_VCS_FOUND` in the container, but that's a fragile coincidence, not a guarantee.
- **`gradle-wrapper.properties`** — pin `distributionSha256Sum` for `gradle-8.13-all.zip` (verified against services.gradle.org).
- **`Containerfile.tools`** — pin Rust via a rustup shim that rewrites cargokit's hard-coded `rustup run stable` to the pinned `RUST_VERSION` (rustup ≥1.28 rejects `toolchain link stable`, so aliasing the reserved channel isn't possible). Pre-install bdk_dart's own `1.85.1` toolchain + Android targets and the two NDK/CMake versions AGP resolves, so no toolchain or SDK component is fetched from the network mid-build.
- **`.dockerignore`** — exclude everything not needed to build, with the debug + beta-upload keystores explicitly re-included so `build.gradle` can sign.

## Verification tooling (fail-closed)

- **`makefile` `verify-rustc-pins`** — greps the embedded `rustc version` out of every shipped Rust `.so` and compares it against the pinned toolchains running live inside `bull-app`, catching a regression of the shim above. Fails closed on: missing `strings`, empty toolchain lookup, stripped libs, or zero tracked libs. Loops **all** ABIs of a fat APK (not just the first). `BDK_RUST_VERSION` is a single make variable passed as `--build-arg`, so the makefile and Containerfile can't drift.
- **`reproducibility/compare_apk_entries.sh`** — the authoritative verdict: hashes every zip entry's raw bytes (excluding only legacy JAR signature files). Escapes shell-glob metacharacters in entry names before `unzip -p` (an entry named with `* ? [ ]` would otherwise extract a different pattern-matched entry and hash a real difference as `IDENTICAL`), and extracts to a temp file so a failed per-entry extraction aborts instead of hashing empty input (POSIX sh has no `pipefail`). Empty/unreadable archives exit 2, never a false `IDENTICAL`.
- **`reproducibility/verify_build.sh` / `test.sh`** — missing built splits are hard failures (were silently-skipped warnings); bidirectional split matching; distinguish compare exit 2 (unreadable APK) from exit 1 (real byte differences) so `RESULTS.md` doesn't present an infrastructure error as a diff. Dirty-tree guard (`--allow-dirty` to override).
- **`reproducibility/README.md`** — documents the pipeline, the shim, and the known gaps that need org/infra access (image archiving per release with a beta-secret caveat; mirroring the personal-account `bitbox-api-rs` fork; build provenance attestation).


ethicnology:
@Claude can you audit this PR ?

PR #2440, opened 2026-07-11
Fix/lwk concurrent wollet access v2

i5hi:


i5hi:
@claude please review - we had users on ios and android get the app startup error - reproduced the issue and fixed it for an android user via an update to this branch - we dont have an ios user with the issue to test nad prove the ios side - but its likely similar - lwk issue or the issue with the keychain unlock

i5hi:
@claude please review again - consider any regression risks and whether the pin migration for ios safe

PR #2441, opened 2026-07-12
[privacy] Implement Joinstr coinjoin (proof of concept)

kwsantiago:
This pull request implements a proof of concept for Joinstr, a coinjoin protocol coordinated over Nostr. Peers combine one input each into a single transaction that pays every peer an equal-denomination output back to their own wallet.

Read more: [joinstr.xyz](https://joinstr.xyz/)

Uses the [`joinstr_flutter`](https://github.com/rust-joinstr/joinstr) bindings. Behind superuser settings, testnet only, hot BIP84 wallets only.

TODO:

- [ ] Add screenshots
- [x] Tests

### DIFF
--- a/lib/features/joinstr/domain/joinstr.dart
+++ b/lib/features/joinstr/domain/joinstr.dart
@@ -10,6 +10,7 @@ enum JoinstrIssue {
   invalidElectrumUrl,
   invalidPoolConfig,
   invalidRelayUrl,
+  torUnavailable,
   poolNotFound,
   coinjoinFailed,
 }
--- a/lib/features/joinstr/domain/joinstr.dart
+++ b/lib/features/joinstr/domain/joinstr.dart
@@ -53,10 +53,8 @@ class JoinstrPool {
   });
 
   /// Seconds until the pool expires relative to [now], clamped at zero.
-  int secondsUntilExpiry(DateTime now) {
-    final remaining = expiresAtUnixSec - now.millisecondsSinceEpoch ~/ 1000;
-    return remaining > 0 ? remaining : 0;
-  }
+  int secondsUntilExpiry(DateTime now) =>
+      Joinstr.secondsUntilExpiry(expiresAtUnixSec, now);
 }
 
 abstract final class Joinstr {
@@ -115,6 +113,13 @@ abstract final class Joinstr {
       address: parsed.useSsl ? 'ssl://${parsed.host}' : parsed.host,
       port: parsed.port,
     );
+  }
+
+  /// Seconds until [expiresAtUnixSec] relative to [now], clamped at zero.
+  /// Shared by pools and rounds.
+  static int secondsUntilExpiry(int expiresAtUnixSec, DateTime now) {
+    final remaining = expiresAtUnixSec - now.millisecondsSinceEpoch ~/ 1000;
+    return remaining > 0 ? remaining : 0;
   }
 
   /// Formats a remaining duration for pool tiles, e.g. "1h 5m", "12m 30s",
--- a/lib/features/joinstr/domain/joinstr.dart
+++ b/lib/features/joinstr/domain/joinstr.dart
@@ -1,3 +1,4 @@
+import 'package:bb_mobile/core/electrum/domain/value_objects/electrum_url.dart';
 import 'package:bb_mobile/core/errors/bull_exception.dart';
 import 'package:bb_mobile/core/wallet/domain/entities/wallet.dart';
 
@@ -99,33 +100,21 @@ abstract final class Joinstr {
     return denom > 0 ? denom : null;
   }
 
-  /// Splits a stored electrum url into the address and port the bindings take.
+  /// Splits a stored electrum url into the address and port the bindings
+  /// take, via the electrum module's canonical [ElectrumUrl] parse.
   ///
   /// The `ssl://` prefix is deliberately preserved: joinstr only negotiates TLS
   /// when the address starts with it, so stripping the scheme would silently
   /// downgrade an SSL-only server such as `:50002` to plaintext.
   static ({String address, int port}) parseElectrumUrl(String url) {
-    final trimmed = url.trim();
-    final schemeEnd = trimmed.indexOf('://');
-    final scheme = schemeEnd == -1
-        ? ''
-        : trimmed.substring(0, schemeEnd).toLowerCase();
-    final hostPort = schemeEnd == -1
-        ? trimmed
-        : trimmed.substring(schemeEnd + 3);
-
-    final colon = hostPort.lastIndexOf(':');
-    if (colon <= 0 || colon == hostPort.length - 1) {
+    final parsed = ElectrumUrl.tryParse(url);
+    if (parsed == null) {
       throw JoinstrException(JoinstrIssue.invalidElectrumUrl, detail: url);
     }
-
-    final host = hostPort.substring(0, colon);
-    final port = int.tryParse(hostPort.substring(colon + 1));
-    if (port == null || port < 1 || port > 65535) {
-      throw JoinstrException(JoinstrIssue.invalidElectrumUrl, detail: url);
-    }
-
-    return (address: scheme == 'ssl' ? 'ssl://$host' : host, port: port);
+    return (
+      address: parsed.useSsl ? 'ssl://${parsed.host}' : parsed.host,
+      port: parsed.port,
+    );
   }
 
   /// Formats a remaining duration for pool tiles, e.g. "1h 5m", "12m 30s",
--- a/lib/features/joinstr/domain/joinstr.dart
+++ b/lib/features/joinstr/domain/joinstr.dart
@@ -1,0 +1,149 @@
+import 'package:bb_mobile/core/errors/bull_exception.dart';
+import 'package:bb_mobile/core/wallet/domain/entities/wallet.dart';
+
+enum JoinstrIssue {
+  bitcoinOnly,
+  mainnetNotSupported,
+  watchOnlyWallet,
+  unsupportedScriptType,
+  noEligibleCoin,
+  invalidElectrumUrl,
+  invalidPoolConfig,
+  poolNotFound,
+  coinjoinFailed,
+}
+
+class JoinstrException extends BullException {
+  final JoinstrIssue issue;
+  final int? denominationSat;
+  final String? detail;
+
+  JoinstrException(this.issue, {this.denominationSat, this.detail})
+    : super('Joinstr: ${issue.name}');
+}
+
+/// A coinjoin pool advertised on a nostr relay.
+class JoinstrPool {
+  final String id;
+
+  /// Canonical JSON, passed back to the bindings to join.
+  final String rawJson;
+  final int denominationSat;
+  final int peers;
+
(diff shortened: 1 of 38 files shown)


lib/features/joinstr/domain/usecases/list_joinstr_pools_usecase.dart
```diff
@@ -0,0 +1,23 @@
...
+    String? relay,
+    Duration back = const Duration(days: 1),
+    Duration wait = const Duration(seconds: 5),
+  }) async {
+    final pools = await _datasource.listPools(
+      relay: relay ?? ApiServiceConstants.defaultNostrRelayUrl,
+      back: back,
+      wait: wait,
+    );
+    pools.sort((a, b) => a.denominationSat.compareTo(b.denominationSat));
```
1440000bytes: `list_pools `returns everything advertised in the last 24h, so most of these pools are already expired. They still render with a live join button, and joining one publishes a signature over the input for a pool that can only time out. Filter before sorting:

```dart
final now = DateTime.now();
pools.removeWhere((p) => p.secondsUntilExpiry(now) == 0);
pools.sort((a, b) => a.denominationSat.compareTo(b.denominationSat));
```

lib/features/joinstr/joinstr_locator.dart
```diff
@@ -0,0 +1,51 @@
...
+      ),
+    );
+    locator.registerFactory<InitiateJoinstrPoolUsecase>(
+      () => InitiateJoinstrPoolUsecase(
+        datasource: locator<JoinstrDatasource>(),
+        resolvePeerContextUsecase: locator<ResolveJoinstrPeerContextUsecase>(),
+      ),
+    );
+
+    locator.registerFactory<JoinstrCubit>(
```
1440000bytes: `joinCoinjoin `and `initiateCoinjoin `block until broadcast or pool timeout, but this cubit is a factory and `BlocProvider `closes it when the user leaves the screen. The ffi call keeps running, the txid lands in a closed cubit and is lost, and re-entering creates a fresh cubit with isRunning false, so the same coin can be offered to a second pool while the first round is still signing with it.

Register it as a singleton and provide it with `BlocProvider.value` so navigation cannot kill a running round:

```dart
locator.registerLazySingleton<JoinstrCubit>(
```

and in settings_router.dart:

```dart
builder: (context, state) => BlocProvider.value(
  value: locator<JoinstrCubit>()..load(),
  child: const JoinstrScreen(),
),
```

load() should also return early while a round is waiting.

lib/features/joinstr/ui/screens/joinstr_screen.dart
```diff
@@ -0,0 +1,246 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final cubit = context.read<JoinstrCubit>();
+    final state = context.watch<JoinstrCubit>().state;
+
+    return Column(
+      crossAxisAlignment: CrossAxisAlignment.stretch,
+      children: [
+        TextFormField(
+          initialValue: state.denominationSat,
```
1440000bytes: `denominationChanged `strips non digits in the cubit but the field is only bound through `initialValue`, so typing 1a2 displays 1a2 while state holds 12. What the user sees is not what gets submitted. Filter at the field so display and state cannot diverge:

```dart
inputFormatters: [FilteringTextInputFormatter.digitsOnly],
```

Same for the peers and fee rate fields.

lib/features/joinstr/domain/usecases/resolve_joinstr_peer_context_usecase.dart
```diff
@@ -0,0 +1,72 @@
...
+        JoinstrIssue.invalidElectrumUrl,
+        detail: failure.logMessage,
+      ),
+    );
+
+    // A fresh address each round: joinstr's own address-reuse check is gated on
+    // an electrum client the peer flow never supplies, so it never runs.
+    final address = await _getReceiveAddressUsecase.execute(
+      walletId: wallet.id,
+      generateNew: true,
```
1440000bytes: `generateNew`: `true `consumes a fresh index on every attempt, including failures like no eligible coin or a pool that times out before filling. A few failed joins in a row walk the receive chain toward the gap limit with addresses that never see funds. Select the eligible coin first and only derive the address once the round is actually going ahead, or reuse the last unused address until a round broadcasts.

lib/features/joinstr/domain/joinstr.dart
```diff
@@ -0,0 +1,149 @@
...
+  /// A joinstr input must satisfy
+  /// `denomination + 500 <= value <= denomination + 5000`, enforced by every
+  /// other peer in `CoinJoin::add_input`. A coin outside this window is
+  /// rejected after we have already published a signature over it, so the
+  /// window is checked here before anything is signed.
+  static const int minInputSurplusSat = 500;
+  static const int maxInputSurplusSat = 5000;
+
+  /// Derivation indexes scanned on each branch when listing coins.
+  static const int scanDepth = 100;
```
1440000bytes, line 70: A wallet with more than 100 used indexes on either branch will report no eligible coin even when one exists past the cutoff, and nothing tells the user why. Derive the scan range from the wallet's next derivation index instead of a constant, or at least mention the scanned range in the `noEligibleCoin `error.

1440000bytes:
Thanks for implementing joinstr in bull wallet. I have made the UI/UX here consistent with joinstr-kmp and joinstr-floresta in this branch: https://github.com/1440000bytes/bullbitcoin-mobile/tree/joinstr-pr2441

It also includes some tests and bug fixes. You can find the APK for it here: https://github.com/1440000bytes/bullbitcoin-mobile/actions/runs/29418299164

lib/features/joinstr/domain/usecases/list_joinstr_pools_usecase.dart
kwsantiago: Fixed in the integrated branch (aa1808f39): `list_joinstr_pools_usecase` now filters `pools.where((p) => p.secondsUntilExpiry(now) > 0)` before sorting, so expired pools no longer render with a live Join button. Covered by a test.

lib/features/joinstr/joinstr_locator.dart
kwsantiago: Fixed: JoinstrCubit is now `registerLazySingleton`, provided via `BlocProvider.value` in settings_router, and `load()` early-returns while a round is waiting. A round that blocks in the bindings survives navigation, and the txid is persisted in the usecase before it returns, so it can't be lost to a closed cubit. `isRunning` gating also prevents offering the same coin to a second pool.

lib/features/joinstr/ui/screens/joinstr_screen.dart
kwsantiago: Fixed via a `_FilteredTextField` that syncs its controller back to the cubit's filtered value in didUpdateWidget, so the displayed and submitted values can no longer diverge.

lib/features/joinstr/domain/usecases/resolve_joinstr_peer_context_usecase.dart
kwsantiago: Acknowledged as a fast-follow. Deliberately not taking the one-line `generateNew: false` here: this flow's own comment notes joinstr's address-reuse check never runs, so reusing the last address would send a coinjoin output to a reused address, a real on-chain privacy leak that links rounds. The right fix (select the eligible coin before deriving the address, or reveal only when a round actually proceeds) needs restructuring across the usecase/datasource boundary; tracking it rather than rushing a privacy-sensitive change.

lib/features/joinstr/domain/joinstr.dart
kwsantiago: Acknowledged as a fast-follow: `noEligibleCoin` should surface the scanned range (or derive scanDepth from the wallet's next derivation index) so a coin past index 100 isn't silently missed. Tracking it.

lib/features/joinstr/domain/usecases/resolve_joinstr_peer_context_usecase.dart
kwsantiago: Fixed in ce5ca3c. Instead of `generateNew: true` on every attempt, resolvePeerContext now reserves one fresh address in the store and reuses it across retries; the initiate/join usecases clear the reservation on a successful broadcast. So a failed round (no-eligible-coin or pool-timeout) reuses its reserved-but-unfunded address rather than burning a new one, and each completed coinjoin still gets a distinct address, so no on-chain reuse.

lib/features/joinstr/domain/joinstr.dart
kwsantiago: Fixed in 6b38712. The noEligibleCoin error now names the scanned range: the l10n string gained a {scanned} placeholder and the screen passes Joinstr.scanDepth, so it reads '...in the first 100 addresses' instead of silently cutting off. (Deriving scanDepth from the wallet's next index would need index plumbing the wallet entity doesn't expose cheaply; went with the messaging fix you offered as the minimum.)

1440000bytes:
Thanks for addressing the previous review comments. I have added tor, fixed UI/UX and moved joinstr to home screen in these commits:

https://github.com/1440000bytes/bullbitcoin-mobile/commit/31de2591be0b1ab4a7cd9d56662b23040128fcbc
https://github.com/1440000bytes/bullbitcoin-mobile/commit/b4ac7332014286ff81a5bb6dc8be34e650df8371
https://github.com/1440000bytes/bullbitcoin-mobile/commit/0b393bf244aac4c92877771d0e3b816d2aec27e9
https://github.com/1440000bytes/bullbitcoin-mobile/commit/2bdab007577cf9736187c3599a0856bd3ff855f6
https://github.com/1440000bytes/bullbitcoin-mobile/commit/d0955f5380c3dd764dbce1fc58e088e22d510fd4

Pull request to add socks5 proxy in rust-joinstr: https://github.com/rust-joinstr/joinstr/pull/36
APK: https://github.com/1440000bytes/bullbitcoin-mobile/actions/runs/29457022360

kwsantiago:
Merged your 5 commits (Tor, UI/UX, home-screen tile, tests) into the branch. Two deltas on top:

- Kept two earlier bug fixes that weren't on your branch: reserve the output address across retries (avoids walking the gap limit on failed rounds), and surface the scanned range in the no-eligible-coin error.
- Repinned joinstr_flutter `b75eea0` -> `d1bb9f8` (rust-joinstr master tip), which adds the rust-joinstr/joinstr#39 hardening: reject SOCKS5 no-auth to preserve circuit isolation, and carry the proxy in resumed State.

analyze clean, joinstr tests green.

1440000bytes:
Fixed multiple bugs, improve UI/UX and tested it with a coinjoin transaction on testnet3: https://mempool.space/testnet/tx/865f5b7eac9059b7339997ccaef712c5bd393f41b1f9e8822b95d0ec436e0e6d

Commits: https://github.com/1440000bytes/bullbitcoin-mobile/commits/joinstr-pr2441/
APK: https://github.com/1440000bytes/bullbitcoin-mobile/actions/runs/29862972366

Changes made in rust-joinstr: https://github.com/rust-joinstr/joinstr/pull/40

<details>

<summary>Screenshots</summary>








</details>

pievalentin:
Very cool stuff!

PR #2443, opened 2026-07-13
feat(payjoin): upgrade to official `payjoin` pub.dev bindings, drop isolate architecture

### LINKED ISSUE #1446: Upgrade to Payjoin 1.0.0
Still in pre-release but tracking for later https://github.com/payjoin/rust-payjoin/releases/tag/payjoin-1.0.0-rc.1

ethicnology:
Migrates the payjoin integration from our custom `payjoin_flutter` fork (`SatoshiPortal/payjoin-dart`, git dependency) to the official `payjoin` package on pub.dev (`payjoin.org`, uniffi-dart bindings, `^0.1.1`), and rewrites `PdkPayjoinDatasource` around its new, synchronous typestate API.

This upgrade was worked on over several months (PR #2041, opened 2026-04-22). Thanks to everyone who helped get it across the line:

- **spacebear** — authored the upgrade and the isolate-removal rewrite
- **Dan Gould** — upstream payjoin/PDK guidance and review
- **Ben Allen** — review

## What changed

- **Dependency**: `payjoin_flutter` (git ref) → `payjoin: ^0.1.1` (hosted, verified publisher). Removes the platform plugin wiring (`ios/Podfile.lock`, `linux/flutter/generated_plugins.cmake`) that came with the old Flutter plugin fork, and the now-unneeded `PConfig.initializeApp()` call in `main.dart`.
- **Dropped the isolate-based background architecture.** Receiver/sender polling used to run in two long-lived `Isolate`s spawned lazily and communicating over `SendPort`/`ReceivePort`. The new PDK exposes a synchronous typestate machine with an in-memory JSON session persister, so polling now runs as plain `Timer.periodic` in the main isolate (`PdkPayjoinDatasource`). Net effect: much less code, no isolate bring-up latency, no serialization boundary between isolate and main.
- **`BdkWalletDatasource`**: added `createIsMineChecker` / `createPsbtSigner`, which preload a bdk wallet once and return plain synchronous closures, matching the new PDK callback signatures (`bool Function(Uint8List)`, `String Function(String)` instead of `FutureOr<...> Function`). Also hardens Electrum client creation against a `rustls` `CryptoProvider` install race that could otherwise fail one of two concurrently-syncing wallets.
- **Re-enabled `integration_test/payjoin_test.dart`**, which had been fully commented out on `develop`. Rewrote it against the new API and added draining of leftover in-flight payjoin state in `setUpAll` to stop previous (possibly crashed) runs from starving the test wallets.

## Known follow-ups (not addressed in this PR)

- The payjoin typestate walk (`processReceiveSession` and its `_check*`/`_commit*`/`_finalize*` chain) has no unit test coverage — only the integration test, which isn't run in CI and needs funded testnet wallets. Worth a follow-up with fakes around the repository/datasource seam.
- `integration_test/payjoin_test.dart` still has three empty stub tests (resume-after-restart, insufficient-funds paths) and a commented-out multi-payjoin test (`TODO: Fix this test`) — pre-existing gaps, now visible again now that the file is no longer fully disabled.

### DIFF
--- a/integration_test/payjoin_test.dart
+++ b/integration_test/payjoin_test.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:bb_mobile/core/fees/domain/fees_entity.dart';
+import 'package:bb_mobile/core/payjoin/data/datasources/local_payjoin_datasource.dart';
 import 'package:bb_mobile/core/payjoin/domain/entity/payjoin.dart';
 import 'package:bb_mobile/core/payjoin/domain/repositories/payjoin_repository.dart';
 import 'package:bb_mobile/core/payjoin/domain/usecases/receive_with_payjoin_usecase.dart';
@@ -33,6 +34,7 @@ Future<void> main({bool isInitialized = false}) async {
   final addressRepository = locator<WalletAddressRepository>();
   final utxoRepository = locator<WalletUtxoRepository>();
   final payjoinRepository = locator<PayjoinRepository>();
+  final localPayjoinDatasource = locator<LocalPayjoinDatasource>();
   final receiveWithPayjoinUsecase = locator<ReceiveWithPayjoinUsecase>();
   final sendWithPayjoinUsecase = locator<SendWithPayjoinUsecase>();
   final prepareBitcoinSendUsecase = locator<PrepareBitcoinSendUsecase>();
@@ -49,6 +51,34 @@ Future<void> main({bool isInitialized = false}) async {
 
   setUpAll(() async {
     await locator<SetEnvironmentUsecase>().execute(Environment.testnet);
+
+    // Drain any persisted payjoin state so the test starts clean. Ongoing
+    // payjoins left behind by a previous (possibly crashed) run keep their
+    // inputs frozen via getUtxosFrozenByOngoingPayjoins(), which would starve
+    // the sender wallet. _resumePayjoins in PayjoinRepositoryImpl's
+    // constructor runs unawaited and writes its own updates concurrently, so
+    // we expire + recheck until the ongoing set stays empty for several polls.
+    const pollInterval = Duration(milliseconds: 500);
+    const requiredStableChecks = 3;
+    const maxIterations = 40;
+    var stableChecks = 0;
+    for (var i = 0; i < maxIterations; i++) {
+      final ongoing = await localPayjoinDatasource.fetchAll(
+        onlyUnfinished: true,
+      );
+      if (ongoing.isEmpty) {
+        stableChecks++;
+        if (stableChecks >= requiredStableChecks) break;
+      } else {
+        stableChecks = 0;
+        for (final payjoin in ongoing) {
+          await localPayjoinDatasource.update(
+            payjoin.copyWith(isExpired: true),
+          );
+        }
+      }
+      await Future.delayed(pollInterval);
+    }
 
     final receiverSeed = await seedRepository.createFromMnemonic(
       mnemonicWords: receiverMnemonic.split(' '),
--- a/integration_test/payjoin_test.dart
+++ b/integration_test/payjoin_test.dart
@@ -5,7 +5,7 @@ import 'package:bb_mobile/core/payjoin/domain/entity/payjoin.dart';
 import 'package:bb_mobile/core/payjoin/domain/repositories/payjoin_repository.dart';
 import 'package:bb_mobile/core/payjoin/domain/usecases/receive_with_payjoin_usecase.dart';
 import 'package:bb_mobile/core/payjoin/domain/usecases/send_with_payjoin_usecase.dart';
-import 'package:bb_mobile/core/seed/data/models/seed_model.dart';
+import 'package:bb_mobile/core/seed/data/repository/seed_repository.dart';
 import 'package:bb_mobile/core/settings/domain/settings_entity.dart';
 import 'package:bb_mobile/core/utils/constants.dart';
 import 'package:bb_mobile/core/wallet/data/repositories/wallet_address_repository.dart';
@@ -16,7 +16,6 @@ import 'package:bb_mobile/features/send/domain/usecases/prepare_bitcoin_send_use
 import 'package:bb_mobile/features/settings/domain/usecases/set_environment_usecase.dart';
 import 'package:bb_mobile/locator.dart';
 import 'package:bb_mobile/main.dart';
-import 'dart:io' show Platform;
 
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart' show TestWidgetsFlutterBinding;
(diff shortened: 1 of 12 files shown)


ethicnology:
Following the review feedback, this branch now includes a hardening pass on top of the original upgrade. All fixes were verified end-to-end with real payjoin roundtrips on testnet
### Review fixes

- **`5b79418` fix(payjoin): guard polls against overlap, stranding and missed expiry** — addresses the 🔴 race condition: `Timer.periodic` doesn't await its async callback, so a slow OHTTP relay could make ticks overlap, double-process the same session, and end up broadcasting the original transaction — silently cancelling the payjoin. Added per-session in-flight guards, moved `timer.cancel()` after all fallible work (a throw now leaves the timer armed so the next tick retries instead of stranding the session), re-check `timer.isActive` before emitting, and added a local expiry backstop (`createdAt + expireAfterSec`) so polling is bounded even if the PDK never surfaces an "expired" error. Also bounds each poll with Dio connect/receive timeouts.
- **`135dd68` fix(wallet): restrict payjoin psbt signer to standard sighashes** — `createPsbtSigner` now uses the same sign options as `signPsbt` (`allowAllSighashes: false`) since it signs the receiver's contribution to an externally-supplied transaction, and logs when the signed PSBT isn't finalized instead of silently returning it.
- **`4d7aad4` fix(wallet): extend CryptoProvider race guard to dry scans** — the rustls CryptoProvider install-race retry now covers `_performDryScan` too, via a shared `_createElectrumClient` helper.
- **`fe8b712` refactor(payjoin): surface underlying errors in datasource logs** — relay fetch failures, the PDK's input-selection rejection reason, and corrupt persisted event logs are no longer swallowed silently. This is what made the root cause below diagnosable.

### Tests

- **`db48dea` test(payjoin): unit-test session persisters and relay fallback** — first unit tests for the payjoin module (15 tests, fully offline): JSON round-trip/corruption handling for both session persisters, and the multi-relay fallback loop via an injected `OhttpKeysFetcher` seam.
- **`1215ed5` test(payjoin): keep integration wallets payjoin-capable across runs** — the integration test used to degrade the shared testnet wallets with every run (accumulating dust and mismatched UTXOs) until payjoins permanently failed. Root cause is an upstream rust-payjoin bug: `WantsInputs::try_preserving_privacy`'s documented fallback ("a simple consolidation is otherwise chosen if available") never runs, because `avoid_uih` drains the shared `Peekable` candidate iterator before returning its error, so `select_first_candidate` always sees an empty iterator and fails with "No candidates available for selection" (payjoin crate 0.23.0, `src/receive/v1/mod.rs` — reproduced in isolation). The test now sweeps each wallet to a single UTXO before the roundtrip, giving the selection heuristic one unambiguous candidate and making the test self-sustaining: verified with consecutive successful payjoin roundtrips on a physical device. Mnemonics can now also be passed via `--dart-define` for on-device runs (`Platform.environment` doesn't reach the app process on a physical device); CI's env-var path is unchanged.

ethicnology:
@Claude you can review and audit the whole branch against develop

spacebear21:
https://pub.dev/packages/payjoin/versions/0.1.2 is now published! Importantly, it uses the latest version of uniffi-dart, which uses uniffi 0.31.2 ensuring reproducible builds.

ethicnology:
- **`dc2cca5` build(deps): bump payjoin to 0.1.2** — the 0.1.2 bindings use the latest uniffi-dart (uniffi 0.31.2), which restores reproducible native builds.
- **`3ea1abb` fix(payjoin,wallet): address review findings** — `_buildInputPair` now throws on a missing input value instead of defaulting to zero (which would sign over the wrong segwit-committed amount and produce an invalid signature); `_decodeEvents` builds the list eagerly so a list with non-string entries is caught as a corrupt log instead of throwing later on every poll tick; `_resumePayjoins` emits the updated (expired) model rather than the stale one; dropped the misleading "not finalized" log on the receiver's partial signing, the full-PSBT info log, and the unused `OhttpRelaysUnavailableException`; surfaced terminal all-relays-failed errors to production logs; removed a stray `required` field from a freezed factory.
- **`49c7c0f` test(payjoin): cover corrupt-log and relay-timeout edge cases** — adds a list-of-non-strings decode case and covers the `postBytes` relay choke point (a Dio receive-timeout propagates unwrapped so the relay loops can fall back; success returns the response bytes).
- **`d42d1c2` test(payjoin): correct multi-payjoin integration test timeout unit** — a pre-existing `Duration(minutes:)`/`seconds:` typo (30 minutes instead of 30 seconds).

ethicnology:
@Claude last round of audit

ethicnology (commented):
@i5hi `payjoin-upgrade` is good to merge. Several non-blocking hardening items can be addressed in a follow-up branch, kept out of here to keep the scope tight:

- **Shorter session lifetime (anti-probing).** Expiry drops 24h → 5min: a payjoin is a synchronous exchange, so a long window only lengthens the period a probing sender can hold an exposed UTXO and cheaply replace the original tx (BIP78).
- **Exposed-UTXO reuse (anti-probing).** Once a proposal reveals a receiver UTXO, a later attempt should re-contribute that same already-burned coin rather than expose a fresh one. Follow-up labels contributed UTXOs (`_labelExposedUtxos`) and prefers them in selection (`filterAvailableUtxos`), closing a probing vector.
- **Don't race our own in-flight payjoin on expiry.** `_processExpiredPayjoin` broadcasts the original only when `proposalPsbt == null`; once a proposal has gone out the sender owns broadcasting the payjoin (same inputs), so broadcasting the original would just race it.
- **Corrupt persisted log → replay loop.** `_decodeEvents` resets silently to `[]`, resurrecting the session empty and re-polling until expiry. 
- `CorruptPayjoinSessionException` that retires the session.
- **Datasource lifecycle leak.** No `dispose()`, so a never-resolved session can leave `Timer.periodic` + `StreamController`s alive. Follow-up adds an idempotent `dispose()`, delegated from the repository.
- **`// Todo: listen for the broadcast of the transaction`** (`_resumePayjoins`). Follow-up implements watch-for-broadcast so the receiver completes/labels the session once the tx lands on-chain.
- **Transient relay blip.** A momentary all-relays failure can irrevocably cancel a payjoin; follow-up wraps proposal posting in a bounded retry.
- **Signer robustness.** `createPsbtSigner` throws `FailedToSignPsbtException` when the wallet matches no proposal input (fails at the boundary, not downstream); CryptoProvider install retry is bounded with backoff.
- **User disclosure.** Receive screen now warns that payjoin reveals one of your UTXOs to the sender.

DanGould:
heads up: one database breaking change in https://github.com/payjoin/rust-payjoin/pull/1747 coming before the change. This freezes the persisted data so it shouldn't be a problem when we introduce planned updates. This PR doesn't need to change with code, just gotta update the dep. 

Your "Follow-up implements watch-for-broadcast so the receiver completes/labels the session once the tx lands on-chain." suggestion is something we're actively trying to accomodate in our own library with 1.1 [here](https://github.com/payjoin/rust-payjoin/pull/1692). I don't recommend building it on the current `check_for_transaction` API. The new API gives you:

`classify(chain_view)` function => Pending | Detected { outcome: Cooperative|Fallback|Other, confirmations }` that maps directly onto the sending / confirming / complete / sent-as-regular-transfer / needs-review states, and `check_for_transaction` gets              deprecated at the same time. SatoshiPortal/payjoin can use the same exact approach for receiver classification when this lands

i5hi (approved):

PR #2453, opened 2026-07-15
feat(backup): recurring backup health reminders and Backup Settings redesign

BullishNode:
## Summary

Recurring backup-health reminders, plus the Backup Settings screen redesign they belong to. Rebased onto current `develop`.

The whole feature answers one question — **"if this phone vanished right now, could you get your money back?"** — and gives each state exactly one verb: nothing backed up ⇒ back up; encrypted vault only ⇒ add a physical backup; physical backup done ⇒ occasionally confirm it still works.

## Reminder engine

State per wallet is three fields: `last_tested`, `last_acknowledged`, `crossed_ten_million_sats`.

| Posture | Reminder |
|---|---|
| Nothing backed up | **No popup** — owned by the existing every-launch warning and the screen hero |
| Encrypted vault only | Every **90 days**: add a physical backup |
| Physical backup done (± vault) | Every **365 days** since last test: test your backup |
| Any posture, first time balance ≥ 10,000,000 sats | **One popup, once per wallet lifetime** |

- Fires on app open, wallet home only; silent during onboarding.
- Dismiss snoozes a full cycle and never writes a tested timestamp — the screen keeps telling the truth ("Last tested 14 months ago") while the popup is quiet.
- `last_tested` resets only on completing a verification; creating a backup counts as day-zero tested (creation requires a test).
- One popup = one question = one primary action.
- Balance = total of wallets whose **keys are on the device** (mainnet, locally signing, not hardware) + Ark. Read lazily at evaluation time; never watched.

### Simplified relative to the first version of this PR
Deleted: balance tiers and their persisted state, tier-driven escalation, the pending-action bookkeeping, and the both-verified "review your backups" nag (it had no real action). There is deliberately **no PIN / vault-freshness reminder**: a vault-only user's real exposure is having a single recovery path that depends on someone else, and the fix for that is a physical backup — not testing a PIN.

### Behaviour fix
The schedule anchor was `latest(vault, physical)`, so a user with a fresh vault and a two-year-old physical test was never reminded. The anchor now follows the thing being urged. Regression-tested.

## Backup Settings screen

- **Status rows** — Physical Backup and Encrypted Vault, tested/not tested, with "Last tested …" underneath.
- **Hero** — one card naming the single most useful action, or nothing at all when the rows already say everything: urgent *Back up your wallet* when nothing is backed up, *Add a physical backup* for vault-only, *Test your backup* when a physical test is stale, and no hero when it is fresh.
- **Start backup is a button**, not a menu row — creating a backup is an action, not a setting — filled in primary red, present in every state except the one where the hero already offers it.
- **Menu** — "Recoverbull" is now **"Encrypted vault settings"**, matching the status row's vocabulary.
- The encrypted vault option card now also carries a **"Uses Tor"** tag, since reaching the key server over Tor is worth knowing before choosing that option.

## Encrypted vault flow

- **The permission gate is gone** (−763 lines). It was not an OS permission and gated nothing: tapping Continue wrote a local boolean whose only effect was whether the screen showed again. A user who has chosen the encrypted vault does not need a consent screen for the thing they just chose.
- **The custom-server warning moved to the moment of choice.** Instead of disclosing the server to every user on every vault flow, saving a *custom* key server URL now requires confirming a warning. Reverting to the default needs no confirmation; URL validation runs first.
- User-facing "Recoverbull" is retired as a **feature** name throughout, while `recoverbull.com` (a real domain) and "RecoverBull Server" (the server's actual product name, which a superuser configuring a URL needs) are kept.

## Validation

- `flutter analyze --fatal-warnings --fatal-infos`: no issues
- repo-wide `dart format --set-exit-if-changed`: 0 changed
- `dart fix --dry-run`: nothing to fix
- ARB validation: 27/27 locales
- Full suite: **860 passed**, plus `bull_ui` 17 and `bull_ui_catalogue` 1

## Notes for the reviewer

- **Please squash on merge.** The first commit briefly adds ~50 lines of fork-specific feature graph to `FEATURES.md` which the second removes; the final tree is clean, but squashing keeps that out of upstream history entirely.
- `isPermissionGranted` is now dead weight: a `NOT NULL` column baked into drift schemas v10–v14, written but never read. Dropping it needs a schema bump and migration on live data, so it is deliberately left for a future schema change rather than done here.
- Retired strings were deleted from all locales and re-added English-only, so a stale translation cannot resurface the old wording; strings whose wording did not change kept their existing translations.


BullishNode:
@thibistaken experimental idea, but thouroughly reviewed by AI

BullishNode:
@ethicnology I have removed forcing the user to "accept" the default recoverbull server but I've added a "uses Tor" label on the encrypted backup description card. Also, if user adds custom recoverbull server, there is still a little warning.

BullishNode:
Screenshot_1785195196
Screenshot_1785195156
Screenshot_1785195138
Screenshot_1785195122
Screenshot_1785193839

some screenshots

i5hi:
@claude please add your review

i5hi:
### Second review pass (Claude, multi-angle deep run — posted by @i5hi)

This complements the automated review above rather than repeating it: everything below is **new**. Agreements with the earlier review are listed at the end. Same overall verdict — well-built PR, strong domain tests, honest PR description (every claim in it verified against the code) — but the popup *lifecycle* around the evaluate/acknowledge loop has real gaps.

#### Findings (most severe first)

**1. "Not now" cannot silence a future-dated anchor — the blocking popup refires after every wallet sync.**
`lib/features/backup_settings/domain/usecases/evaluate_backup_health_reminder_usecase.dart:89-97` + `domain/backup_health_reminder.dart:50-52`
`isBackupReminderDue` returns due when `anchor.isAfter(now)` (the fail-toward-reminding rule the earlier review praised), and the anchor is `latest(urgedTestedAt, lastAcknowledgedAt)`. A `latestPhysicalBackup` stamped while the device clock was fast (then corrected) outvotes any acknowledge forever: ack writes `now`, `_latest` still picks the future tested date, still due. Every `WalletBloc` emit re-evaluates via `didUpdateWidget`, so the `canPop:false` modal reappears seconds after each "Not now" until real time passes the stored timestamp. Fail-toward-reminding is right; ack-can't-snooze is not — consider `anchor = max(anchor, ack)` clamped to `now`, or treating a future anchor as `now` at evaluation time.

**2. The 10M-sats milestone popup flaps around the threshold.**
`lib/features/backup_settings/presentation/cubit/backup_health_reminder_cubit.dart:51-57`
The latch persists only on user interaction, and a re-evaluation that lands below threshold emits `BackupHealthReminderHidden` unconditionally — yanking a visible popup mid-read. Concrete: 9,999,000 confirmed + 2,000 unconfirmed → popup; the unconfirmed tx is RBF-replaced → next sync hides it; re-broadcast → it's back. Once shown, either latch on display or don't withdraw a visible one-time notice on a balance dip.

**3. Vault-only users get two contradictory blocking popups in one launch.**
`lib/features/recoverbull/domain/complete_encrypted_vault_backup_usecase.dart` (called with only `defaultWallet.id`) + `wallet_state.hasNoBackup()`
Vault creation flags only the default **bitcoin** wallet; the liquid default keeps both flags false, so `hasNoBackup()` stays true and the every-launch "no backup" modal still shows (pre-existing quirk — only the vault *test* flow flags both defaults). Dismissing it then lets the new 90-day vault-only reminder fire immediately after: "you have no backup" followed by "add a physical backup to your vault", breaking this PR's own one-popup-one-question rule.

**4. Upgrade path: a custom key server saved before this PR never re-confirms.**
`lib/features/recoverbull/ui/pages/settings_page.dart:64-72`
The old gate re-fired on the next vault flow after *every* URL change (`store()` reset `isPermissionGranted` on save — it still writes it, nothing reads it). The new warning fires only for `entered != default && entered != _originalUrl`, i.e. new saves. A device upgrading with a pre-existing custom `.onion` sends its next vault key there with zero disclosure anywhere in the flow. New saves are airtight (single caller, validation first, dismiss stores nothing) — this is purely the migration case, and it's privacy-sensitive enough to deserve an explicit "intended" or a one-time re-confirmation for already-custom URLs.

**5. `fetch()` converts IO failures to `Ok(blank record)` — and acknowledge writes the blank back.**
`lib/features/backup_settings/data/shared_preferences_backup_health_reminder_repository.dart:35-61`
Corrupt JSON → default record is a fine read-time tradeoff. But acknowledge is fetch → copyWith → save, so one transient `SharedPreferences` read failure during an ack durably persists `crossedTenMillionSats: false` over a latched `true` — breaking the once-per-lifetime invariant the tests pin. Distinguish decode failure (default) from IO failure (`Err`); acknowledge's `Err` branch is currently dead code because fetch never errors.

**6. Usecase calls usecase in onboarding.**
`lib/features/onboarding/domain/usecases/create_onboarding_wallets_usecase.dart:16-27`
`CreateOnboardingWalletsUsecase` wraps core's `CreateDefaultWalletsUsecase` — orchestration belongs in the bloc, and the sanctioned wrap target is a repo/facade (the sibling `CompletePhysicalBackupVerificationUsecase` wrapping `TestWalletBackupFacade` in this same PR does it right).

**7. The due-ness rule executes inside a widget `build()`.**
`lib/features/backup_settings/ui/screens/backup_settings_screen.dart:126-175`
`_hero` derives the posture and calls `isBackupReminderDue(now: DateTime.now(), ...)` in UI, bypassing cubit and usecase — the 90/365 policy now runs in two layers that can drift, and the wall-clock read is why the widget tests have to keep wide margins from the boundaries. Belongs on the state via the cubit. (The earlier review noted the `DateTime.now()` asymmetry as fine; the layering duplication is the sharper cost.)

**8. The blocking-overlay shell is now copy-pasted three times** (nit).
`backup_health_reminder_overlay.dart:100-125` duplicates the `Positioned.fill > PopScope > Material > ... > rounded-top sheet` chrome of `backup_warning_overlay.dart:50-77` and the legacy-storage overlay byte-for-byte. One extracted shell in `core/widgets` stops them drifting.

#### Agreements with the earlier review
The events-dispatched-in-`build()` finding (its #1), the evaluate-stomps-in-flight-save race (#2, same mechanism as this list's popup lifecycle theme), the backup-succeeded-but-reported-failed path (#3), the verify-mnemonic retry dead-end (#4), and the URL-normalization fragility (#5) all verified as real here too — no need to restate them.

#### Verified clean (so nobody re-checks)
Schedule tests would genuinely fail on the pre-fix anchor code (checked against the PR's first commit); no drift schema change needed and none made; all 27 locale files parse, zero references to the 9 deleted keys anywhere, all 26 new keys have descriptions; `isPermissionGranted` has zero readers in `lib/` exactly as the body says; all router references to moved/deleted screens resolve.

Issue #2416, opened 2026-07-07
payjoin improments


BullishNode:
1. On tx confirmation screen, display that payjoin will be attempted 
2. Combine the Payjoin in progress spinner with the screen that asks if user wants to send without payjoin
3. Show timer for when fallback transaction will be broadcast (if we know it?)

ethicnology:
delivered in 6.13.0

PR #2461, opened 2026-07-16
 feat(payjoin): upgrade, settings, anti-probing hardening, and status/reliability fixes

### LINKED ISSUE #782: Allow user to view payjoin settings
App settings should have a section for payjoin settings, where the user can view (and potentially edit) payjoin-related stuff. Here are some items it can house:

- Payjoin relay/directory information (in the future users might prefer to use a different relay that they like or trust)

- Timeout duration (how long to wait for receiver action before abandoning payjoin and just broadcasting it)
This might be a big one for UX since it affects total transaction time (payjoin-duration + on-chain confirmation duration)

- fee-related payjoin parameter settings (happy to chat and see which if any parameters should be user-customizable)

### LINKED ISSUE #947: Better Payjoin Status
Report from testing:

When a payjoin transaction has been accepted it is written "proposed" in the Payjoin status field. It’s a bit misleading IMO. Shouldn’t we write "accepted" instead?

Image

We should have more meaningful and simple to undertand statuses for payjoin

### LINKED ISSUE #1446: Upgrade to Payjoin 1.0.0

### LINKED ISSUE #1584: Default payjoin settings
Turned on by default, the "default payjoin" toggle in Bitcoin settings will make it so that the wallet both generates a payjoin endpoint for recieve and will attempt to pay payjoin endpoints. If turned off, no payjoin.

ethicnology:
Closes:
- https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1446
- https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/947
- https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/782
- https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/1584
- https://github.com/SatoshiPortal/bullbitcoin-mobile/issues/2416

Non-exhaustive summary:
- https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2041 thanks to @spacebear21 @DanGould and @benalleng
- Migrated off the unmaintained `payjoin_flutter` fork onto the official `payjoin` package (rust-payjoin via uniffi), currently pinned to 1.0.0-rc.4 bindings.
- Configurable minimum receive amount (default 10,000 sat) below which a payjoin is declined and the payment broadcasts normally — raises the cost of BIP78 UTXO probing.
- Configurable session expiry defaulted to 1 minute.
- Enable/disable payjoin globally.
- Show to the user url of ohttp relays and payjoin directory (still in payjoin settings page).
- `PayjoinStatus.aborted` names the outcome, not the mechanism: the payjoin was aborted and the payment landed via a plain broadcast of the original transaction instead.
- Fixed the root cause of several stale-status symptoms: the local datasource never mapped `isCompleted`/`isExpired` back from the database row, so every fetch silently returned a "never completed" session no matter what had actually been persisted — surfacing as stuck "requested"/"proposed" statuses long after a payjoin had actually completed or aborted.
- Transaction details now derives the *displayed* payjoin status from the broadcast transaction itself (not the possibly-lagging session row), and resolves straight to the transaction view instead of a payjoin-session placeholder that swaps out moments later.
- Session watchers now settle once a session resolves instead of continuing to poll: no more redundant re-broadcasting of the original transaction on a stale expiry, and our own broadcasts stay watched (via direct, sync-coordinator-bypassing lookups) until visible in the local wallet, instead of depending on one best-effort sync that can be throttled or lost to a race — no more manual resyncs needed to see a just-completed payjoin.
- Explains the BIP78 fee contribution on the receive side of a completed payjoin (the mining fee for the receiver's contributed input, deducted from the amount received) so the net amount doesn't look like a bug.
- System label ("Payjoin") for transactions that actually completed via a real payjoin — never applied to a plain-broadcast fallback, which isn't one.
- Actually enforces "prefer re-contributing an already-exposed UTXO over a fresh one": the coin-selection call now tries the already-exposed candidates first, and only falls back to the full candidate set if none of them work for this specific proposal — the underlying privacy heuristic (UIH2 avoidance) always keeps the final say, we only control which set it gets to choose from. A separate, temporary "Payjoin exposed" system label tracks UTXOs currently committed to an in-flight or failed proposal, distinct from the "Payjoin" label above; it's removed once the transaction actually completes, since the coin is spent and the transaction itself now carries the definitive tag.
- Preserves the user's own typed label on a completed payjoin send.


ethicnology:
Hey @Claude I expect your review/audit for this huge PR

PR #2468, opened 2026-07-17
feat: trezor integration

### LINKED ISSUE #2294: Review/Merge hardware wallet PRs

ethicnology:
Testing and finalizing @anipy1 contribution to be merged in develop


BullishNode (changes requested):
## Request changes — two findings

### 1. [P1] A previously signed transaction survives edits and can be broadcast instead of the transaction currently shown

`SendCubit.updateSignedBitcoinTx` stores the Trezor-signed raw transaction, but neither [`backClicked`](https://github.com/SatoshiPortal/bullbitcoin-mobile/blob/19ddf60736ae97d82766ee48798bcf3192b73c27/lib/features/send/presentation/bloc/send_cubit.dart#L206-L215) nor the [amount-change path](https://github.com/SatoshiPortal/bullbitcoin-mobile/blob/19ddf60736ae97d82766ee48798bcf3192b73c27/lib/features/send/presentation/bloc/send_cubit.dart#L923-L932) clears it.

Reproduction:

1. Build and sign transaction A with Trezor.
2. Navigate back and change the amount or recipient, producing transaction B.
3. Return to confirmation. The screen displays the state for B, but `signedBitcoinTx` still contains A.
4. [`onConfirmTransactionClicked`](https://github.com/SatoshiPortal/bullbitcoin-mobile/blob/19ddf60736ae97d82766ee48798bcf3192b73c27/lib/features/send/presentation/bloc/send_cubit.dart#L2043-L2065) sees a non-null signed transaction, skips rebuilding and signing, and broadcasts A.

This creates normal-flow wrong-payment potential. It is P1 rather than P0 because transaction A was previously reviewed and approved on the device, but the app must never broadcast it while showing the user the details of B.

Fix: invalidate `signedBitcoinTx` and all related finalized artifacts whenever any transaction-defining input changes, and again when rebuilding the unsigned PSBT. Add a regression test covering sign A → back/edit → review B → confirm, asserting that A cannot be broadcast.

### 2. [P1 security, medium confidence] The import callback uses predictable, unauthenticated request correlation

The app accepts any URI matching `bullbitcoin://trezor-callback` and forwards it to the connector after only checking the scheme and host ([`TrezorDeeplinkListener`](https://github.com/SatoshiPortal/bullbitcoin-mobile/blob/19ddf60736ae97d82766ee48798bcf3192b73c27/lib/features/trezor/ui/trezor_deeplink_listener.dart#L47-L56)). The pinned connector generates the callback ID from the current millisecond timestamp and accepts the matching payload without authenticating its origin ([connector source](https://github.com/anipy1/trezor_connect/blob/b649c02569878f0a3edce913449fcb5c4177088e/lib/trezor_connect.dart#L70-L114)). During import, the returned xpub and descriptor are then used to construct the wallet ([repository mapping](https://github.com/SatoshiPortal/bullbitcoin-mobile/blob/19ddf60736ae97d82766ee48798bcf3192b73c27/lib/features/trezor/data/trezor_device_repository_impl.dart#L90-L106)).

A malicious local app that can observe or estimate when the request was created could race a forged callback containing its own xpub and descriptor. If successful, Bull Bitcoin would import an attacker-controlled watch-only wallet and later show receive addresses controlled by that attacker.

Confidence is medium because this requires a hostile local app plus successful timing observation/guessing, and no end-to-end exploit was demonstrated. The underlying trust-boundary weakness is nevertheless concrete and should be fixed before shipping xpub import.

Fix: use a cryptographically random, single-use state value; bind it to the expected operation and request; reject missing, mismatched, expired, or replayed state; strictly validate the response schema and account data; and prefer an app-owned verified universal/App Link where feasible.

ethicnology:
**Merge blocked** upstream dependency `trezor_connect` (anipy1/trezor_connect@b649c02, fork of cake-tech/trezor_connect) need to be prioritized for review. 

We need either `cake-tech` to merge @anipy1 's PR and then review their dependency or make our own fork, still a review of this new dependency is required before integrating it to the codebase.

 - https://github.com/cake-tech/trezor_connect/pull/1
 - https://github.com/cake-tech/trezor_connect/pull/2

Also registerCallback() uses a predictable timestamp as the sole correlation token for Trezor Suite deeplink callbacks. This must be reviewed and patched before this branch merges. Finding #1 (stale signed tx) is fixed and tested in 541a5bb46; finding #2 remains open and gates the merge.

cc: @i5hi @thibistaken @BullishNode

PR #2478, opened 2026-07-20
feat(payjoin): harden core engine (watchers, resume, guards)

ethicnology:
Hardens the core payjoin engine: watchers, session-resume logic, and guards against invalid states. Models the fallback-to-normal-tx path explicitly via PayjoinStatus.aborted + a DB isAborted flag, instead of deriving it from isCompleted && txId == null. Extends the Payjoin entity with logRef, canManuallyBroadcastOriginal, and isCompleted/isAborted/isExpired getters.

Builds on #2443's typestate API. Addresses part of #2416 (payjoin improvements).

Merge order: 3rd, after PR1. Base of PR3, PR4, PR5 — must merge before all three.
PR1 → PR2 (this) → {PR3, PR4, PR5}

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -72,6 +72,7 @@ graph TB
     LABELS --> CORE
     PAY --> RECIPIENTS
     PAYJOIN --> UTXO_MGMT
+    PAYJOIN --> LABELS
     PIN_CODE --> CORE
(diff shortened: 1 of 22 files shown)


ethicnology:
@Claude audit this PR please

PR #2480, opened 2026-07-20
feat(receive): payjoin toggle badge in TopBar with eligibility guard

ethicnology:
Moves the payjoin toggle badge into the TopBar (new TopBar.action slot), visible only in Bitcoin mode. Badge color comes from the theme (context.appColors.success/error, new onSuccess token), never hardcoded. Long-press on the badge navigates to payjoin settings. Payjoin is disabled when there's no eligible confirmed UTXO/balance (isPayjoinToggleable). If the entered amount is below the payjoin minimum, the pj= param is dropped from the QR without destroying the session, with an explanatory InfoCard. Reacts live to the payjoin-enabled setting (stream from PR1).

Relates to #947 (Better Payjoin Status), #1584 (Default payjoin settings).

Merge order: 4th (parallel with PR3, PR5), after PR2. No dependency on PR3/PR5.
PR2 → PR4 (this)   ‖ PR2 → PR3   ‖ PR2 → PR5


ethicnology:
@Claude audit this PR please

PR #2494, opened 2026-07-21
feat(consolidation): Liquid UTXO consolidation with decoy outputs

wired-pasteque:
## What

Adds Liquid wallet UTXO consolidation: when a wallet accumulates too many
confirmed L-BTC UTXOs, it's offered (via a banner on wallet/send/swap
screens) a guided flow to sweep them into fewer outputs, keeping the wallet
well under the ~256-input confidential-transaction limit before it ever
blocks a real send or swap.

## The decoy output, and why it's there

A naive consolidation transaction is an obvious on-chain signature: many
inputs, one output, no counterparty — a many-in/one-out self-transfer. On a
transparent chain that's a privacy leak (heuristics like this are a
well-known way to cluster a wallet's UTXOs together). On Liquid, amounts and
assets are already blinded by confidential transactions, but the *shape* of
the transaction (input/output count) is still visible.

Each consolidation batch is built with **two outputs instead of one**: the
real drained amount, plus a second, small "decoy" output back to a
fresh address of the wallet's own. This makes the transaction look like an
ordinary two-output payment rather than a textbook consolidation. Because
Liquid hides amounts, the decoy value itself is never visible to
anyone but the wallet — only the shape of the tx (2 outputs, not 1) is
public, and that shape is indistinguishable from a normal payment.

The decoy UTXO is then frozen (best-effort) so it's never accidentally
swept back into a future consolidation round.

## Also included

- **Frozen-UTXO enforcement in real sends/swaps** — Liquid sends and
  auto-swap executions now explicitly exclude frozen UTXOs (previously only
  the UI bookkeeping did), so a frozen decoy or user-frozen coin can never
  be accidentally spent by a normal transaction.
- **Address-reservation datasource** — prevents two addresses (e.g. two
  consolidation batches, or a batch running close to another
  address-generating event) from ever being handed out with the same
  index, closing a real address-reuse bug found during testing.
- **Sync gap-limit fix** — a multi-batch consolidation round now syncs the
  wallet against the actual highest reserved address index afterward,
  closing a real "balance disappeared" bug where funded addresses could
  land outside the default 20-address gap-limit scan.
- **`buildCustomTx`** — a new explicit-UTXO-list Liquid transaction builder
  (vs. the old coin-selecting builder), needed so consolidation, frozen-UTXO
  exclusion, and the decoy output can all specify exact inputs/outputs.
  
## Testing

Tested using small amounts of UTXOs.

https://github.com/user-attachments/assets/5ff3e93b-9486-4df4-9ad4-14c244e4aef8

### DIFF
--- a/integration_test/coins_test.dart
+++ b/integration_test/coins_test.dart
@@ -4,12 +4,11 @@ import 'package:bb_mobile/core/fees/domain/fees_entity.dart';
 import 'package:bb_mobile/core/seed/data/models/seed_model.dart';
 import 'package:bb_mobile/core/settings/domain/settings_entity.dart';
 import 'package:bb_mobile/core/storage/sqlite_database.dart';
-import 'package:bb_mobile/core/wallet/data/datasources/bdk_wallet_datasource.dart'
-    show NoSpendableUtxoException;
 import 'package:bb_mobile/core/wallet/data/datasources/frozen_wallet_utxo_datasource.dart';
 import 'package:bb_mobile/core/wallet/data/repositories/wallet_repository.dart';
 import 'package:bb_mobile/core/wallet/domain/entities/wallet.dart';
 import 'package:bb_mobile/core/wallet/data/repositories/wallet_address_repository.dart';
+import 'package:bb_mobile/core/wallet/domain/wallet_build_tx_exceptions.dart';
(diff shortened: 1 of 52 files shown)


ethicnology:
@wired-pasteque can you rebase this branch?

Issue #15, opened 2023-06-29
Unsigned PSBT download button not working


BullishNode:
Downloading an unsigned PSBT doesn't work. Should be opening file directory so the user can save it directly onto a Cold Card virtual disk.

BullishNode:
@mocodesmo the unsigned PSBT is saved automatically in the dowloads folder. It should prompt to decide where to save the file, because it will usually be either a virtual disk (USB) or an SD card, not downloads folder.

i5hi:
This has been fixed since 0.1.8

PR #2498, opened 2026-07-23
Accept lightning-prefixed LNURLs

BullishNode:
## Summary

- Handle the `lightning:` scheme explicitly before the existing payment-request heuristics.
- Delegate LNURL normalization and validation to Satoshifier.
- Preserve the BOLT11 path for `lightning:`-prefixed invoices without first treating them as invalid LNURLs.
- Pin the Bull SDK monorepo dependencies together to the companion parser fix.

The existing parser could extract some prefixed LNURLs through its broad embedded-LNURL heuristic. This change makes ownership explicit: Satoshifier normalizes LNURLs, while the mobile parser only classifies the payload and routes it to the correct parser.

## Testing

- `TZ=UTC make checks`
  - analyzer and formatting checks
  - 572 application tests
  - 17 `bull_ui` tests
  - `bull_ui_catalogue` test
- Focused Linux native integration tests: 7/7 passed, covering BIP21 LNURL, BIP21 BOLT11, lowercase and uppercase `lightning:` schemes, prefixed LNURLs, prefixed Lightning Addresses, prefixed BOLT11 invoices, and encoded Lightning Address URLs.
- Android production-debug APK built and installed successfully. The focused Android test was not run because the existing emulator data uses a newer database schema than `develop`; its data was deliberately left intact.

## Dependency

Companion Bull SDK PR: https://github.com/SatoshiPortal/bull_sdk/pull/15

Merge the SDK PR first, then repin this PR to the resulting durable SDK commit before merging it.

### DIFF
--- a/integration_test/payment_request_test.dart
+++ b/integration_test/payment_request_test.dart
@@ -5,7 +5,10 @@ import 'package:flutter_test/flutter_test.dart';
 
 Future<void> main({bool isInitialized = false}) async {
   TestWidgetsFlutterBinding.ensureInitialized();
-  if (!isInitialized) await Bull.init();
+  if (!isInitialized) {
+    await Bull.initLogs();
+    await Bull.initFlutterRustBridgeDependencies();
+  }
 
   const btcAddress = 'bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq';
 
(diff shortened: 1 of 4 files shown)


BullishNode:
Thanks for the review. Addressed as follows:

1. The mobile PR is now draft. Bull SDK #15 must merge first; this PR will then be repinned to the resulting durable SDK commit before it is made ready.
2. Added native integration coverage for both `LIGHTNING:LNURL1...` and `lightning:user@domain.com`. Both pass. Satoshifier deliberately receives the complete URI and removes the scheme with an anchored, case-insensitive expression, keeping normalization in one place.
3. Confirmed that `LnurlParser.parse` constructs `boltz.Lnurl` and awaits `lnurl.validate()` before returning, so parse-time LNURL validation was not removed.
4. Documented why the BOLT11 vector differs and that parsing extracts, rather than enforces, its expiry timestamp.
5. The missing `await` predates and is unrelated to this parser change, so it is intentionally excluded from this focused PR.

The expanded focused native suite passes 7/7.

PR #2509, opened 2026-07-25
refactor(payjoin): UX updates from clean-install testing

i5hi:
1. Autoswap announcement: copy direction corrected (Liquid → Bitcoin) and only shown when the Liquid balance actually reaches the trigger threshold (reuses AutoSwap.passedRequiredBalance)
2. Payjoin eligibility no longer requires a confirmed balance — unconfirmed UTXOs are contributable (BDK listUnspent), so payjoin activates immediately on funding
3. Payjoin announcement removed from home — payjoin education lives in the disclaimer pop-up and settings
4. Receive screen: payjoin control moved from the TopBar chip to a toggle row under the address
5. Payjoin settings: disclaimers become a one-time pop-up on first enable (re-openable via a settings row); minimum amount, session expiry and servers now live under a collapsed Advanced section
6. Send confirmation: payjoin shown as a ticked details row (like self-transfer), replacing the info card

PR #2513, opened 2026-07-28
fix(payjoin): harden session recovery and convergence

ethicnology:
This PR is stacked on #2509 and should be reviewed after it. It also carries the merge that propagates the native dependency pin from #2506.

## `b098d5821` Keep Session Recovery in the Foreground

**Problem:** Payjoin recovery was coupled to dependency-locator setup. Because locator setup can run in different execution contexts, registering dependencies could start protocol processing outside the foreground lifecycle. Obsolete Workmanager schedules could also keep background execution paths alive.

**Solution:** Register the Payjoin repository lazily and start recovery explicitly from foreground app initialization. Cancel previously registered Workmanager schedules while keeping foreground cold-start log pruning.

## `bcf77bc23` Make Payjoin Sessions Converge

**Problem:** A session is driven by several independent signals: PDK callbacks, user actions, expiry timers, startup recovery, SQLite state, network publication, and blockchain observations. These could overlap, while whole-row writes and broad error boundaries let stale handlers overwrite newer state or repeat protocol side effects.

This was one class of concurrency and convergence bugs with several possible outcomes:

- two receiver sessions selecting the same UTXO before either proposal was persisted
- proposal, expiry, fallback, and watcher callbacks acting on the same session concurrently
- a duplicate sender request published before SQLite rejected it
- a stale transition replacing a newer terminal outcome
- a local failure after successful publication entering the competing original-transaction fallback
- a watcher observing the right transaction but stopping before its terminal state was persisted
- expired or partially processed sessions remaining stranded after restart

**Solution:** Make session processing serialized, monotonic, and retryable:

- serialize all effects for a session with a session-specific lock
- serialize receiver UTXO selection across sessions until the proposal state is recorded
- reject duplicate sender sessions before publishing another request
- replace whole-row updates with conditional SQLite transitions using the priority `completed > aborted > expired > pending`
- treat successful publication as an irreversible boundary: persistence, synchronization, txid derivation, labeling, and watcher failures cannot trigger the competing fallback afterward
- keep watchers active when a terminal transition fails to persist, so a later observation retries
- reconcile both Payjoin and original transaction ids for receiver and sender sessions
- sweep recoverable expired sessions at startup, isolating failures per session

Repository tests and independent SQLite-connection tests cover the transition races and retry behavior.

## `03aace4c5` Persist Payjoin Disable Intent First

**Problem:** Disabling Payjoin settled active receivers before storing the setting. If settlement failed, Payjoin stayed enabled and could keep accepting requests despite the user's choice.

**Solution:** Persist the disabled setting first, then settle every receiver best-effort. A failed settlement stays durable and is retried at foreground startup, while new receiver creation fails closed.

## `ad9e30ddb` Keep the Receive Amount Summary Reactive

**Problem:** The receive summary derived its sats or fiat suffix through a non-reactive bloc read, so unit, currency, or exchange-rate changes could leave stale text on screen.

**Solution:** Derive the suffix through a `ReceiveBloc` selector so the summary rebuilds whenever a relevant input changes.

## `4363594d6` Make Integration-Test Gaps Explicit

**Problem:** Integration cleanup could mark rows expired while their PDK polling timers were still active. Three unimplemented scenarios also had empty bodies and reported success without testing anything.

**Solution:** Stop PDK polling before expiring cleanup rows, and mark the restart and insufficient-funds scenarios as explicitly skipped with the fixture or harness each still requires.

## `84dac6271` Drop the Unused Ongoing-Only Fetches

**Problem:** `fetchReceivers` and `fetchSenders` still carried an `onlyOngoing` branch with no caller, since the disable sweep and startup recovery read every row and decide from the freshest one under a session lock. A test stub also still referenced the dead parameter.

**Solution:** Remove the parameter and its branch, leaving filtered reads to `fetchAll`, and drop the stale stub.

## Deliberately Unchanged

- **Unconfirmed receiver UTXOs stay supported.** Suitable confirmed candidates are preferred, but fresh wallets can Payjoin immediately.
- **Proposal publication still precedes durable persistence of `proposed`.** This PR prevents an immediate competing fallback after a persistence failure, but does not introduce a transactional outbox. The blockchain arbitrates, and both transactions spend the same sender inputs, so the worst case is a lost negotiation falling back to a plain transaction — never a double payment.
- **Session locks are not purged on terminal outcomes.** Removing a lock while a callback is still queued on it would break mutual exclusion; the map is in-memory only and cleared on restart.
- **A failed receiver settlement is only logged.** It is retried at the next foreground cold start, not at app resume.
- **The receiver path with no candidate UTXO has no test.** Restart behavior is covered by the `resumePayjoinsOnStartup` unit tests.
- No key material, signing primitive, or secret-handling path changes.

### DIFF
--- a/integration_test/payjoin_test.dart
+++ b/integration_test/payjoin_test.dart
@@ -4,6 +4,7 @@ import 'dart:io' show Platform;
 import 'package:bb_mobile/core/blockchain/domain/usecases/broadcast_bitcoin_transaction_usecase.dart';
 import 'package:bb_mobile/core/fees/domain/fees_entity.dart';
 import 'package:bb_mobile/core/payjoin/data/datasources/local_payjoin_datasource.dart';
+import 'package:bb_mobile/core/payjoin/data/datasources/pdk_payjoin_datasource.dart';
 import 'package:bb_mobile/core/payjoin/domain/entity/payjoin.dart';
 import 'package:bb_mobile/core/payjoin/domain/repositories/payjoin_repository.dart';
 import 'package:bb_mobile/core/payjoin/domain/usecases/receive_with_payjoin_usecase.dart';
@@ -38,6 +39,7 @@ Future<void> main({bool isInitialized = false}) async {
   final utxoRepository = locator<WalletUtxoRepository>();
   final payjoinRepository = locator<PayjoinRepository>();
   final localPayjoinDatasource = locator<LocalPayjoinDatasource>();
+  final pdkPayjoinDatasource = locator<PdkPayjoinDatasource>();
   final receiveWithPayjoinUsecase = locator<ReceiveWithPayjoinUsecase>();
   final sendWithPayjoinUsecase = locator<SendWithPayjoinUsecase>();
   final prepareBitcoinSendUsecase = locator<PrepareBitcoinSendUsecase>();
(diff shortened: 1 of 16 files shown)


ethicnology:
@Claude review that PR

PR #2541, opened 2026-07-30
fix(recipients): tolerate missing SINPE owner names

### LINKED ISSUE #2529: SINPE Móvil recipients silently dropped when ownerName is absent
## Symptom
Device logs show repeated `SEVERE Error parsing recipient element — Bad state: ownerName is required for SINPE_MOVIL_CRC`; affected recipients never appear in the recipients list.

## Diagnosis
Mobile requires `ownerName` for SINPE Móvil (`lib/features/recipients/application/dtos/recipient_details_dto.dart:430-443`, same for the SINPE IBAN variants), but the API's schema has `RecipientSinpeMovil.ownerName String?` — nullable **by design** (populated by a Ridivi lookup that can fail or be empty on older records). The recipients gateway's per-element try/catch correctly limits the damage to dropping those recipients.

## Fix
Make `ownerName` optional in the mobile DTO/model for SINPE Móvil and SINPE IBAN types; display fallback chain `ownerName → label → phone number / IBAN`. Note: the server derives a pay order's `beneficiaryName` from the same field, so the pay success screen needs the same fallback (tracked in the pay success issue).

ethicnology:
SINPE Movil/IBAN recipients were dropped when ownerName was absent, since a server-nullable-by-design field was required client-side. Made nullable with label/identifier fallback for display.

Closes #2529

### DIFF
--- a/lib/features/recipients/application/dtos/recipient_details_dto.dart
+++ b/lib/features/recipients/application/dtos/recipient_details_dto.dart
@@ -398,49 +398,43 @@ class RecipientDetailsDto {
         );
 
       // COSTA RICA
+      // ownerName is nullable by design for the SINPE types: the server fills it
+      // from a Ridivi lookup that can fail or be empty on older records.
+      // Requiring it here silently dropped those recipients (#2529).
       case RecipientType.sinpeIbanUsd:
         if (iban == null) {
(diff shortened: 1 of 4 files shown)


ethicnology:
@basantagoswami can you review and test please ?

PR #2542, opened 2026-07-30
fix(exchange): load every order history page

### LINKED ISSUE #2528: Transactions list only ever fetches the newest 50 orders
## Limitation (filed as follow-up, deliberately out of scope of the parsing fix)
`listOrderSummaries` hardcodes `paginator: {page: 1, pageSize: 50}` (`lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart:250`), so exchange transactions older than the newest 50 orders can never appear in the app, and the repository's offset/limit parameters paginate within that same first page. Needs real paging (fetch-on-scroll or page-through) wired from the transactions list.

ethicnology:
listOrderSummaries only fetched page 1 (50 orders), silently dropping older transactions. Now pages through until the backend returns a short page.

Closes #2528

### DIFF
--- a/lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
+++ b/lib/core/exchange/data/datasources/bullbitcoin_api_datasource.dart
@@ -238,28 +238,45 @@ class BullbitcoinApiDatasource implements BitcoinPriceDatasource {
     );
   }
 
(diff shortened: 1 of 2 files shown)


ethicnology:
@basantagoswami can you rebase, review and test please ?

PR #2545, opened 2026-07-30
feat(payjoin): complete Exchange integration and extract workspace package

ethicnology:
This PR completes the Payjoin Exchange integration, hardens session recovery and transaction handling, and extracts the Payjoin protocol into the pure-Dart `packages/bull_payjoin` workspace package.

It is the final delivery branch for the Payjoin series and includes the preceding reliability, Exchange UX, fee-selection, and transaction-history work.

## Payjoin Reliability

- Keep session recovery in the foreground application lifecycle and remove obsolete background recovery schedules.
- Serialize effects per session and UTXO selection across sessions.
- Use conditional, monotonic database transitions so stale callbacks cannot overwrite terminal states.
- Reconcile both negotiated Payjoin transactions and original fallback transactions.
- Resume unfinished sessions and retain watchers until persisted state converges.
- Persist the user's disable intent before settling active receivers.
- Continue settling remaining receivers when one session fails.
- Prevent duplicate sender publication and competing fallback broadcasts.
- Refresh the Receive amount summary whenever its amount, unit, currency, or exchange-rate inputs change.
- Make incomplete integration scenarios explicit instead of reporting empty tests as successful.
- Remove obsolete ongoing-only datasource queries.

## Exchange And Payjoin Integration

- Expose Payjoin capabilities through Exchange order contracts.
- Support Exchange-bound Payjoin sessions.
- Negotiate Buy payouts through Payjoin.
- Negotiate Sell and Pay payins through Payjoin.
- Preserve nominal Payjoin amounts in transaction history.
- Resolve Exchange Payjoins by transaction ID.
- Reconcile missing Exchange order labels before transaction hydration.
- Report disabled Payjoin as disabled rather than offline.
- Add Payjoin-specific Exchange localization and generic asynchronous status UI.

## Exchange Reliability And UX

- Parse Exchange orders independently so one unsupported or malformed order cannot hide the complete order history.
- Support known missing order statuses and tolerate future unknown order values.
- Correct reward amount, fiat-versus-sats, nullable exchange-rate, and confirmation-deadline handling.
- Surface Buy order-creation failures instead of leaving the Continue action unresponsive.
- Clarify Buy payout wallet, payout method, network, purchased amount, and fiat amount.
- Show Pay recipient, amount, confirmation state, and retryable errors.
- Carry the refreshed post-broadcast order into Pay success state.
- Close the Sell post-broadcast double-payment window.
- Upgrade BIP21 URI handling for Exchange Payjoin flows.

## Network Fees

- Add functional fee selection to Sell and Pay instead of always using the fastest rate.
- Support presets, custom sat/vB fees, absolute fees, previews, and relay-floor validation.
- Rebuild the final PSBT at broadcast time so refreshed Exchange amounts use the selected fee rate.
- Prevent stale fee calculations from restoring pre-broadcast state.

## Workspace Extraction

- Add `packages/bull_payjoin` as a pure-Dart package with no dependency on Flutter, `bb_mobile`, `get_it`, or `path_provider`.
- Add `packages/primitives` for shared `Result`, `Failure`, Bitcoin network, money, fingerprint, script type, and outpoint types.
- Remove the legacy `lib/core/payjoin` module.
- Expose narrow role interfaces for sending, receiving, sessions, policy, diagnostics, and lifecycle operations.
- Keep one shared internal engine instance behind the public `Payjoin` facade.
- Keep Payjoin UI and orchestration in the owning Send, Receive, Buy, Sell, Pay, Transactions, Settings, and Status features.
- Add application adapters for wallet operations, blockchain access, transaction monitoring, labels, legacy storage, and logging.
- Replace the root success-screen scaffold with the generic `BullSuccessScreen` from `bull_ui`.

## Persistence And Migration

- Give `bull_payjoin` its own Drift database and committed schema baseline.
- Import existing sessions and policy transactionally from the legacy application database.
- Verify migrations using row counts and a deterministic digest.
- Store a migration marker so completed imports are not repeated.
- Keep the legacy Payjoin tables for rollback compatibility.
- Extend `make drift-migrations` to verify both application and Payjoin schemas.
- Allow tests to inject an isolated Payjoin database path.

## Post-Extraction Audit Fixes

- Replace stale policy read-modify-write operations with serialized targeted updates.
- Reject Signet and Regtest instead of silently treating them as Bitcoin Testnet.
- Keep ordinary transaction history available when Payjoin is unavailable.
- Match abandoned Buy receivers by the stable `pj` endpoint when Exchange rewrites the BIP21 amount.
- Add typed storage, wallet, broadcast, lifecycle, and migration telemetry.
- Hash raw session identifiers before they reach logs or Sentry.
- Preserve sanitized stack traces without forwarding developer messages or protocol payloads.
- Replace audited domain-to-data imports with narrow domain contracts.
- Move `NoSpendableUtxoException` out of the BDK datasource boundary.

## Security

- Wallet models, seed models, Drift rows, PDK objects, and internal repositories are not exported.
- Mnemonics and private wallet material remain inside the application wallet adapter and are loaded only for the active signing operation.
- Raw session identifiers, BIP21 URIs, PSBTs, and developer error messages are not forwarded through structured telemetry.
- Payjoin request networks and domain inputs are validated at package boundaries.

## Non-Goals

- Do not create Payjoin-specific UI inside `bull_payjoin`.
- Do not add Liquid Payjoin support.
- Do not drop the legacy Payjoin tables in this release.
- Do not expose application data models or infrastructure through the package API.
- Do not broaden the UI Kit migration beyond the approved generic components.

## Recommended Review Order

1. Payjoin reliability and Exchange behavior commits.
2. `packages/primitives`.
3. `packages/bull_payjoin/lib/src/domain` and public exports.
4. Persistence, migration, and engine implementation.
5. Application setup and runtime adapters.
6. Feature use-case and dependency-injection migrations.
7. Build tooling, Drift configuration, and tests.


ethicnology:
@Claude review me that please

ethicnology:
@Claude audit this PR

ethicnology:
This push addresses the seven high-priority findings from the audit, in five commits. Each fix is covered by a regression test that **fails on the unfixed base and passes after the fix** (red→green), and the whole suite is green: `flutter analyze --fatal-warnings --fatal-infos` reports no issues, `dart fix --dry-run` is clean, `make unit-test` exits 0.

### Findings → commits

| Finding | Commit | What changed |
|---|---|---|
| **H1** — manual fallback broadcast the stale caller entity | `a568d1a` | Broadcast the fresh persisted entity, not the caller's possibly-stale copy, so a post-retry stale snapshot can't send the superseded original while the new session is marked aborted against it. |
| **H2** — sender retry left stale watchers, new fallback watcher never armed | `a568d1a` | Fallback watchers are now keyed per `(session, txid)`. A retried sender watches **both** the superseded and the new original — required because BIP78 lets the receiver broadcast the old original after a timeout, and seeing it land must resolve the retried session to aborted rather than let its own expiry fallback broadcast the new original on top of an already-settled payment. |
| **H3** — signed original posted before any local persistence | `a568d1a` | Write-ahead persist before the directory POST, then refresh the row to the post-POST session state. |
| **H4** — background recovery ran through the retained workmanager handler | `81f700e` | The background isolate registers the payjoin roles but no longer starts session recovery; only the foreground composition root resumes sessions. |
| **H5** — deterministic payjoin-open failure bricked all bitcoin sends | `dddc4a5` | Invalid legacy rows are quarantined (logged, skipped) instead of failing the import; a corrupt legacy policy falls back to defaults. An unreadable `payjoin.sqlite` is moved aside (with its WAL/SHM) and reopened fresh once, re-populated from the legacy tables. |
| **H6** — payjoin latch was bloc-scoped, not order-scoped | `b7f017c` | Teardown handlers now refuse to rebuild an order while a payment is in flight; the payjoin watch is tagged with its order id and a resolution for a different order is ignored; post-broadcast success is gated on the exchange actually seeing the payin (same `payinStatus` gate as the order poll). |
| **H7** — rate conversion awaited outside any try | `b7f017c` | Moved inside the try so a price-API failure surfaces an error state instead of a permanently stuck spinner. |

### A regression found by running it for real

While bringing the integration test up, I found and fixed a **regression in the H3 fix itself** (`a568d1a`): moving the persist ahead of the POST made the returned session model capture the *pre-POST* session state (`WithReplyKey`), so the proposal poll replayed a state that is never `PollingForProposal` and the negotiation silently stalled. The fix refreshes the model (and the persisted row) to the post-POST state. This is exactly the class the reproducers were meant to catch, and it only surfaced end-to-end — worth a look during review.

### End-to-end verification

The payjoin integration test was **excluded from CI's aggregate** and had never run green anywhere. It now runs in CI (`de71a41`), skipping gracefully when the `TEST_ALICE_MNEMONIC` / `TEST_BOB_MNEMONIC` fixtures are absent instead of failing the job. With funded testnet wallets it passes end-to-end: a full payjoin completes (receiver contributes a confirmed UTXO, sender finalizes and broadcasts) and a receiver expires cleanly with no sender. Watch/test timeouts were loosened to fit the ~30s directory long-poll cycle and testnet variance.

### Residual risks / follow-ups (not in this push)

- **Retry double-original**: after a sender retry, two signed originals exist (the receiver holds both if both POSTs succeeded), and BIP78's specified receiver fallback can broadcast the old one at any time. If the inputs differ, both can confirm. The watchers now *detect* this and converge the session, but nothing prevents it outright — the complete fix is to pin the retry to the expired session's inputs so the two originals conflict. Proposed as a documented follow-up.
- The ~27 minor findings (Sell latching the unsigned PSBT's txid for bip44/bip49, `SellError`/`PayError`/`WithdrawError` leaking dev messages, `aborted` not fully terminal vs `markCompleted`, `Dio` leaked per open attempt, `PayjoinPolicyStore.save()` bypassing the lock, `receive_state.dart` scientific-notation BIP21 amount, etc.) are untouched and tracked separately.

i5hi (approved):

PR #2547, opened 2026-07-31
feat(core): add human-assisted wallet entropy

BullishNode:
## Disclaimer

**We have no reason to believe that BDK has any RNG or entropy-generation defect.** This is an experimental hardening project intended to reduce reliance on any single entropy path and to reduce attack surface. It is not a response to an ongoing incident, known vulnerability, or suspected defect in BDK.

## Summary

- Generate 12-word BIP39 mnemonics from 128 bits of explicitly combined entropy.
- Require both a fresh 64-byte operating-system CSPRNG draw and a completed 500-sample finger ceremony for every generated wallet.
- Count only qualified pointer observations: Flutter-synthesized events, consecutive duplicate positions, and non-finite movement data cannot advance ceremony progress.
- Mix additional touch-contact measurements already supplied by Flutter: minor radius, contact size, orientation, and tilt, alongside position, delta, event and arrival timing, pressure, and major radius.
- Combine inputs through a domain-separated, length-framed SHA-512 pool with a counter and retained 256-bit state.
- Use BDK only for deterministic `Mnemonic.fromEntropy` encoding; do not invoke BDK's random mnemonic constructor.
- Fail closed if either required input is missing or the OS source fails its catastrophic sanity checks.
- Remove CPU-jitter, system-statistics, IMU/magnetometer, time-strengthening, and ceremony-bypass paths.
- Add no motion-sensor dependency, motion permission, camera access, microphone access, or persistent gesture storage.
- Keep new-wallet output at the existing 12-word product standard.

## Why

The goal is defense in depth against a hypothetical defect or supply-chain compromise affecting one randomness path. The human transcript is a physically distinct, deliberately uncredited input: it may preserve meaningful unpredictability if the OS RNG becomes predictable, but the implementation does not claim a fixed number of entropy bits from finger movement.

Five hundred is a conservative pacing threshold, not an assertion of 500 bits or any fixed per-event contribution. Pointer observations are correlated and device pipelines differ, so the gate rejects known low-value callbacks while the cryptographic pool mixes the complete accepted transcript.

Motion sensors were deliberately excluded after considering privacy and complexity. Their incremental entropy is difficult to quantify, they expand permissions and lifecycle behavior, and prior plugin defects demonstrate the reliability surface they add. The existing explicit finger ceremony supplies the independently motivated physical input without introducing another permission or dependency.

This is not designed to protect against a malicious application or runtime that can observe or replace both the OS draw and the pointer transcript.

## Verification

- `make checks`
- Whole-project `flutter analyze --fatal-warnings --fatal-infos` — no issues
- `dart fix --dry-run` — `Nothing to fix!`
- Formatting — 1,613 files unchanged
- 856 app tests passed
- 17 `bull_ui` tests passed
- 1 catalogue test passed
- Independent Python `hashlib` known-answer vector for the 500-sample SHA-512 pool construction

## Draft checklist

- [ ] Real-device wallet creation on Android
- [ ] Real-device wallet creation on iOS
- [ ] Create-and-restore round trip using the generated 12-word mnemonic
- [ ] Product/security review of the mandatory finger ceremony


BullishNode:
## Entropy strategy and design rationale

### Threat model

This change is a hedge against a narrow but serious failure mode: the wallet's nominal CSPRNG path returns predictable bytes because of a platform defect, integration bug, or supply-chain compromise. It is not evidence of such a defect in BDK, Dart, Android, or iOS.

The construction requires two inputs for every generated wallet:

1. A fresh 64-byte draw from Dart's platform-backed `Random.secure()` path.
2. A completed transcript of 500 qualified pointer samples generated by deliberate finger movement.

Neither input is treated as optional. There is no predictable fallback, low-bit reseed, or path that silently continues after a source failure. This is directly motivated by the class of failure documented in Block's analysis of Coldcard's predictable fallback and 32-bit reseed: https://engineering.block.xyz/blog/predictable-rng-fallback-and-32-bit-reseed-in-coldcard-firmware

The security claim is deliberately conditional: if either the OS draw or the touch transcript remains unknown to the attacker, the combined result should remain computationally unpredictable under the SHA-512 assumptions. If the entire app/runtime is malicious and can observe or replace both inputs, this design cannot help.

### Bitcoin Core inspiration

This is inspired by the shape of Bitcoin Core's RNG, not a line-for-line port.

Bitcoin Core maintains a process-wide 256-bit RNG state. Its `MixExtract` operation finalizes SHA-512 over newly gathered material plus the prior state and a counter, returns up to the first 32 digest bytes, and retains the last 32 bytes as the next state. Its strong path mixes fresh OS randomness, and the OS failure path terminates rather than substituting weak randomness. See Bitcoin Core's current design documentation and implementation:

- RNG design overview: https://github.com/bitcoin/bitcoin/blob/67efced1fc83a0b7215cc1513e7c4754fee0f12f/src/random.h#L25-L72
- `RNGState::MixExtract`: https://github.com/bitcoin/bitcoin/blob/67efced1fc83a0b7215cc1513e7c4754fee0f12f/src/random.cpp#L340-L455
- Strong seeding and OS entropy: https://github.com/bitcoin/bitcoin/blob/67efced1fc83a0b7215cc1513e7c4754fee0f12f/src/random.cpp#L470-L610

Our pool follows the same useful core pattern:

- 32-byte retained state.
- Monotonic counter.
- SHA-512 mix/extract.
- Output from the first digest half and rekeyed state from the second half.
- Fresh OS entropy required at strong extraction time.
- Best-effort cleanup of intermediate buffers.

We add explicit domain and length framing for ceremony start, individual samples, ceremony completion, and OS input. This prevents ambiguous concatenations and cross-purpose reuse. A completed ceremony is a one-use capability: extraction consumes it, and starting a new ceremony invalidates an older unconsumed completion.

We intentionally do not copy every Bitcoin Core source. CPU jitter, system statistics, generic timestamps, and mobile IMU data were removed because we could not justify a defensible entropy estimate for them in this environment. Keeping many impressive-looking inputs without a clear threat-model contribution would be noise theater. The design instead has two understandable mandatory components.

### Why finger swipes instead of the camera

A camera can contain physical sensor noise, but the application normally receives frames after a device-specific image-processing pipeline rather than controlled raw sensor samples. Compression, denoising, auto-exposure, stabilization, frame buffering, and vendor behavior make the actual contribution difficult to characterize. Camera frames also create large sensitive buffers with more copies and a less convincing best-effort wiping story. Activating a camera during seed creation also has a materially different privacy and trust signal for users.

Finger input is smaller and more explicit. Each sample commits to the event kind, pointer identifier, coordinates, deltas, framework timestamp, an independent monotonic-clock reading, pressure, major and minor touch radius, contact size, orientation, and tilt. Samples are serialized into fixed 112-byte buffers, mixed immediately, and then wiped on a best-effort basis. No motion-sensor plugin or additional ceremony sensor lifecycle is required.

The 500-sample threshold is a pacing and freshness gate, not an estimate of 300 events' entropy. Flutter-synthesized events, consecutive duplicate positions, and non-finite movement data do not advance the gate. Completion time remains device-dependent. Taps remain accepted for users who cannot perform a continuous gesture. We make no claim such as “one bit per event.”

The camera could be reconsidered as an optional additional source only if we define the exact capture pipeline, retention rules, supported platforms, and a conservative entropy argument. It should not replace the OS CSPRNG floor or the explicit human contribution merely because it yields more bytes.

### Why 12 words

The product currently uses 12-word BIP39 mnemonics. This change therefore extracts exactly 16 bytes (128 bits) and passes them to BDK's deterministic `Mnemonic.fromEntropy` encoder. The OS source is intended to provide the full security floor. The finger transcript is supplemental and uncredited, not a reason to market the mnemonic as having more than 128 bits of entropy.

### What each component actually contributes

- **OS CSPRNG:** the primary entropy source and intended 128-bit security floor. A 64-byte draw is required, checked for exact length, rejected if all bytes are identical, and rejected if it exactly repeats the previous process-local draw. These checks detect catastrophic failures only; they are not statistical certification.
- **Finger transcript:** a physically distinct hedge. If the OS bytes are predictable but the attacker cannot predict the gesture transcript, it can still prevent precomputation or mass exploitation. Its effective security is the transcript's real unpredictability in that scenario, not a guaranteed 128 bits.
- **SHA-512 pool and retained state:** a cryptographic combiner and state-evolution mechanism. Under the hash assumptions, mixing a weak or attacker-known component does not cancel a strong unknown component. Retained state also separates repeated inputs across ceremonies.
- **BDK:** deterministic BIP39 encoding only. The PR deliberately avoids creating a second nominal RNG draw through BDK because two user-space CSPRNGs may ultimately share the same OS entropy root and should not be described as independent sources.

### Limitations

- Dart's garbage collector may copy sensitive values; zeroization is best effort. A forensics-grade memory threat model would require moving the pool and secret handling into controlled native memory.
- Pointer delivery and `Random.secure()` both execute inside the application trust boundary. Full process compromise defeats both.
- The sanity checks catch obvious catastrophic RNG behavior, not biased or adversarially structured output.
- Real-device Android/iOS validation and a create-and-restore round trip remain required before this leaves draft status.

BullishNode:
Follow-up after adversarial gesture capture

I instrumented the ceremony locally in a debug-only build (the capture harness and traces are not part of this PR; wallet generation was disabled during capture). The useful result was negative: 500 accepted pointer events are not an entropy estimate. Human emulator traces completed in roughly 6–8 seconds, a tiny repetitive gesture still passed, and one deterministic Android-injected straight swipe reached 500 accepted samples in 3.5 seconds.

Commit 3523d5fdc therefore adds anti-degeneracy gates while preserving the original design:

- completion requires at least 500 accepted samples;
- at least 10 monotonic seconds must separate the first and final accepted samples;
- movement must span at least 50% of both canvas axes;
- progress reflects the slowest incomplete gate, so it cannot display 100% early;
- every accepted sample is still mixed, now with explicit sequence order and pointer device kind in addition to coordinates, deltas, timestamps and contact metadata;
- regression tests prove that fast 500-sample and narrow repetitive transcripts cannot complete the ceremony.

These conditions improve ordinary gesture diversity but are deliberately not credited as entropy bits. They cannot distinguish a sufficiently adaptive script or protect against an OS that observes or fabricates touch input. The human input remains a supplemental hedge against a broken RNG when the rest of the input path is honest. This follow-up intentionally does not add camera/sensor permissions or re-architect the RNG stack.

PR #2551, opened 2026-08-02
feat(mnemonic): checksum-aware last word entry and leak hardening

ethicnology:
The last word of a BIP39 sentence carries a part of entropy and the whole checksum:

| sentence | possible last words | out of 2048 |
|---|---|---|
| 12 words | 128 | 6.3 % |
| 15 words | 64 | 3.1 % |
| 18 words | 32 | 1.6 % |
| 21 words | 16 | 0.8 % |
| 24 words | 8 | 0.4 % |

The count is `2^(11 - CS)`: the last word holds the `CS` checksum bits preceded by the `11 - CS` trailing bits of the entropy, and each entropy tail yields exactly one valid word.

## What this changes

**Checksum-aware last word**

- On the last field, the suggestion chips and the count label come from `Mnemonic.lastWordCandidates` (bip39_mnemonic 4.1.0, published for this): it derives the answer from the checksum instead of searching the wordlist, so it costs one SHA256 per word returned.
- The last field's badge answers "can this word close the sentence" instead of "is this a word": a wordlist word that is not one of the candidates is flagged before submit, and re-evaluated as earlier words change.
- Auto fill deliberately keeps resolving against the **full** wordlist, not the candidates. If an earlier word was mistyped into another valid word, the real last word is no longer a candidate, yet a two-letter prefix of it can single out a different one — completing to it would silently restore a wrong-but-checksum-valid mnemonic, the exact error the checksum exists to catch. The user's own word is completed instead, and the checksum then surfaces the typo. Tapping a chip remains an explicit, safe choice.
- An auto-filled word is certain by construction (its prefix had a single match left), so the field swallows further edits until it is emptied; only the completed last word dismisses the keyboard, since the sentence is then complete.

**Security — this touches recovery-phrase entry, please review accordingly**

- Entry errors never echo a typed word. `bip39_mnemonic` builds its exception messages out of the offending word, so the new sealed `MnemonicEntryFailure` family drops those strings at the boundary (`logMessage` carries the exception's runtime type only) and the catch-all variant shows a generic localized message.
- The import mnemonic page is excluded from screenshots and the recents thumbnail for as long as it is mounted (`PrivacyScreen`).

**Performance**

- One controller per word instead of one `setState` per keystroke: a keystroke rebuilds **277 widgets instead of 1093**, and never rebuilds a `TextField`. Counted with `debugPrintRebuildDirtyWidgets`.

**Also**

- `labeled_text_input` moves from the labels feature into `lib/core/widgets/inputs/`.
- The unused seed word-list repository and use case are dropped.
- The widget's behaviour is pinned down by 520 lines of new widget tests, including regression tests for the typo trap above.


ethicnology:
@claude audit that please

ethicnology:
@wired-pasteque can you perform a manual QA please?

ethicnology:
Audit addressed in `2d6aa56ab`..`a19ef26b1` (24/24 widget tests green locally, `analyze` clean):

1. **Locked field swallowing backspace** — fixed in `2d6aa56ab`: a deletion edit now empties the field and unlocks it (same end state as the clear icon), so the undo instinct works while the stray-keystroke protection stays. Regression test added.
2. **Chip tap not dismissing the keyboard** — fixed in `429c4bdf1`: a tapped chip on the last field now unfocuses it, same as the auto fill. Test added.
3. **Count label outliving the completed word** — fixed in `5aab880e2`: the label now leaves on the same condition as the chips row. Test added.
4. **Synchronous FocusNode disposal in `didUpdateWidget`** — fixed in `a19ef26b1`: retired nodes and locks are disposed in a post-frame callback, matching the parent's handling of the old controllers.
5. **One-frame lock gap** — left as is: unreachable in practice, as the audit itself notes (IME events arrive per frame).
6. **`on bip39.MnemonicException` coverage** — verified against the bip39_mnemonic 4.1.0 source: the only throws reachable from `Mnemonic.lastWordCandidates` are `MnemonicUnexpectedPartialSentenceLengthException` and `MnemonicWordNotFoundException`, both extending `MnemonicException`. The raw-string throw in `bits.dart` is unreachable on this path — its input is always a `0`/`1` string built by `toRadixString(2)`.
7. **Translations** — matches the repo workflow: keys land in `app_en.arb`/`app_fr.arb` first, the other locales fall back to English until translated.

ethicnology:
1. **Locked field swallowing backspace** — fixed in `451d3df2f`: a deletion edit now empties the field and unlocks it (same end state as the clear icon), so the undo instinct works while the stray-keystroke protection stays. Regression test added.
2. **Chip tap not dismissing the keyboard** — fixed in `2e979a303`: a tapped chip on the last field now unfocuses it, same as the auto fill. Test added.
3. **Count label outliving the completed word** — fixed in `1b62d94f0`: the label now leaves on the same condition as the chips row. Test added.
4. **Synchronous FocusNode disposal in `didUpdateWidget`** — fixed in `b143f357c`: retired nodes and locks are disposed in a post-frame callback, matching the parent's handling of the old controllers.
5. **One-frame lock gap** — left as is: unreachable in practice, as the audit itself notes (IME events arrive per frame).
6. **`on bip39.MnemonicException` coverage** — verified against the bip39_mnemonic 4.1.0 source: the only throws reachable from `Mnemonic.lastWordCandidates` are `MnemonicUnexpectedPartialSentenceLengthException` and `MnemonicWordNotFoundException`, both extending `MnemonicException`. The raw-string throw in `bits.dart` is unreachable on this path — its input is always a `0`/`1` string built by `toRadixString(2)`.
7. **Translations** — matches the repo workflow: keys land in `app_en.arb`/`app_fr.arb` first, the other locales fall back to English until translated.

@wired-pasteque the branch was force-pushed (rebase), please re-pull before the manual QA. Worth exercising on device: last-field suggestions on a 24 word phrase, backspace on an auto-filled word, and a phrase with one mistyped-but-valid word (it must be rejected, not silently completed)

wired-pasteque:
@ethicnology, checked on my side - everything behaves as expected.

Tested on both 12-word (128 candidates) and 24-word (8 candidates) phrases - each works with the matching count, and they refresh when an earlier word changes. Backspace on an auto-filled word empties the field and unlocks it, so retyping works. The mistyped-but-valid case behaves correctly: autofill completes to my own word and the submit is rejected on the checksum - no silent completion.

Also pushed a commit: the required-label empty case was the only one still using a one-shot snackbar - every other error in this flow shows inline on the field. Changed it to match, so the style is consistent and the message stays visible until it's fixed.

PR #2554, opened 2026-08-03
fix(wallet): remove unsafe external links

leesalminen:
## Summary

Remove unsafe or broken external actions from wallet import screens:

- **Buy a device** from watch-only import
- **Wallet guides** from watch-only import
- **Buy a device** from Coldcard import

This removes the Coinkite store and broken documentation redirects while preserving all wallet import paths. The stale localization keys are removed as well.

## Validation

- `make checks`
- repository-wide scan confirms no `*.coinkite.com` or `*.ethicnology.com` references remain


leesalminen:
@claude follow up review pls?

Issue #11, opened 2023-06-27
Change address must inherint label of outbound transaction label


BullishNode:
When making an outbound transaction, the change address utxo must inherit the label of the outbound transaction.

We know which utxo is the change utxo (supplied automatically by txbuilder) by the fact that it is not the recipient's address that was input by the user.

Currently not working as of 0.1.2

i5hi:
Will be addressed in #40

i5hi:
Fixed in 0.1.88

PR #2562, opened 2026-08-06
fix(test_wallet_backup): keep the mnemonic out of the bloc state

ethicnology:
`TestWalletBackupState` (freezed) held the full `mnemonic` and `passphrase` in bloc state. Freezed's generated `toString()` included both fields in clear text, so any state log, crash report, or debug session could expose the user's 12/24 words. This violates our two load-bearing rules: secrets are ephemeral (read from secure storage at the point of use, never cached in bloc state) and secrets never reach logs.

The rewrite also surfaced two pre-existing correctness flaws:

- `VerifyPhysicalBackupUsecase` (then dead code) always verified against the **default mainnet** wallet, not the selected one.
- `VerifyMnemonicScreen` pushed the success screen **before** the verification result was known.

## What changes

- **State carries no secret.** `TestWalletBackupState` is now `statusError`, `wallets`, `selectedWallet`, `verificationStatus` — the regenerated `toString()` is clean by construction.
- **Sealed reads at the point of use.** Both screens fetch the mnemonic/passphrase through `TestWalletBackupBloc.loadSelectedWalletMnemonic()`, a presentation-layer method that forwards the use-case result without storing it. The word-order game lives in ephemeral widget state (rule: local widget state stays in `StatefulWidget`).
- **Verification re-reads the seed.** `VerifyPhysicalBackupUsecase` (previously orphaned) is now wired and takes the selected wallet's fingerprint; it compares and returns only a `bool`. Success navigation happens only on `verificationStatus.success`.
- The hardcoded English error string is replaced by the existing `testBackupErrorIncorrectOrder` l10n key.
- Two orphaned planning documents removed from the repo root (separate commit).

## What it deliberately does not change

- `import_mnemonic`'s cubit also holds the mnemonic in state — same class of issue, left for a dedicated PR.
- Remaining dev-facing English error strings in the bloc (`'Failed to load wallets: $e'`) — pre-existing rule-#11 violation, follow-up.
- `CompletePhysicalBackupVerificationUsecase` semantics are unchanged (still marks default wallets).

## Alternatives considered

- **Keeping the mnemonic in a private bloc field** (not in state): rejected — it still caches key material for the bloc's whole lifetime, which the "secrets are ephemeral" rule forbids.
- **Screens calling the use case directly via get_it**: rejected — it breaks the `ui → presentation → usecase` layering; the bloc method keeps the chain while never persisting the secret.

## Reading order

1. `chore: remove orphaned planning documents from the repo root` — independent cleanup.
2. `fix(test_wallet_backup): keep the mnemonic out of the bloc state` — the fix, with its tests.


ethicnology:
@wired-pasteque can you test physical backup flow?

PR #2575, opened 2026-08-07
fix(send): show the order number instead of the internal uuid

### LINKED ISSUE #2572: Transfer ID shows internal UUID instead of the order number
## Description
On the Confirm Send screen, the "Transfer ID" row displays a raw internal UUID. 

This is an internal identifier that should never be exposed in the UI - it means nothing to the user and it leaks implementation detail. The value that should be shown here, the order number, is currently being surfaced in the "Note" row instead, which is the wrong place for it - Note should hold user free-text, not a system identifier.

Beyond being the wrong value, the UUID wraps onto two lines and makes the row visually heavy compared to the rest of the screen.

### Expected behavior
- The internal UUID is not displayed to the user anywhere in the UI.
- The Transfer ID row shows the order number
- The Note row shows the user's actual note, or is hidden entirely when no note was entered.

Image

wired-pasteque:
Fixes: #2572

Simulator Screenshot - iPhone 17e - 2026-08-07 at 16 47 39

### DIFF
--- a/test/features/send/presentation/bloc/send_cubit_test.dart
+++ b/test/features/send/presentation/bloc/send_cubit_test.dart
@@ -354,6 +354,8 @@ void main() {
 
   setUpAll(() {
     registerFallbackValue(_FakeNewLabel());
+    registerFallbackValue(_bitcoinLocalWallet());
+    registerFallbackValue(BigInt.zero);
   });
 
   setUp(() {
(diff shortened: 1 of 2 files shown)


test/features/send/presentation/bloc/send_cubit_test.dart
```diff
@@ -532,4 +534,59 @@ void main() {
       },
     );
   });
+
+  group('SendCubit.onAmountConfirmed note handling', () {
+    /// A Lightning send whose invoice carries a description. The description
+    /// belongs to whoever created the invoice, not to the user, so it must not
```
ethicnology: "not to the user"

The label belongs to both the sender and the receiver

"refund for beers" should appears on both side as a label

ethicnology (changes requested):

test/features/send/presentation/bloc/send_cubit_test.dart
wired-pasteque: done

Issue #2661, opened 2026-08-12
Security (Low, MEMPOOL F-6): default-server guard is bypassed by non-canonical hosts


basantagoswami:
**Severity:** Low

## Summary

The “same as default” check compares normalized strings without canonical host parsing. A trailing DNS dot or userinfo prefix can make the default host appear distinct and allow it to be saved as custom.

## Affected areas

- `set_custom_mempool_server_usecase.dart`
- URL normalization

## Impact

The settings UI can show a redundant custom server and misleadingly present the built-in default as replaced. No independent security control is bypassed.

## Remediation

Compare canonical parsed hosts, strip trailing dots, reject userinfo, and reuse the hardened URL parser for equality checks.

Issue #2651, opened 2026-08-12
Security (Low, BITBOX F-6): abandoned operations hold the device lock for up to 60 seconds


basantagoswami:
**Severity:** Low

## Summary

Leaving a BitBox screen does not cancel the in-flight device request. The old operation can retain shared locks until its 60-second read timeout, causing the next action to wait behind it.

## Affected areas

- BitBox USB bridge read loop
- BitBox Rust API mutex scope
- Device-screen disposal lifecycle

## Impact

After backing out, subsequent BitBox operations can appear hung for up to a minute before self-recovering.

## Remediation

Introduce cancellation on route disposal or forcibly invalidate the device session so waiting reads and locks fail promptly.

Issue #2649, opened 2026-08-12
Security (Low, BITBOX F-4): unsupported P2WSH signing can abort the app


basantagoswami:
**Severity:** Low

## Summary

The BitBox signing dependency contains a `todo!()` branch for P2WSH and wrapped-P2WSH inputs. Release panic behavior aborts the process instead of returning a catchable error.

The current single-signature wallet model does not generate these inputs, so the branch is latent today.

## Affected areas

- BitBox dependency `script_config_from_utxo`
- Future multisignature signing support

## Impact

If the supported input set expands without addressing this branch, signing such a transaction will terminate the app.

## Remediation

Replace the panic with a typed unsupported-output error and add a regression test before enabling any P2WSH wallet path.

Issue #2648, opened 2026-08-12
Security (Low, BITBOX F-3): Android USB permission returns before the user responds


basantagoswami:
**Severity:** Low

## Summary

On Android, the USB plugin returns success immediately after requesting permission rather than waiting for the user's response. The app then opens the device too early and misclassifies the resulting no-permission error.

## Affected areas

- `BitboxFlutterPlugin.kt`
- `bitbox_device_datasource.dart`

## Impact

The first USB connection predictably fails with a generic message even when the user grants permission moments later.

## Remediation

Resolve the request only from the Android permission-result broadcast and map all no-permission variants to the dedicated permission failure.

PR #2669, opened 2026-08-12
fix: resolve 61 security audit findings, mitigate 7, document 5 upstream

### LINKED ISSUE #2593: Security (High, SEND F-1): post-sign edits show a new payment but broadcast the old signed transaction
**Severity:** High

## What

After a Ledger/BitBox transaction is signed, navigating back and changing the recipient, amount, or wallet does not clear `signedBitcoinTx`.

The app rebuilds and displays the new payment on the review screen, but confirmation skips signing and broadcasts the previously signed transaction. The success screen also reports the new values.

Affected code:

- `send_cubit.dart:219-229` — back navigation does not clear the signed transaction
- `send_cubit.dart:2093-2136` — signing is skipped when `signedBitcoinTx` remains set
- `send_cubit.dart:2013-2016` — the old signed hex is broadcast verbatim
- `send_screen.dart:902-967,1032-1034` — review and confirmation UI

## Impact

Loss or misdirection of funds. The valid transaction sent on-chain pays the old recipient and amount from the old wallet, while the review and success screens claim the corrected payment was sent.

## Remediation

- Clear `signedBitcoinTx` whenever any transaction-shaping input changes.
- Bind signed transactions to the same payment fingerprint used by fee previews.
- Before broadcasting signed hex, decode it and verify that its inputs and outputs match the reviewed payment.
- Add regression tests covering recipient, amount, and wallet changes after hardware signing.

### LINKED ISSUE #2594: Security (High, SEND F-2): stale MAX state can turn a fixed-amount payment into a full-balance send
**Severity:** High

## What

The internal `sendMax` state survives recipient and payment-request changes even though the visible MAX toggle resets.

A user can enable MAX for an amountless address, navigate back, and then open a fixed-amount BIP21 payment request. Because the request already contains an amount, the amount screen is skipped. The stale `sendMax` value remains enabled, causing transaction construction to ignore the requested amount and drain the wallet’s spendable balance to the recipient.

Affected code:

- `send_cubit.dart:940` — `sendMax` is set but not reset when the request changes
- `send_cubit.dart:1589,1679` — transaction construction resolves the stale drain state
- `send_cubit.dart:1421,1481` — preview builders receive `sendMax` directly
- `bdk_wallet_datasource.dart:243-244` — drain mode ignores `amountSat`

## Impact

The recipient can receive the wallet’s entire spendable balance when the user intended to pay a fixed invoice amount. An attentive user may notice the changed amount during review, but the flow provides no explicit warning that MAX remained enabled.

## Remediation

- Reset `sendMax` whenever the recipient or payment request changes.
- Reset it whenever a request containing an embedded amount is accepted.
- Reject drain mode for fixed-amount payment requests.
- Apply the same guarded drain calculation to preview and final transaction construction.
- Add regression tests covering back navigation followed by a fixed-amount request.

### LINKED ISSUE #2596: Security (High, IMPORT F-6): watch-only descriptor import accepts and exposes private extended keys
**Severity:** High

## What

The watch-only descriptor import accepts an extended private key (`xprv`) where only public key material should be allowed.

Descriptor parsing uses `Bip32Keys.fromBase58(..., bypassVersion: true)`, so private version bytes are accepted. The imported private key is then:

- Stored in wallet metadata as `xpub`
- Persisted in the SQLite database
- Displayed in the wallet details screen as “Pubkey”
- Made available through copy and descriptor-export functionality

Affected code:

- `satoshifier/lib/descriptor.dart:165-168,227-231` — parses extended keys while bypassing version validation
- `wallet_metadata_service.dart:223-246` — stores the value as public wallet metadata
- `wallet_metadata_datasource.dart:9-14` — persists it in SQLite
- `wallet_details_screen.dart:79-89` — presents it as copyable public information

## Impact

A user can unknowingly import full-spend private key material through a flow presented as watch-only. The application then treats the secret as shareable public metadata.

Copying the displayed “Pubkey,” exporting the descriptor, sharing a backup, or exposing the application database can disclose the private key and result in complete loss of funds controlled by it.

## Remediation

- Reject all private extended-key version bytes in watch-only import paths.
- Require an explicitly public extended key after normalizing supported SLIP-132 versions.
- Add a shared public-key-only validator and use it for descriptor, raw xpub, and label imports.
- Detect previously imported private keys and warn affected users without displaying or exporting the secret.
- Add regression tests for `xprv`, `yprv`, `zprv`, testnet private variants, and valid public variants.

### LINKED ISSUE #2597: Security (High, LABELS F-2): BIP329 xpub labels accept and re-export private extended keys
**Severity:** High

## What

BIP329 records with type `xpub` accept extended private keys such as `xprv`.

Validation only checks that the value is valid Base58Check data with the expected decoded length. It does not verify that the version bytes represent a public extended key.

The accepted value is stored in the labels database and exported verbatim as public wallet metadata.

Affected code:

- `label_entity.dart:109-123` — validates checksum and length but not public/private version bytes
- `bip329_codec.dart:196-201` — accepts the xpub reference unchanged
- `bip329_codec.dart:52-53` — exports the value unchanged
- `labels_table.dart` — persists the value in the labels database

## Impact

A labels file can place full-spend private key material into storage that the wallet treats as public metadata.

If the user later exports or shares their labels, the private key is included verbatim. Disclosure gives an attacker control of every wallet derived from that key and can result in complete loss of funds.

## Remediation

- Reject private extended-key versions in all `xpub` label records.
- Reuse the same public-key-only validator used by watch-only descriptor imports.
- Support known public SLIP-132 versions explicitly instead of validating only length.
- Prevent export of any existing record that is identified as private key material.
- Warn affected users and provide a safe way to delete contaminated records.
- Add regression tests for mainnet and testnet public/private extended-key variants.

### LINKED ISSUE #2598: Security (High, LOG F-4): secret key material can be written to shareable application logs
**Severity:** High

## What

The local TSV logger stores `error.toString()` verbatim. Some parsing exceptions embed their input value in the error message, allowing secret material to enter `bull_logs.tsv`.

Confirmed paths include:

- Mistyped Recoverbull vault keys, potentially exposing nearly the entire key
- Decrypted vault plaintext containing mnemonic material
- Invalid BIP39 words from restored mnemonics
- Stored exchange API-key fragments
- Newly generated exchange API keys returned in an invalid response

Affected code:

- `logger.dart:61-69` — writes the complete exception string into the TSV error column
- `recoverbull_repository.dart:40-66` — logs exceptions produced while handling vault keys and plaintext
- `restore_vault_usecase.dart:45` — logs mnemonic parsing failures
- `update_latest_encrypted_backup_usecase.dart:62-66` — logs restored mnemonic failures
- `bullbitcoin_api_key_datasource.dart:44` — logs API-key decoding failures
- `exchange_auth_screen.dart:295` — can log a response containing a newly generated API key

## Impact

The application explicitly allows and encourages users to export these logs or attach them to support conversations.

A shared log can therefore disclose a near-complete vault key, seed material, or an exchange API key. In some cases, only a small brute-force search would be required to reconstruct the original secret.

## Remediation

- Never pass secret-processing exceptions directly to the logger.
- Map failures at these boundaries to constant, non-sensitive messages.
- Log only the exception type where diagnostic classification is required.
- Add logger-level redaction for long hex, Base64, mnemonic-like, and API-key-like values as defense in depth.
- Audit every key, seed, mnemonic, descriptor, and decrypted-backup parsing path for raw exception logging.
- Add tests asserting that representative secrets never appear in generated TSV rows.

### LINKED ISSUE #2599: Security (High, NATIVE F-1): iOS backups include wallet databases, payjoin data, and logs
**Severity:** High

## What

Security-sensitive application data is stored under the iOS Documents directory without applying `NSURLIsExcludedFromBackupKey`.

The directory is included in iCloud and Finder/iTunes device backups by default. No backup-exclusion handling is present in the application.

Affected data includes:

- The primary Drift/SQLite application database
- Bitcoin and Liquid wallet-engine databases
- Payjoin state
- Wallet descriptors, labels, transaction history, and related metadata
- Shareable TSV application logs

Affected code:

- `sqlite_database.dart:65-66,112` — places the application database under Documents
- `payjoin_setup.dart:33` — places the Payjoin database under Documents
- `bdk_facade.dart:194` — places BDK wallet databases under Documents
- `lwk_facade.dart:14` — places LWK wallet databases under Documents
- `main.dart:116` — places logs under Documents
- No code applies the iOS backup-exclusion resource key

## Impact

Anyone able to access a device backup can obtain the wallet’s structure and history, including descriptors, labels, counterparties, swap and Payjoin activity, and logs.

If private extended keys were accepted through affected import or label paths, those secrets can also be copied into the backup, escalating the issue to complete loss of funds.

## Remediation

- Move regenerable and security-sensitive databases from Documents to an appropriate Application Support or Library location.
- Apply `NSURLIsExcludedFromBackupKey` to files that must remain local and must not enter device backups.
- Define an explicit backup policy for every persisted file.
- Migrate existing installations and apply exclusions to already-created files.
- Test the final application container with iOS backup tooling to verify the files are excluded.
- Document which recovery data is intentionally backed up and ensure it is encrypted appropriately.

### LINKED ISSUE #2600: Security (Medium, SEED F-1): SeedSigner signed transactions cannot be finalized or broadcast
**Severity:** Medium

## What

The SeedSigner signed-PSBT return path is broken. SeedSigner returns a trimmed PSBT containing the unsigned transaction and partial signatures, without the full UTXO data.

The wallet first tries to finalize it with `bitcoin_base`, which fails because the trimmed PSBT lacks that UTXO data. Its fallback combines the signed PSBT with the original PSBT, but then extracts a raw unsigned transaction before finalization and passes those raw transaction bytes to a PSBT parser.

That parser requires PSBT magic bytes, so the fallback always fails.

Affected code:

- `broadcast_signed_tx_cubit.dart:28-62` — SeedSigner parsing, finalization, and fallback path
- `broadcast_signed_tx_cubit.dart:49` — passes serialized raw transaction bytes to `Psbt.deserialize`
- The regression was introduced during the migration from `bdk_flutter` to `bdk_dart`

## Impact

SeedSigner-signed transactions returned through QR or paste cannot be broadcast from the wallet.

Users whose funds are secured exclusively by SeedSigner cannot spend through the intended flow. Workarounds such as importing the seed into a hot wallet undermine the security model the user selected.

## Remediation

- Combine the original and signed PSBTs using BDK.
- Finalize the combined PSBT before extracting its transaction.
- Verify that finalization succeeded and fail with a specific user-facing error otherwise.
- Broadcast the serialized transaction extracted from the finalized PSBT.
- Add a regression fixture matching SeedSigner’s trimmed PSBT output.
- Surface signed-PSBT scan and finalization errors directly on the scan page.

### LINKED ISSUE #2601: Security (Medium, SEND F-3): ambiguous broadcast retries can pay the recipient twice
**Severity:** Medium

## Summary

A failed broadcast is treated as proof that the transaction was not accepted. Retrying from the confirmation screen builds and signs a new transaction instead of rebroadcasting the original signed payload.

If the first transaction reached the node but the response was lost, both transactions can confirm when they spend different inputs, paying the recipient twice.

## Affected areas

- `broadcast_bitcoin_transaction_usecase.dart`
- `send_cubit.dart`

## Impact

An ambiguous timeout or disconnect can become a double payment. A malicious or unreliable Electrum server can trigger the ambiguous failure condition.

## Remediation

- Preserve and retry the same signed transaction after a broadcast error.
- Before rebuilding, query for the original transaction ID and clearly warn that the first attempt may have succeeded.
- Add coverage for acceptance followed by a dropped response.

### LINKED ISSUE #2602: Security (Medium, SEND F-4): swap-provider outages can strand or misclassify Lightning sends
**Severity:** Medium

## Summary

Failures while loading swap limits are not handled consistently in the send flow. When the swap provider is unavailable, valid Lightning invoices may be reported as invalid, while other paths leave the amount screen permanently disabled.

## Affected areas

- `send_cubit.dart`
- `send_screen.dart`

## Impact

Lightning and chain-swap sends can become unusable during a provider outage, with misleading or absent feedback and no in-screen recovery.

## Remediation

- Handle every `loadSwapLimits` failure at the call site.
- Reset the submission/disabled state on every error exit.
- Distinguish provider unavailability from invalid payment data.
- Add tests for invoice, Lightning-address, and cross-network sends during an outage.

### LINKED ISSUE #2603: Security (Medium, SEND F-6): pending swaps can be reused across wallets
**Severity:** Medium

## Summary

Pending send swaps are reused by invoice without checking the wallet. A swap created while wallet A is selected can therefore be reused for a payment funded by wallet B.

The funding transaction, history attribution, and refund-key origin can then disagree. If the seed that created the swap key is later unavailable, recovery of a failed funded swap may be impossible.

## Affected areas

- `create_send_swap_usecase.dart`
- `boltz_swap_repository.dart`
- `send_cubit.dart`

## Impact

Swap records can be attributed to the wrong wallet, and the failure/refund path can strand funds in the worst case.

## Remediation

Bind swap lookup and reuse to the selected wallet ID and swap type. Create a new swap whenever that binding does not match, and add a cross-wallet regression test.

### LINKED ISSUE #2605: Security (Medium, LABELS F-1): BIP329 imports can silently freeze arbitrary outpoints
**Severity:** Medium

## Summary

BIP329 records with `spendable:false` immediately create durable coin freezes during import. The import has no freeze review or confirmation, and its success message reports only label counts.

Unattributed outpoints can also enter the global frozen set and outlive wallet deletion.

## Affected areas

- `bip329_codec.dart`
- `import_labels_usecase.dart`
- `wallet_freeze_adapter.dart`
- `prepare_bitcoin_send_usecase.dart`

## Impact

A crafted labels file can silently make known wallet coins unavailable to every send path. Recovery exists through coin control but is not disclosed by the import flow.

## Remediation

- Make freeze import explicitly opt-in or require confirmation listing the affected outputs.
- Only apply freezes to outputs currently owned by the target wallet.
- Include freeze counts in the import result and provide a direct recovery path.

### LINKED ISSUE #2606: Security (Medium, LABELS F-3): imported labels can forge privileged system labels
**Severity:** Medium

## Summary

Imported BIP329 text can exactly match internal system-label names. Those imported labels then receive privileged behavior: they cannot be deleted and can cause receive-address generation to skip labeled addresses without a bound.

Enough forged labels on consecutive future addresses can push the displayed receive address beyond the wallet scan gap.

## Affected areas

- `bip329_codec.dart`
- `label_system.dart`
- `wallet_address_repository.dart`
- `bdk_wallet_datasource.dart`
- `labels_widget.dart`

## Impact

Incoming funds sent to the displayed address may not be detected by normal synchronization and can appear lost until an expert rescan is performed.

## Remediation

- Namespace trusted system labels separately from user/imported text.
- Reject or escape reserved names at import.
- Bound address skipping to the configured scan gap and show a recovery warning.
- Enforce system-label privileges in the repository layer.

### LINKED ISSUE #2607: Security (Medium, QR F-1): one invalid BBQR frame can wedge the scan session
**Severity:** Medium

## Summary

One invalid BBQR frame can permanently wedge the current scan session. Out-of-range part indexes and zero totals are accepted, while switching to a stream with a different part count clears only part of the decoder state.

After the bad frame, valid frames cannot complete the scan and recovery requires leaving the page.

## Affected areas

- `bbqr_options.dart`
- `bbqr.dart`
- `broadcast_signed_tx_cubit.dart`
- `broadcast_signed_tx_page.dart`

## Impact

Anyone able to place a QR in the camera view can deny the current signed-transaction scan session. The UI offers little or no explanation.

## Remediation

- Reject totals below one and indexes outside the declared range.
- Fully reset decoder state when stream parameters change or joining fails.
- Surface a visible recoverable error.
- Add regression coverage for invalid indexes, zero totals, and mixed streams.

### LINKED ISSUE #2608: Security (Medium, QR F-2): unbounded UR sequence lengths can exhaust device memory
**Severity:** Medium

## Summary

The UR fountain decoder trusts the first frame's declared sequence length and allocates memory proportional to it without a practical bound.

A small crafted QR can therefore request a very large allocation before the stream is validated.

## Affected areas

- `lib/core/urqr/urqr.dart`
- Pinned `bc-ur-dart` fountain decoder
- All `QrScannerWidget` consumers

## Impact

A malicious QR can pin substantial memory, wedge the scanner, or terminate the mobile app due to memory exhaustion.

## Remediation

- Enforce conservative sequence-count and message-size limits before passing frames to the decoder.
- Reset state and report a recoverable error when limits are exceeded.
- Fix or patch the pinned decoder so it validates before allocation.
- Test oversized declarations across every scanner entry point.

### LINKED ISSUE #2609: Security (Medium, LOG F-1): Sentry scrub failure can send unsanitized breadcrumbs
**Severity:** Medium

## Summary

Sentry's `beforeSend` scrub attempts to clear an unmodifiable stack-frame map and throws for events containing stack traces. The SDK catches that exception and sends the partially processed event before the breadcrumb scrub has run.

## Affected areas

- `report.dart`
- Sentry event-processing configuration

## Impact

For users who enabled reporting, breadcrumbs containing wallet metadata can leave the device despite the intended privacy scrub.

## Remediation

- Remove or safely isolate the unsupported frame mutation.
- Perform breadcrumb minimization before any operation that can fail.
- Make each scrub step fail closed and independently testable.
- Add a regression test asserting the final serialized event contains no sensitive breadcrumb fields.

### LINKED ISSUE #2610: Security (Medium, LOG F-2): navigation breadcrumbs expose transaction and wallet identifiers
**Severity:** Medium

## Summary

Navigation breadcrumbs retained for Sentry include route arguments. Bull Wallet routes place transaction, swap, order, payjoin, and wallet identifiers in path or query parameters, so those identifiers are serialized with recent navigation history.

## Affected areas

- `report.dart`
- Router paths for transaction, swap, order, payjoin, and wallet screens

## Impact

Crash reports from consenting users can reveal which specific financial records they recently viewed, linked to the installation identifier.

## Remediation

Keep route names only. Remove `from_arguments`, `to_arguments`, path parameters, and query parameters from navigation breadcrumbs before transmission, with serialization-level tests.

### LINKED ISSUE #2611: Security (Medium, LOG F-3): native crash reporting bypasses user consent
**Severity:** Medium

## Summary

The native Sentry SDK is initialized in release builds regardless of the user's reporting choice. Native crash handling and related platform integrations can upload events directly, bypassing the Dart consent gate and scrub.

## Affected areas

- `report.dart`
- Native Sentry SDK configuration on Android and iOS

## Impact

Native crash data can be collected for users who declined reporting or never granted consent. These events also bypass the application's privacy minimization.

## Remediation

- Gate native crash handling, watchdog termination tracking, and app-hang tracking on consent.
- Disable or close native integrations when consent is revoked.
- Verify behavior on both platforms with consent enabled and disabled.
- Update consent copy to match any unavoidable residual collection.

### LINKED ISSUE #2612: Security (Medium, NATIVE F-2): iOS startup failures can reactivate disabled background tasks
**Severity:** Medium

## Summary

iOS registers four periodic background tasks at native startup. Dart later cancels them because background execution is intentionally disabled, but that cancellation is skipped if application initialization fails first.

## Affected areas

- `ios/Runner/AppDelegate.swift`
- `lib/main.dart`
- `background_tasks/handler.dart`

## Impact

After an initialization error, iOS may still launch disabled wallet-sync and swap-processing work. This reintroduces background/foreground database and swap-state races the application intended to avoid.

## Remediation

- Do not register disabled tasks in `AppDelegate`.
- Keep registration and cancellation behind one persisted application-level policy.
- Ensure the initialization-error path also confirms cancellation.
- Add an iOS test for startup failure before work-manager initialization.

### LINKED ISSUE #2613: Security (Medium, BIP85 F-1): the current default seed can display the wrong derived secret
**Severity:** Medium

## Summary

Stored BIP85 paths are re-derived from the current default seed without checking the stored source fingerprint. After the default wallet changes, the same row can silently display a different mnemonic or entropy value.

## Affected areas

- `fetch_all_bip85_derivations_with_entropy_usecase.dart`
- `bip85_datasource.dart`
- Default-wallet deletion and selection flows

## Impact

A user can record the newly displayed value as a backup even though it does not control the wallet originally created from that row. If the original seed is gone, funds in the derived wallet may become unrecoverable.

## Remediation

- Compare each row's `xprvFingerprint` with the derivation key before revealing a secret.
- Group or label derivations by source wallet and visibly reject mismatches.
- Handle rows when their source wallet is deleted.
- Add a regression test that changes the default seed between creation and display.

### LINKED ISSUE #2614: Security (Medium, BIP85 F-2): derived secrets lack screen-capture protection
**Severity:** Medium

## Summary

The BIP85 page displays full derived key material without the screenshot, screen-recording, and task-switcher protection used by the application's other mnemonic screens.

## Affected areas

- `bip85_home_page.dart`
- `bip85_derivation_widget.dart`
- `core/mixins/privacy_screen.dart`

## Impact

Revealed derived mnemonics or entropy can be captured by platform snapshots, recording, malware, cloud-backed screenshots, or brief physical access. Possession of that material permits theft from the derived wallet.

## Remediation

- Apply the existing privacy-screen lifecycle to the entire BIP85 route.
- Exclude revealed secrets from accessibility semantics where appropriate.
- Verify screenshot and app-switcher behavior on Android and iOS.

### LINKED ISSUE #2616: Security (Medium, BROADCAST F-1): PushTx reports success without verifying broadcast
**Severity:** Medium

## Summary

The NFC PushTx path reports an in-app broadcast success after merely attempting to open the tag's URL. The launch result and actual network broadcast are not verified, and the tag can supply an arbitrary URI.

## Affected areas

- `broadcast_signed_tx_cubit.dart`
- `broadcast_signed_tx_page.dart`

## Impact

A malicious tag or ordinary connectivity failure can make the user believe an irreversible payment was broadcast when it was not. The signed transaction remains unspent, but the false confirmation can cause real-world payment disputes or loss of the only retained copy.

## Remediation

- Never map external URL launch to `isBroadcasted`.
- Show a neutral “opened externally; verify broadcast” state.
- Require HTTPS and validate/confirm the destination host.
- Check the launch result and verify the tag checksum before review.

### LINKED ISSUE #2617: Security (Medium, BROADCAST F-2): signed-transaction review bypasses Tor
**Severity:** Medium

## Summary

When Tor is enabled, the signed-transaction review screen fetches input parent transactions through a direct socket even though the subsequent broadcast uses the configured SOCKS5 proxy.

## Affected areas

- `electrum_servers_adapter.dart`
- `core/electrum/frameworks/drift/datasources/electrum_remote_datasource.dart`
- Signed-transaction review flow

## Impact

The Electrum server or a network observer can correlate the user's real IP with the exact transaction about to be broadcast. Onion Electrum servers also cannot serve this review path.

## Remediation

- Route review fetches through the same proxy-aware client as broadcast.
- If a protected fetch is unavailable while Tor is enabled, omit input enrichment with a visible explanation instead of going direct.
- Add integration coverage proving no direct socket is opened in Tor mode.

### LINKED ISSUE #2618: Security (Medium, FEES F-1): unbounded fee-oracle values enable manipulation and send denial
**Severity:** Medium

## Summary

All values from the configured fee server are accepted without practical bounds, ordering checks, or safe arithmetic. A malicious response can pin presets to unusable rates, invert tiers, reject every custom fee, or trigger integer overflow in the minimum-fee comparison.

## Affected areas

- `mempool_fees_model.dart`
- `mempool_fees_mapper.dart`
- `fees_entity.dart`
- Transaction creation fee validation

## Impact

A compromised fee oracle can cause stuck transactions, misleading overpayment choices, or denial of send flows. Overflow can also invalidate the intended absolute-fee floor.

## Remediation

- Reject or clamp each rate to documented safe bounds.
- Enforce monotonic fee tiers.
- Cap the relay-floor input independently.
- Use checked or `BigInt` arithmetic for absolute-fee comparisons.
- Add hostile-response tests at the parsing and transaction-building boundaries.

### LINKED ISSUE #2619: Security (Medium, FEES F-2): fee requests have no timeout and can block transaction flows
**Severity:** Medium

## Summary

The fee HTTP client has no connection, send, or receive timeout. A server that accepts a connection and stops responding leaves every caller waiting indefinitely.

## Affected areas

- `fees_datasource.dart`
- Send transaction creation
- Replace-by-fee initialization
- Buy acceleration
- Service-status aggregation

## Impact

A stalled custom or default fee server silently blocks sends, RBF, buy acceleration, and completion of the whole status page.

## Remediation

- Configure bounded connect, send, and receive timeouts consistent with server validation.
- Map timeout failures to recoverable UI states.
- Ensure the status page reports other services independently.
- Consider permitting explicit custom fees when presets are unavailable.

### LINKED ISSUE #2621: Security (Medium, MEMPOOL F-1): editing a custom server can permanently break sends and settings
**Severity:** Medium

## Summary

Editing a custom mempool server URL inserts a new database row without deleting the previous URL. Because the table key includes the URL, the update cannot replace the original row, and later single-row lookups throw when both remain.

## Affected areas

- `mempool_servers_table.dart`
- `set_custom_server_bottom_sheet.dart`
- `mempool_settings_cubit.dart`
- `set_custom_mempool_server_usecase.dart`
- `mempool_server_storage_datasource.dart`

## Impact

Fee estimation, all transaction sends, explorer links, and the settings screen can become persistently unusable. The broken settings screen does not offer an in-app recovery path.

## Remediation

- Atomically replace the existing custom row for the network.
- Use a stable identity or enforce one custom server per network/asset.
- Migrate or safely collapse existing duplicates.
- Make reads degrade to a recoverable reset state.
- Add edit and duplicate-row regression tests.

### LINKED ISSUE #2623: Security (Medium, MEMPOOL F-3): deceptive URLs can connect to an attacker-controlled host
**Severity:** Medium

## Summary

The custom mempool URL parser accepts userinfo, query strings, and fragments instead of validating the parsed host. A value such as `mempool.space@evil.example` can appear trustworthy while the HTTP client connects to the attacker-controlled host.

## Affected areas

- `mempool_url_parser.dart`
- `MempoolServer.tryCreateCustom`
- `mempool_server_item.dart`

## Impact

Socially engineered server text can redirect fee estimation and explorer links to an attacker while retaining a trusted-looking prefix in the UI.

## Remediation

- Parse with `Uri` and validate `uri.host` rather than the raw string.
- Reject userinfo, query, fragment, and unexpected path components.
- Validate the port range and normalize hostnames before display and comparison.
- Add deceptive-URL regression cases.

### LINKED ISSUE #2624: Security (Medium, TXUI F-1): exchange history writes unverified privileged labels
**Severity:** Medium

## Summary

Every transaction-history load writes privileged exchange system labels using server-provided addresses and transaction IDs. These writes occur without ownership checks or user review, including on read-only actions such as CSV export.

Because system-labeled receive addresses are skipped, malicious server data can also push address generation beyond the wallet's scan gap.

## Affected areas

- `get_transactions_usecase.dart`
- `label_exchange_orders_usecase.dart`
- `label_entity.dart`
- `wallet_address_repository.dart`
- `labels_widget.dart`

## Impact

A compromised exchange service can create undeletable, misleading exchange markers. With knowledge of future addresses, it can make a displayed receive address fall outside normal synchronization and cause incoming funds to appear missing.

## Remediation

- Verify that every address and transaction belongs to the intended wallet before labeling.
- Write labels only from an explicit order-completion event, not history reads.
- Restrict labels to accurately supported order types.
- Separate trusted system metadata from user/server-provided strings.

### LINKED ISSUE #2625: Security (Medium, TXUI F-2): CSV export allows spreadsheet formulas and trusts server figures
**Severity:** Medium

## Summary

Transaction CSV export writes swap-provider strings without neutralizing spreadsheet formula prefixes. It also prefers provider-reported fee and status values over available on-chain truth.

## Affected areas

- `csv_transaction_export_formatter.dart`
- `transaction.dart`
- Swap-to-transaction export mapping

## Impact

Opening an exported file in common spreadsheet software can evaluate attacker-controlled formulas and potentially disclose transaction history. Exported accounting data can also contain incorrect fees, statuses, or negative amounts.

## Remediation

- Neutralize every string cell beginning with `=`, `+`, `-`, `@`, tab, or carriage return.
- Prefer verified on-chain fees and confirmation state.
- Represent gross amount and fees separately or prevent invalid negative totals.
- Test exports with hostile provider fields in major spreadsheet applications.

### LINKED ISSUE #2626: Security (Low, SEND F-7): missing exchange rates can crash the amount screen
**Severity:** Low

## Summary

When exchange-rate loading fails, the send state can still treat an empty/default currency as fiat with a zero exchange rate. Converting typed fiat then produces infinity and throws during widget rendering.

## Affected areas

- `send_state.dart`
- `amount_conversions.dart`
- `send_cubit.dart`

## Impact

The amount screen can crash while offline or during a price-service outage. No funds are moved.

## Remediation

Treat non-positive exchange rates as unavailable, prevent fiat conversion until a valid rate exists, and default the input currency to the Bitcoin unit when price data is absent.

### LINKED ISSUE #2627: Security (Low, SEND F-8): closing during broadcast loses confirmation and bookkeeping
**Severity:** Low

## Summary

Closing the send screen while broadcast is awaiting Electrum closes the cubit, but the network operation continues. When it returns, state emission throws and post-broadcast bookkeeping does not run.

## Affected areas

- `send_screen.dart`
- `send_cubit.dart`

## Impact

The transaction may broadcast without confirmation, label persistence, swap linkage, or follow-up synchronization. This creates ambiguity that can lead users to retry.

## Remediation

Prevent dismissal while broadcast is active or move the operation to a route-independent service. Guard closed-state emissions and complete durable bookkeeping immediately after a successful broadcast.

### LINKED ISSUE #2628: Security (Low, SEND F-9): MAX confirmation overstates the recipient amount
**Severity:** Low

## Summary

For a plain MAX send, the confirmation headline shows the entire spendable balance even though the recipient receives that balance minus the network fee.

## Affected areas

- `send_cubit.dart`
- `send_confirm_screen.dart`

## Impact

The review overstates the recipient amount. It errs in the safe direction but makes the confirmation internally inconsistent.

## Remediation

After transaction construction, derive the displayed recipient amount from actual outputs or set the confirmed MAX amount to spendable balance minus the absolute fee.

### LINKED ISSUE #2629: Security (Low, SEND F-10): malformed Bitcoin URIs can become Lightning-address payments
**Severity:** Low

## Summary

Lightning URL and email detection uses unanchored regular expressions after BIP21 parsing fails. A malformed Bitcoin URI containing an email-like substring can therefore be silently reclassified as a Lightning-address payment.

## Affected areas

- `payment_request.dart`

## Impact

The app can contact an unexpected domain and switch payment protocols even though the user scanned a Bitcoin URI.

## Remediation

Anchor Lightning-address and LNURL patterns to the complete input, and do not fall through to a different payment class after a recognized URI scheme fails validation.

### LINKED ISSUE #2630: Security (Low, SEND F-11): payment-request control characters reach stored labels
**Severity:** Low

## Summary

BIP21 labels and Lightning invoice descriptions can contain control characters that flow into the stored transaction label without normalization.

## Affected areas

- `payment_request.dart`
- `utils.dart`
- `send_cubit.dart`
- Label storage boundary

## Impact

Newlines, tabs, and other hidden characters can produce misleading UI and malformed downstream exports.

## Remediation

Trim, length-limit, and strip control characters at the shared label-domain boundary so every source and storage path receives the same validation.

### LINKED ISSUE #2631: Security (Low, SEND F-12): stale coin-control selections block transaction creation
**Severity:** Low

## Summary

Refreshing the available UTXO list does not prune selected outpoints. If a selected coin is spent elsewhere, it disappears from coin control but remains in state and is supplied to transaction construction.

## Affected areas

- `send_cubit.dart`

## Impact

Every build fails with an unknown-UTXO error, while the stale selection is no longer visible or removable from the current UI.

## Remediation

Intersect selected outpoints with the refreshed UTXO set and notify the user when a selection became unavailable.

### LINKED ISSUE #2632: Security (Low, SEND F-13): raw swap errors expose payment metadata
**Severity:** Low

## Summary

Several swap paths wrap and display raw exception strings and also send those strings to logs or telemetry. Backend text can contain invoice, address, URL, or implementation details.

## Affected areas

- `send_cubit.dart`
- `send_screen.dart`
- `decode_invoice_usecase.dart`

## Impact

Payment metadata can appear in the UI, shareable logs, and consented telemetry, while attacker-controlled backend text can influence a security-relevant screen.

## Remediation

Map backend errors to typed failures and localized messages. Keep diagnostic detail local and redact payment material before any logging or reporting boundary.

### LINKED ISSUE #2633: Security (Low, IMPORT F-1): duplicate watch-only import silently overwrites metadata
**Severity:** Low

## Summary

Duplicate watch-only detection queries only default wallets even though imported watch-only wallets are non-default. Re-import therefore reaches an upsert keyed by deterministic wallet ID and silently replaces metadata.

## Affected areas

- `wallet_repository.dart`
- `wallet_metadata_datasource.dart`

## Impact

Re-import can overwrite the wallet label, reset backup flags, and discard birthday or synchronization metadata, forcing an unnecessary full scan.

## Remediation

Check the candidate ID against all existing wallets, return a typed duplicate error, and add coverage that preserves metadata on repeat import.

### LINKED ISSUE #2634: Security (Low, IMPORT F-2): failed mnemonic import leaves a seed that blocks retry
**Severity:** Low

## Summary

Mnemonic import persists the seed before wallet creation. If creation fails or the process is interrupted, an orphaned seed remains and the duplicate check rejects every later import attempt.

## Affected areas

- `import_wallet_usecase.dart`
- `check_duplicate_mnemonic_usecase.dart`

## Impact

The user cannot import a valid mnemonic through the normal flow and receives no guidance that an orphaned seed must be removed.

## Remediation

Make seed persistence and wallet creation atomic, or remove an unreferenced seed on failure. Duplicate checks should distinguish a real wallet from a recoverable orphan.

### LINKED ISSUE #2635: Security (Low, IMPORT F-3): watch-only import hides and ignores network mismatch
**Severity:** Low

## Summary

Watch-only import accepts the network encoded by the input without comparing it with the active application environment, and the review screen does not display that network.

## Affected areas

- `import_watch_only_cubit.dart`
- Watch-only xpub and descriptor import use cases
- `watch_only_details_widget.dart`

## Impact

A wrong-network wallet can be reported as successfully imported and then disappear because the home screen filters it out.

## Remediation

Show the parsed network during review and reject or explicitly confirm mismatches against the active environment.

### LINKED ISSUE #2636: Security (Low, IMPORT F-5): aborted device derivation choice disables the scanner
**Severity:** Low

## Summary

The watch-only scanner marks a frame handled before asking for a Passport/Keystone derivation. Returning without a selected descriptor does not release that latch, so all later frames are ignored.

## Affected areas

- `scan_watch_only_screen.dart`

## Impact

Dismissing the derivation prompt, scanning an unsupported payload, or scanning a normal single-key export can disable the scanner until the page is reopened.

## Remediation

Reset the handled flag on every non-success exit and accept supported plain-descriptor exports directly. Display a useful error for unsupported formats.

### LINKED ISSUE #2637: Security (Low, QR F-3): failed UR streams permanently wedge the scanner
**Severity:** Low

## Summary

The first multipart UR frame permanently anchors stream parameters, and a failed completed stream leaves the decoder in a terminal state. The widget never resets the reader after either condition.

## Affected areas

- Pinned `bc-ur-dart` decoder
- `qr_scanner_widget.dart`

## Impact

An unrelated or hostile frame can deny the current UR scan session. Integrity is preserved by the checksum, but recovery requires leaving and reopening the page.

## Remediation

Reset the reader after decode errors and successful completion, surface a recoverable error, and consider allowing the decoder to resume after a final checksum failure.

### LINKED ISSUE #2638: Security (Low, SEED F-2): SeedSigner static and Specter xpub imports fail silently
**Severity:** Low

## Summary

SeedSigner static and Specter-legacy xpub exports are not recognized by the watch-only parser, and the scan screen silently swallows the parse failure.

## Affected areas

- Satoshifier parser registry
- `watch_only_xpub_parser.dart`
- `scan_watch_only_screen.dart`

## Impact

Users choosing either non-default SeedSigner export format remain in an unexplained scan loop and cannot import the wallet.

## Remediation

Add support for origin-prefixed xpubs, either assemble or visibly reject Specter fragments, and always report unsupported scan formats.

### LINKED ISSUE #2639: Security (Low, SEED F-3): Coldcard PushTx checksum is not verified
**Severity:** Low

## Summary

The Coldcard PushTx handler parses the NFC checksum field but never verifies it against the signed transaction payload. The network marker is also ignored.

## Affected areas

- `broadcast_signed_tx_cubit.dart`

## Impact

Corrupted or substituted data reaches transaction review without the protocol's intended integrity check. Review still limits direct fund risk, but the defense-in-depth guarantee is absent.

## Remediation

Verify the specified truncated SHA-256 checksum before review, require it when the format mandates it, and reject network markers that conflict with the active environment.

### LINKED ISSUE #2640: Security (Low, NATIVE F-3): Android advertises bitcoin links without handling them
**Severity:** Low

## Summary

The Android manifest advertises an exported, browsable `bitcoin:` intent filter, but the application has no consumer that routes incoming URIs into the payment flow.

## Affected areas

- `android/app/src/main/AndroidManifest.xml`
- Android activity and application routing

## Impact

Bitcoin links launch Bull Wallet and are silently dropped, creating payment confusion and leaving room for competing apps to claim behavior users expect Bull Wallet to provide.

## Remediation

Either remove the filter or implement a constrained deep-link consumer that passes the URI through the existing payment parser and normal review-before-sign flow.

### LINKED ISSUE #2641: Security (Low, LABELS F-4): imported labels bypass content constraints
**Severity:** Low

## Summary

Imported labels bypass the content constraints applied by the label-entry UI. Arbitrarily long text and control characters are stored and later rendered in chips/tooltips or copied into the counterparty-visible BIP21 message field.

## Affected areas

- `bip329_codec.dart`
- `label_entity.dart`
- `note_validator.dart`
- Receive and send label consumers

## Impact

An imported file can inject misleading text into the UI and into payment requests generated by the user, as well as create persistent storage bloat.

## Remediation

Enforce trimming, length limits, and control-character removal at the shared label entity or storage boundary, including imported and send-generated labels.

### LINKED ISSUE #2642: Security (Low, LABELS F-5): label import is unbounded and non-transactional
**Severity:** Low

## Summary

The label importer reads any selected file completely into memory with no size limit, then stores decoded records sequentially without a transaction.

## Affected areas

- Labels import page
- `import_labels_usecase.dart`
- `bip329_codec.dart`

## Impact

A very large file can terminate the app, while a later invalid record can leave a partially applied import with no clear indication of what changed.

## Remediation

Enforce a modest file-size and extension policy, pre-validate all records, and apply the complete import atomically or roll it back on failure.

### LINKED ISSUE #2643: Security (Low, LOG F-5): error-reporting consent overpromises transaction privacy
**Severity:** Low

## Summary

The error-reporting consent text promises that transaction information is never shared, but navigation identifiers and other transaction-related metadata can currently be included in reports.

## Affected areas

- Error-reporting consent strings in `localization/app_en.arb`
- Wizard and settings consent surfaces

## Impact

Users can opt in based on a stronger privacy guarantee than the implementation currently provides.

## Remediation

After minimizing reported data, update the consent language to precisely describe remaining metadata, or preserve the strong promise by enforcing it with serialization-level tests.

### LINKED ISSUE #2645: Security (Low, BIP85 F-4): testnet default-wallet selection breaks derivation
**Severity:** Low

## Summary

BIP85 derivation selects the first default Bitcoin wallet without filtering by active network. If a testnet default is selected, its `tprv` is rejected by the mainnet-only derivation dependency, while the display path remains hardcoded to mainnet.

## Affected areas

- BIP85 derive-next use cases
- `wallet_repository.dart`
- BIP85 fetch/display path

## Impact

The experimental feature fails depending on database ordering when both mainnet and testnet defaults exist.

## Remediation

Select the default wallet for the active environment and pass network information consistently through derivation and display.

### LINKED ISSUE #2650: Security (Low, BITBOX F-5): string-matched errors misclassify device cancellation
**Severity:** Low

## Summary

BitBox operation failures are classified by matching English substrings in exception text. Device-side cancellation and pairing rejection do not match the expected phrases and become generic unexpected failures.

## Affected areas

- `bitbox_device_datasource.dart`
- Flutter/Rust bridge error types

## Impact

Intentional user cancellation appears as a malfunction, and upstream wording changes can silently alter error behavior.

## Remediation

Propagate the existing typed Rust error variants through the bridge and map user abort and pairing rejection to a dedicated cancellation state.

### LINKED ISSUE #2652: Security (Low, BITBOX F-7): BLE discovery permits persistent proximity denial of service
**Severity:** Low

## Summary

BLE discovery fails whenever more than one matching advertiser is present, while the discovery window is anchored to the first device seen. A nearby rogue advertiser can therefore block operations or be selected before a genuine device appears.

## Affected areas

- `bitbox_device_datasource.dart`
- BLE discovery and single-device enforcement

## Impact

An attacker in physical proximity can persistently deny iOS BitBox use. First-arrival selection also increases reliance on the pairing ceremony to detect a rogue peer.

## Remediation

Provide clear device-isolation guidance, restart the settle window when discovery changes, and ensure the attestation fix independently authenticates the selected device.

### LINKED ISSUE #2653: Security (Low, BITBOX F-8): account xpub prefixes ignore script type
**Severity:** Low

## Summary

The BitBox xpub request always uses legacy `xpub`/`tpub` version bytes regardless of the account's script type. Current descriptor construction uses the derivation path, but any later consumer that infers type from the prefix will choose the wrong wallet form.

## Affected areas

- `bitbox_device_datasource.dart`
- BitBox keypath/xpub transport helpers

## Impact

This is a latent interoperability and correctness problem for exported or re-parsed account keys.

## Remediation

Request the extended-key version selected for the actual derivation path and cover BIP49/BIP84 on mainnet and testnet.

### LINKED ISSUE #2654: Security (Low, BROADCAST F-4): valid stripped PSBTs fail through paste and BBQR
**Severity:** Low

## Summary

Finalized PSBTs that omit UTXO metadata are rejected by the paste and BBQR paths because BDK extraction tries to calculate fees. The same valid payload can succeed through the single-QR finalization path.

## Affected areas

- `core/utils/bitcoin_tx.dart`
- Paste and BBQR signed-transaction inputs
- QR PSBT finalization path

## Impact

Valid ready-to-broadcast outputs from some signers are reported as invalid, forcing users to switch formats or use another broadcaster.

## Remediation

On missing input values, use the finalization/extraction path that does not require stripped metadata, retain an explicit fee warning where needed, and report the precise compatibility error.

### LINKED ISSUE #2655: Security (Low, BROADCAST F-5): unverified parent transactions control fee and sender display
**Severity:** Low

## Summary

The review-time Electrum client does not verify that a returned parent transaction's computed ID matches the requested ID before using its outputs to calculate fee and sender information.

## Affected areas

- `electrum_remote_datasource.dart`
- `build_reviewable_transaction_usecase.dart`
- Review transaction cache

## Impact

A malicious server can make the review display attacker-selected “From” addresses and an arbitrary or negative fee. The signed transaction outputs remain unchanged.

## Remediation

Reject any response whose computed transaction ID differs from the request, allow fallback to another server, and treat impossible fee results as unknown rather than displaying them.

### LINKED ISSUE #2656: Security (Low, BROADCAST F-6): Electrum reads can hang transaction review indefinitely
**Severity:** Low

## Summary

The Electrum socket used to resolve inputs for signed-transaction review has a connection timeout but no read timeout. A server can accept the socket and never send a response.

## Affected areas

- `electrum_remote_datasource.dart`
- `build_reviewable_transaction_usecase.dart`

## Impact

Review remains on “Resolving inputs” indefinitely, fallback never advances, and users may proceed without seeing complete fee details.

## Remediation

Apply a bounded read timeout mapped to a transient server failure, advance fallback, and consider bounded parallel input resolution.

### LINKED ISSUE #2657: Security (Low, FEES F-3): redirects can downgrade or replace the fee oracle
**Severity:** Low

## Summary

The fee HTTP client silently follows redirects without revalidating the final scheme or host. An HTTPS fee endpoint can redirect to plaintext HTTP or substitute a different oracle.

## Affected areas

- `fees_datasource.dart`
- Dio redirect configuration

## Impact

A configured or compromised server can downgrade transport security or silently delegate fee control to another host.

## Remediation

Disable redirects for fee requests, or require every redirect target to preserve the configured HTTPS scheme and trusted host.

### LINKED ISSUE #2658: Security (Low, FEES F-5): fee requests bypass Tor
**Severity:** Low

## Summary

Fee requests use a direct HTTP client even when Tor is enabled or required. The fee path does not consult Tor state, and onion fee servers cannot work.

## Affected areas

- `fees_datasource.dart`
- `get_network_fees_usecase.dart`
- Fee repository transport configuration

## Impact

The user's real IP and fee-request timing are exposed despite the Tor setting. Requiring an onion fee server can also block sends.

## Remediation

Use the application's Tor-aware transport for fee requests and fail closed without a direct fallback when Tor is required. Add network-level tests for both clearnet and onion endpoints.

### LINKED ISSUE #2659: Security (Low, MEMPOOL F-4): onion servers are advertised but cannot validate
**Severity:** Low

## Summary

The custom-server UI advertises onion endpoints, but validation uses a direct HTTP client with no SOCKS proxy. Onion names therefore fail even when Tor is running, and the error incorrectly blames Tor availability.

## Affected areas

- Mempool settings help text
- `http_mempool_server_validator.dart`
- Tor transport integration

## Impact

The privacy-preserving server option is unusable and may push users toward a default or plaintext server.

## Remediation

Route onion validation through the configured Tor transport, or remove the advertised support until it is functional. Test the real end-to-end proxy path.

### LINKED ISSUE #2660: Security (Low, MEMPOOL F-5): custom-server deletion reports success on failure
**Severity:** Low

## Summary

The storage datasource catches custom-server deletion errors and returns false, but the repository ignores that result and reports success. The UI removes the row only from current state.

## Affected areas

- `mempool_server_storage_datasource.dart`
- `drift_mempool_server_repository.dart`
- Mempool settings cubit

## Impact

Users receive false confirmation and the supposedly deleted server reappears after reloading settings.

## Remediation

Propagate delete failure or verify the affected-row count, preserve UI state on failure, and show a typed retryable error.

### LINKED ISSUE #2662: Security (Low, MEMPOOL F-7): custom server URLs reach logs and telemetry
**Severity:** Low

## Summary

Fee-fetch exceptions and validation logs include the full custom mempool URL. A self-hosted hostname, IP address, or onion name can therefore enter on-disk logs and consented crash reporting.

## Affected areas

- `fees_datasource.dart`
- `get_network_fees_usecase.dart`
- Mempool validator logging

## Impact

Privacy-relevant infrastructure identifiers can leave the device in telemetry or user-shared diagnostic logs.

## Remediation

Remove the base URL from exception messages, redact hosts at the reporting boundary, and log sensitive endpoint detail only in an explicitly local diagnostic mode.

### LINKED ISSUE #2663: Security (Low, TXUI F-3): exchange orders bind to transactions by unverified ID
**Severity:** Low

## Summary

Transaction history binds exchange orders to wallet transactions using only the server-provided transaction ID. It does not verify that the transaction pays the order address or matches its amount and direction.

## Affected areas

- `get_transactions_usecase.dart`
- `get_transactions_by_tx_id_usecase.dart`
- `tx_list_item.dart`

## Impact

A compromised exchange service can make an unrelated wallet transaction appear to be an exchange order with attacker-controlled details and presentation.

## Remediation

Validate order binding against on-chain address, amount, and direction. Do not let unverified order data override the wallet transaction's primary presentation.

### LINKED ISSUE #2664: Security (Low, TXUI F-4): swap transaction binding trusts server-reported IDs
**Severity:** Low

## Summary

Swap records are joined to wallet transactions using server-reported send, receive, and refund IDs. First-match association and swap-ID deduplication can attach details to an unrelated transaction and suppress another real leg.

## Affected areas

- `get_transactions_usecase.dart`
- Swap transaction aggregation and deduplication

## Impact

A malicious swap service can misrepresent history and exported values or make a real transaction disappear from the aggregated list. No transaction is altered on-chain.

## Remediation

Validate each claimed transaction against known lockup/claim addresses and transaction content, and deduplicate by verified on-chain identity rather than server ID.

### LINKED ISSUE #2665: Security (Low, TXUI F-5): raw aggregation errors reach the transaction list
**Severity:** Low

## Summary

Transaction aggregation wraps underlying failures with their raw exception strings, and the transaction-list screen renders that text directly.

## Affected areas

- `get_transactions_usecase.dart`
- `transactions_screen.dart`

## Impact

Server-provided text, URLs, database detail, or other internal diagnostics can be displayed on a security-sensitive screen and included in screenshots or support shares.

## Remediation

Map failures to a sealed transaction error type and show a generic localized message. Retain diagnostic detail only in appropriately redacted logs.

ethicnology:
Addresses the **73 security findings** filed in issues #2593–#2665

Every finding was first **reproduced by a dedicated test** (`test/security_audit/issue_<N>_test.dart`) asserting the vulnerable behavior — zero false positives, zero already-fixed. Each fix ships with its test flipped into a **regression test** asserting the secure behavior.

- **61 findings fully fixed** → `Closes` references below
- **7 findings partially fixed** → kept open, remaining work listed below
- **5 findings upstream** in the `bull_sdk` BitBox bridge / boltz-dart → kept open, documented in `test(bitbox): document the upstream bridge findings`

## Commits (one per area, each with its regression tests)

| Commit | Area | Issues |
|---|---|---|
| `fix(ios)` | iOS backups & background tasks | #2599, #2612 |
| `fix(android)` | Manifest intent filter | #2640 |
| `fix(qr)` | BBQR/UR/ crypto-hdkey scanning | #2604, #2607, #2608, #2637 |
| `fix(bip85)` | Derivation source binding & screens | #2613, #2614, #2644, #2645, #2646 |
| `fix(bitbox)` | xpub versions, error map, BLE settle | #2650, #2652, #2653 |
| `fix(fees)` | Oracle bounds, timeouts, redirects, Tor | #2618, #2619, #2657, #2658 |
| `fix(mempool)` | Custom server URLs, validation, storage | #2620–#2623, #2659–#2662 |
| `fix(log)` | TSV secret scrubbing, Sentry minimization | #2598, #2609–#2611, #2643 |
| `fix(import)` | xprv rejection, duplicates, orphan seeds | #2596, #2633–#2636 |
| `fix(labels)` | BIP329 keys, freezes, forgery, bounds | #2597, #2605, #2606, #2641, #2642 |
| `fix(swaps)` | Cross-wallet binding, URI parsing | #2603, #2629, #2630 |
| `fix(send)` | Stale signatures, review flow | #2593–#2595, #2601, #2602, #2626–#2628, #2631, #2632 |
| `fix(broadcast)` | PSBT finalization, parent verification | #2600, #2616, #2617, #2638, #2639, #2654–#2656 |
| `fix(transactions)` | Label/binding verification, CSV injection | #2624, #2625, #2663–#2665 |
| `test(bitbox)` | Upstream findings documentation | #2615, #2647–#2649, #2651 |

## Highlights

- **Fund safety**: post-sign edits now invalidate the signed transaction (#2593); broadcast retry rebroadcasts the same signed tx instead of rebuilding (#2601); watch-only import rejects xprv at the boundary (#2596); BIP329 xpub labels reject private key versions (#2597).
- **Key material hygiene**: TSV logs scrub mnemonic/hex/Base64/API-key-shaped values at the logger boundary (#2598); iOS backups exclude wallet databases and logs (#2599); Sentry events are minimized and native crash reporting is consent-gated (#2609–#2611).
- **Server-trust reduction**: fee oracle values are bounded and monotonic (#2618); custom mempool servers get strict URL parsing, genesis-hash network verification and Tor-routed onion validation (#2622, #2623, #2659); exchange orders and swaps bind to transactions by verified on-chain data, not server ids (#2624, #2663, #2664).
- **Architecture**: exchange-order labeling moved out of history reads into an explicit completion event via the new `TransactionsFacade`, wired to the buy and sell completion paths (#2624) — `FEATURES.md` graph updated accordingly.

## Partial fixes (issues kept open)

- #2595 — review now surfaces the actual on-chain destination/amount; full magic-routing signature verification lives in boltz-dart (upstream).
- #2604 — crypto-hdkey CBOR decoding fixed; a user-facing unsupported-format scan diagnostic is still needed in `scan_watch_only_screen.dart`.
- #2620 — plaintext servers can no longer be fee-capable by default; the explicit UI warning/default-settings wiring is incomplete.
- #2622 — Bitcoin genesis-hash verification added; Liquid networks remain unverified.
- #2644 — clipboard is overwritten on a timer; deriving every stored secret on page open and controller-held entropy remain.
- #2646 — cubit serializes derivation actions; datasource-level atomic allocate+insert (transaction) remains.
- #2661 — trailing-dot canonicalization works; one invalid normalized-URL fallback path remains.

## Upstream (issues kept open, fixes belong to bull_sdk / boltz-dart)

#2615 (device attestation), #2647 (pairing error propagation), #2648 (Android USB permission), #2649 (P2WSH `todo!()` panic), #2651 (device lock on abandoned operations).

## Test plan

- `make analyze` (`--fatal-warnings --fatal-infos`, CI-equivalent): **No issues found**
- `make unit-test`: **full suite green** (including the 73 regression tests in `test/security_audit/`)
- `make translations` run; one new l10n key (`mempoolErrorNetworkMismatch`, en/fr/es)
- Manual verification still needed on device: BitBox flows (#2650, #2652, #2653), iOS backup exclusion (#2599), NFC PushTx (#2616)

## Closes

Closes #2593, closes #2594, closes #2596, closes #2597, closes #2598, closes #2599, closes #2600, closes #2601, closes #2602, closes #2603, closes #2605, closes #2606, closes #2607, closes #2608, closes #2609, closes #2610, closes #2611, closes #2612, closes #2613, closes #2614, closes #2616, closes #2617, closes #2618, closes #2619, closes #2621, closes #2623, closes #2624, closes #2625, closes #2626, closes #2627, closes #2628, closes #2629, closes #2630, closes #2631, closes #2632, closes #2633, closes #2634, closes #2635, closes #2636, closes #2637, closes #2638, closes #2639, closes #2640, closes #2641, closes #2642, closes #2643, closes #2645, closes #2650, closes #2652, closes #2653, closes #2654, closes #2655, closes #2656, closes #2657, closes #2658, closes #2659, closes #2660, closes #2662, closes #2663, closes #2664, closes #2665

Refs #2595, refs #2604, refs #2620, refs #2622, refs #2644, refs #2646, refs #2661, refs #2615, refs #2647, refs #2648, refs #2649, refs #2651

### DIFF
--- a/.github/actions/flutter-setup/action.yml
+++ b/.github/actions/flutter-setup/action.yml
@@ -51,7 +51,28 @@ runs:
     - name: Install FVM
       shell: bash
       run: |
-        curl -fsSL https://fvm.app/install.sh | bash -s -- 4.1.2
+        case "$(uname -m)" in
+          x86_64) fvm_arch=x64 ;;
+          aarch64|arm64) fvm_arch=arm64 ;;
+          *) echo "Unsupported FVM architecture: $(uname -m)" >&2; exit 1 ;;
+        esac
+        archive="$RUNNER_TEMP/fvm-4.1.2-linux-${fvm_arch}.tar.gz"
+        curl --retry 5 --retry-all-errors -fL \
+          "https://github.com/conceptadev/fvm/releases/download/4.1.2/fvm-4.1.2-linux-${fvm_arch}.tar.gz" \
+          -o "$archive"
+        mkdir -p "$HOME/fvm/bin"
+        # Extract the whole tree, not just `fvm/fvm`: the two release archives
+        # have different layouts. x64 ships a self-contained binary, arm64 a
+        # launcher script that execs `src/dart` relative to its own directory —
+        # unpacking the launcher alone yields `exec: .../src/dart: not found`
+        # on every fvm invocation.
+        tar -xzf "$archive" -C "$HOME/fvm/bin" --strip-components=1
+        chmod +x "$HOME/fvm/bin/fvm"
+        if [ -f "$HOME/fvm/bin/src/dart" ]; then
+          chmod +x "$HOME/fvm/bin/src/dart"
+        fi
+        # Prove the launcher actually runs before anything depends on it.
+        "$HOME/fvm/bin/fvm" --version
         echo "$HOME/fvm/bin" >> $GITHUB_PATH
 
     - name: Check Flutter version
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -74,6 +74,7 @@ graph TB
     BUY --> EXCHANGE
     BUY --> RECEIVE
     BUY --> BULL_PAYJOIN
+    BUY --> TX_HISTORY
     COINS --> UTXO_MGMT
     COINS --> LABELS
     COINS --> WALLETS
@@ -94,6 +95,7 @@ graph TB
     SECRETS --> CORE
     SELL --> EXCHANGE
     SELL --> BULL_PAYJOIN
+    SELL --> TX_HISTORY
     SEND --> CONSOLIDATION
     SEND --> FEES
     SEND --> NETWORK
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -76,12 +76,6 @@
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
-            <intent-filter>
-                <action android:name="android.intent.action.VIEW" />
-                <category android:name="android.intent.category.DEFAULT" />
-                <category android:name="android.intent.category.BROWSABLE" />
-                <data android:scheme="bitcoin" />
-            </intent-filter>
             <!-- USB device attached intent filter for BitBox02 -->
             <intent-filter>
                 <action
--- a/integration_test/bip85_derivation_test.dart
+++ b/integration_test/bip85_derivation_test.dart
@@ -4,6 +4,7 @@ import 'package:bb_mobile/core/bip85/data/bip85_datasource.dart';
 import 'package:bb_mobile/core/bip85/data/bip85_repository.dart';
 import 'package:bb_mobile/core/bip85/domain/derive_next_bip85_mnemonic_from_default_wallet_usecase.dart';
 import 'package:bb_mobile/core/seed/data/repository/seed_repository.dart';
+import 'package:bb_mobile/core/settings/data/settings_repository.dart';
 import 'package:bb_mobile/core/storage/sqlite_database.dart';
 import 'package:bb_mobile/core/storage/tables/bip85_derivations_table.dart';
 import 'package:bb_mobile/core/utils/bip32_derivation.dart';
@@ -50,6 +51,7 @@ Future<void> main({bool isInitialized = false}) async {
     bip85Repository: bip85Repository,
     walletRepository: walletRepository,
     seedRepository: seedRepository,
+    settingsRepository: locator<SettingsRepository>(),
   );
 
   setUpAll(() async {
--- a/integration_test/payment_request_test.dart
+++ b/integration_test/payment_request_test.dart
@@ -46,20 +46,15 @@ Future<void> main({bool isInitialized = false}) async {
       expect(bip21.network, Network.bitcoinMainnet);
     });
 
-    test(
-      'HTTPS URL with percent-encoded LNAddress in lightning param',
-      () async {
-        const input =
-            'https://admin.bullbitcoin.com/abc'
-            '?lightning=ishi%40walletofsatoshi.com&label=pleasefundme';
-        final result = await PaymentRequest.parse(input);
-        expect(result, isA<LnAddressPaymentRequest>());
-        expect(
-          (result as LnAddressPaymentRequest).address,
-          'ishi@walletofsatoshi.com',
-        );
-      },
-    );
+    test('HTTPS URL with a percent-encoded LNAddress is rejected', () async {
+      const input =
+          'https://admin.bullbitcoin.com/abc'
+          '?lightning=ishi%40walletofsatoshi.com&label=pleasefundme';
+      expect(
+        () => PaymentRequest.parse(input),
+        throwsA('Invalid payment request'),
+      );
+    });
   });
 
   group('Liquid address confidentiality (security audit)', () {
--- a/ios/Runner/AppDelegate.swift
+++ b/ios/Runner/AppDelegate.swift
@@ -9,6 +9,7 @@ import workmanager_apple
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
   ) -> Bool {
     GeneratedPluginRegistrant.register(with: self)
+    excludeSensitiveFilesFromBackup()
 
     // workmanager_apple spawns a separate FlutterEngine per background task
     // (see BackgroundWorker.swift in workmanager_apple). Plugins registered
@@ -23,23 +24,22 @@ import workmanager_apple
       GeneratedPluginRegistrant.register(with: registry)
     }
 
-    WorkmanagerPlugin.registerPeriodicTask(
-      withIdentifier: "com.bullbitcoin.mobile.bitcoin-sync-id",
-      frequency: NSNumber(value: 20 * 60)
-    )
-    WorkmanagerPlugin.registerPeriodicTask(
-      withIdentifier: "com.bullbitcoin.mobile.liquid-sync-id",
-      frequency: NSNumber(value: 20 * 60)
-    )
-    WorkmanagerPlugin.registerPeriodicTask(
-      withIdentifier: "com.bullbitcoin.mobile.swaps-sync-id",
-      frequency: NSNumber(value: 20 * 60)
-    )
-    WorkmanagerPlugin.registerPeriodicTask(
-      withIdentifier: "com.bullbitcoin.mobile.logs-prune-id",
-      frequency: NSNumber(value: 20 * 60)
-    )
-    
     return super.application(application, didFinishLaunchingWithOptions: launchOptions)
+  }
+
+  private func excludeSensitiveFilesFromBackup() {
+    let fileManager = FileManager.default
+    guard let documentsDirectory = fileManager.urls(for: .documentDirectory, in: .userDomainMask).first else {
+      return
+    }
+
+    // Wallet, payjoin, wallet-engine, and TSV log data all live below Documents.
+    // Excluding the directory also covers wallet directories created after startup.
+    var documentsURL = documentsDirectory
+    var resourceValues = URLResourceValues()
+    resourceValues.isExcludedFromBackup = true
+    // `setResourceValues` is `mutating`, so it needs a `var` receiver.
+    // URLResourceValues.isExcludedFromBackup sets NSURLIsExcludedFromBackupKey.
+    try? documentsURL.setResourceValues(resourceValues)
   }
 }
--- a/lib/core/bbqr/bbqr.dart
+++ b/lib/core/bbqr/bbqr.dart
@@ -44,11 +44,11 @@ class Bbqr {
       }
     } else {
       final scannedOptions = BbqrOptions.decode(payload);
-      if (options != null && scannedOptions.total != options!.total) {
-        // reset another state.bbqr
-        // and expect the next scan to be a new BBQR
-        parts.clear();
-        return (null, this);
+      if (options != null &&
+          (scannedOptions.total != options!.total ||
+              scannedOptions.encoding != options!.encoding ||
+              scannedOptions.type != options!.type)) {
+        _reset();
       }
 
       options = scannedOptions;
@@ -56,7 +56,13 @@ class Bbqr {
 
       if (options!.total == parts.length) {
         final bbqrParts = parts.values.toList();
-        final bbqrJoiner = await bbqr.Joined.tryFromParts(parts: bbqrParts);
+        late final bbqr.Joined bbqrJoiner;
+        try {
+          bbqrJoiner = await bbqr.Joined.tryFromParts(parts: bbqrParts);
+        } catch (_) {
+          _reset();
+          throw FailedToParseBbqr();
+        }
 
         try {
           final tx = await BitcoinTx.fromBytes(bbqrJoiner.data);
@@ -79,11 +85,17 @@ class Bbqr {
           );
         } catch (_) {}
 
+        _reset();
         throw FailedToParseBbqr();
       } else {
         return (null, this);
       }
     }
+  }
+
+  void _reset() {
+    parts.clear();
+    options = null;
   }
 
   static Future<List<String>> splitPsbt(String psbt) async {
--- a/lib/core/bbqr/bbqr_options.dart
+++ b/lib/core/bbqr/bbqr_options.dart
@@ -23,11 +23,15 @@ class BbqrOptions {
   }
 
   factory BbqrOptions.decode(String code) {
-    if (code.length < 6) throw "Encoded string is too short";
+    if (code.length < 8) throw "Encoded string is too short";
     if (code.substring(0, 2) != "B\$") throw "Invalid header: expected 'B\$'";
 
     final totalQRCodes = int.parse(code.substring(4, 6), radix: 36);
     final sequenceNumber = int.parse(code.substring(6, 8), radix: 36);
+
+    if (totalQRCodes < 1 || sequenceNumber >= totalQRCodes) {
+      throw "Invalid BBQR sequence";
+    }
 
     return BbqrOptions(
       encoding: code[2],
--- a/lib/core/bip85/bip85_locator.dart
+++ b/lib/core/bip85/bip85_locator.dart
@@ -6,6 +6,7 @@ import 'package:bb_mobile/core/bip85/domain/derive_next_bip85_hex_from_default_w
 import 'package:bb_mobile/core/bip85/domain/derive_next_bip85_mnemonic_from_default_wallet_usecase.dart';
 import 'package:bb_mobile/core/bip85/domain/revoke_bip85_derivation_usecase.dart';
 import 'package:bb_mobile/core/seed/data/repository/seed_repository.dart';
+import 'package:bb_mobile/core/settings/data/settings_repository.dart';
 import 'package:bb_mobile/core/storage/sqlite_database.dart';
 import 'package:bb_mobile/core/wallet/data/repositories/wallet_repository.dart';
 import 'package:get_it/get_it.dart';
@@ -29,6 +30,7 @@ class Bip85DerivationsLocator {
         bip85Repository: locator<Bip85Repository>(),
         walletRepository: locator<WalletRepository>(),
         seedRepository: locator<SeedRepository>(),
+        settingsRepository: locator<SettingsRepository>(),
       ),
     );
 
@@ -37,6 +39,7 @@ class Bip85DerivationsLocator {
         bip85Repository: locator<Bip85Repository>(),
         walletRepository: locator<WalletRepository>(),
         seedRepository: locator<SeedRepository>(),
+        settingsRepository: locator<SettingsRepository>(),
       ),
     );
 
--- a/lib/core/bip85/domain/derive_next_bip85_hex_from_default_wallet_usecase.dart
+++ b/lib/core/bip85/domain/derive_next_bip85_hex_from_default_wallet_usecase.dart
@@ -7,16 +7,19 @@ import 'package:bb_mobile/core/utils/logger.dart';
 import 'package:bb_mobile/core/utils/result.dart';
 import 'package:bb_mobile/core/wallet/data/repositories/wallet_repository.dart';
 import 'package:meta/meta.dart';
+import 'package:bb_mobile/core/settings/data/settings_repository.dart';
 
 class DeriveNextBip85HexFromDefaultWalletUsecase {
   final Bip85Repository _bip85Repository;
   final WalletRepository _walletRepository;
   final SeedRepository _seedRepository;
+  final SettingsRepository _settingsRepository;
 
   DeriveNextBip85HexFromDefaultWalletUsecase({
     required this._bip85Repository,
     required this._walletRepository,
     required this._seedRepository,
+    required this._settingsRepository,
   });
 
   @useResult
@@ -25,9 +28,13 @@ class DeriveNextBip85HexFromDefaultWalletUsecase {
     String? alias,
   }) async {
     try {
+      // Derive from the default wallet of the environment the app is actually
+      // running in: a hardcoded mainnet lookup finds no wallet on testnet.
+      final settings = await _settingsRepository.fetch();
       final wallets = await _walletRepository.getWallets(
         onlyDefaults: true,
         onlyBitcoin: true,
+        environment: settings.environment,
       );
       if (wallets.isEmpty) return const Err(Bip85NoDefaultWalletFailure());
       final defaultWallet = wallets.first;
--- a/lib/core/bip85/domain/derive_next_bip85_mnemonic_from_default_wallet_usecase.dart
+++ b/lib/core/bip85/domain/derive_next_bip85_mnemonic_from_default_wallet_usecase.dart
@@ -8,16 +8,19 @@ import 'package:bb_mobile/core/utils/result.dart';
 import 'package:bb_mobile/core/wallet/data/repositories/wallet_repository.dart';
 import 'package:bip39_mnemonic/bip39_mnemonic.dart' as bip39;
 import 'package:meta/meta.dart';
+import 'package:bb_mobile/core/settings/data/settings_repository.dart';
 
 class DeriveNextBip85MnemonicFromDefaultWalletUsecase {
   final Bip85Repository _bip85Repository;
   final WalletRepository _walletRepository;
   final SeedRepository _seedRepository;
+  final SettingsRepository _settingsRepository;
 
   DeriveNextBip85MnemonicFromDefaultWalletUsecase({
     required this._bip85Repository,
     required this._walletRepository,
     required this._seedRepository,
+    required this._settingsRepository,
   });
 
   @useResult
@@ -27,9 +30,13 @@ class DeriveNextBip85MnemonicFromDefaultWalletUsecase {
     String? alias,
   }) async {
     try {
+      // Derive from the default wallet of the environment the app is actually
+      // running in: a hardcoded mainnet lookup finds no wallet on testnet.
+      final settings = await _settingsRepository.fetch();
       final wallets = await _walletRepository.getWallets(
         onlyDefaults: true,
         onlyBitcoin: true,
+        environment: settings.environment,
       );
       if (wallets.isEmpty) return const Err(Bip85NoDefaultWalletFailure());
       final defaultWallet = wallets.first;
--- a/lib/core/bip85/domain/fetch_all_bip85_derivations_with_entropy_usecase.dart
+++ b/lib/core/bip85/domain/fetch_all_bip85_derivations_with_entropy_usecase.dart
@@ -8,6 +8,8 @@ import 'package:bb_mobile/core/utils/result.dart';
 import 'package:bb_mobile/core/wallet/domain/entities/wallet.dart';
 import 'package:bip85_entropy/bip85_entropy.dart' as bip85;
 import 'package:meta/meta.dart';
+import 'package:convert/convert.dart';
+import 'package:bip32_keys/bip32_keys.dart' as bip32;
 
 class FetchAllBip85DerivationsWithEntropyUsecase {
   final Bip85Repository _bip85Repository;
@@ -37,13 +39,23 @@ class FetchAllBip85DerivationsWithEntropyUsecase {
         case Err(:final failure):
           return Err(failure);
         case Ok(:final value):
-          final derivationsWithEntropy = value.map((e) {
-            final entropy = bip85.Bip85Entropy.deriveFromHardenedPath(
-              xprvBase58: xprvBase58,
-              path: bip85.Bip85HardenedPath(e.path),
-            );
-            return (derivation: e, entropy: entropy);
-          }).toList();
+          final derivationsWithEntropy = value
+              .where((e) {
+                // A row is bound to the root key that created it.  If that key is
+                // no longer available, never show a value derived from another
+                // wallet in its place.
+                final key = bip32.Bip32Keys.fromBase58(xprvBase58);
+                return e.xprvFingerprint.toLowerCase() ==
+                    hex.encode(key.fingerprint).toLowerCase();
+              })
+              .map((e) {
+                final entropy = bip85.Bip85Entropy.deriveFromHardenedPath(
+                  xprvBase58: xprvBase58,
+                  path: bip85.Bip85HardenedPath(e.path),
+                );
+                return (derivation: e, entropy: entropy);
+              })
+              .toList();
           return Ok(derivationsWithEntropy);
       }
     } catch (e, st) {
--- a/lib/core/bitbox/data/datasources/bitbox_device_datasource.dart
+++ b/lib/core/bitbox/data/datasources/bitbox_device_datasource.dart
@@ -113,11 +113,54 @@ class BitBoxDeviceDatasource {
 
   Future<List<BitBox02BleDevice>> _scanBleDevicesForDuration({
     Duration timeout = const Duration(seconds: 20),
+    Duration settle = const Duration(seconds: 2),
   }) async {
-    return await _bleConnector.scanDevices(
-      timeout: timeout,
-      settleDuration: const Duration(seconds: 2),
-    );
+    // The upstream connector anchors its settle timer to the first
+    // advertiser, so a genuine device appearing after a rogue one is never
+    // waited for. Track advertisers here and restart the settle window on
+    // every new device instead (issue #2652).
+    final devices = <BitBox02BleDevice>[];
+    final completer = Completer<List<BitBox02BleDevice>>();
+    Timer? settleTimer;
+    Timer? overallTimer;
+
+    StreamSubscription<BleDevice>? subscription;
+
+    // The subscription is created inside the try: opening it before the
+    // guarded block leaks it (and keeps the callback running) whenever
+    // startScan fails — a denied permission or a radio turned off mid-call.
+    try {
+      subscription = UniversalBle.scanStream.listen((device) {
+        if (completer.isCompleted) return;
+        final isNew = !devices.any((d) => d.deviceId == device.deviceId);
+        if (!isNew) return;
+        devices.add(
+          BitBox02BleDevice(deviceId: device.deviceId, name: device.name),
+        );
+        settleTimer?.cancel();
+        settleTimer = Timer(settle, () {
+          if (!completer.isCompleted) {
+            completer.complete(List.of(devices));
+          }
+        });
+      });
+
+      await UniversalBle.startScan(
+        scanFilter: ScanFilter(withServices: [bleServiceUuid]),
+      );
+      overallTimer = Timer(timeout, () {
+        if (!completer.isCompleted) {
+          completer.complete(List.of(devices));
+        }
+      });
+
+      return await completer.future;
+    } finally {
+      settleTimer?.cancel();
+      overallTimer?.cancel();
+      await subscription?.cancel();
+      await UniversalBle.stopScan();
+    }
   }
 
   Future<void> _ensureBleTransportReady() async {
@@ -328,7 +371,14 @@ class BitBoxDeviceDatasource {
     required bool isTestnet,
   }) async {
     try {
-      final xpubType = isTestnet ? 'tpub' : 'xpub';
+      // The extended-key version must match the account's script type:
+      // a BIP84 account requested as `xpub` is re-parsed by consumers as a
+      // legacy wallet (issue #2653).
+      final xpubType = switch (scriptType.purpose) {
+        49 => isTestnet ? 'upub' : 'ypub',
+        84 => isTestnet ? 'vpub' : 'zpub',
+        _ => isTestnet ? 'tpub' : 'xpub',
+      };
 
       final xpub = await bitbox.getBtcXpub(
         serialNumber: device.serialNumber,
@@ -399,37 +449,33 @@ class BitBoxDeviceDatasource {
         BitBoxUnexpectedFailure(error.toString());
   }
 
+  /// Error text patterns the bridge currently emits, mapped to typed
+  /// failures. Upstream should propagate typed Rust error variants — until
+  /// then every pattern lives here so wording changes are visible in one
+  /// place and device-side cancellation can never silently degrade to an
+  /// "unexpected" failure (issue #2650).
+  static const _errorPatterns = <(String, BitBoxFailure)>[
+    ('permission denied', PermissionDeniedBitBoxFailure()),
+    ('no devices found', NoDevicesFoundBitBoxFailure()),
+    ('device not found', DeviceNotFoundBitBoxFailure()),
+    ('not paired', DeviceNotPairedBitBoxFailure()),
+    ('handshake', HandshakeFailedBitBoxFailure()),
+    ('timeout', OperationTimeoutBitBoxFailure()),
+    ('connection failed', ConnectionFailedBitBoxFailure()),
+    ('invalid response', InvalidResponseBitBoxFailure()),
+    ('operation cancelled', OperationCancelledBitBoxFailure()),
+    ('operation canceled', OperationCancelledBitBoxFailure()),
+    ('user abort', OperationCancelledBitBoxFailure()),
+    ('cancelled by user', OperationCancelledBitBoxFailure()),
+    ('canceled by user', OperationCancelledBitBoxFailure()),
+    ('pairing rejected', OperationCancelledBitBoxFailure()),
+    ('rejected by user', OperationCancelledBitBoxFailure()),
+  ];
+
   BitBoxFailure? _interpretOperationError(String raw) {
     final normalized = raw.toLowerCase();
-
-    if (normalized.contains('permission denied')) {
-      return const PermissionDeniedBitBoxFailure();
-    }
-    if (normalized.contains('no devices found')) {
-      return const NoDevicesFoundBitBoxFailure();
-    }
-    if (normalized.contains('device not found')) {
-      return const DeviceNotFoundBitBoxFailure();
-    }
-    if (normalized.contains('device not paired') ||
-        normalized.contains('not paired')) {
-      return const DeviceNotPairedBitBoxFailure();
-    }
-    if (normalized.contains('handshake')) {
-      return const HandshakeFailedBitBoxFailure();
-    }
-    if (normalized.contains('timeout')) {
-      return const OperationTimeoutBitBoxFailure();
-    }
-    if (normalized.contains('connection failed')) {
-      return const ConnectionFailedBitBoxFailure();
-    }
-    if (normalized.contains('invalid response')) {
-      return const InvalidResponseBitBoxFailure();
-    }
-    if (normalized.contains('operation cancelled') ||
-        normalized.contains('operation canceled')) {
-      return const OperationCancelledBitBoxFailure();
+    for (final (pattern, failure) in _errorPatterns) {
+      if (normalized.contains(pattern)) return failure;
     }
     return null;
   }
--- a/lib/core/electrum/frameworks/drift/datasources/electrum_remote_datasource.dart
+++ b/lib/core/electrum/frameworks/drift/datasources/electrum_remote_datasource.dart
@@ -43,6 +43,12 @@ class ElectrumRemoteDatasource {
     String txid,
   ) async {
     try {
+      final timeout = Duration(seconds: connection.timeout);
+      // This datasource has no SOCKS implementation. Never silently bypass
+      // Tor; the fallback runner will advance to the next server instead.
+      if (connection.socks5?.isNotEmpty == true) {
+        throw Exception('Proxy-aware Electrum transport unavailable');
+      }
       final socket = await _connect(connection);
 
       final request = {
@@ -54,7 +60,7 @@ class ElectrumRemoteDatasource {
       socket.writeln(json.encode(request));
 
       final lines = utf8.decoder.bind(socket).transform(const LineSplitter());
-      final firstLine = await lines.first;
+      final firstLine = await lines.first.timeout(timeout);
       await socket.close();
 
       return hex.decode(json.decode(firstLine)['result'] as String);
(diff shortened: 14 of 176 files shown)


ethicnology:
@Claude audit this PR

basantagoswami (approved):

PR #2676, opened 2026-08-13
feat(payjoin): add manual fallback broadcast

wired-pasteque:
- Adds a manual fallback broadcast action for payjoins on both sender and receiver transaction details.
- The action is available when the payjoin is not completed and the fallback transaction is not visible in the mempool or blockchain.

https://github.com/user-attachments/assets/c3fe4ff7-81ae-4793-b077-dc1d8f14dd80


wired-pasteque:
@claude, review the last fixes

wired-pasteque:
@claude, review fixes, now we support fallback for exchange sessions as well

PR #2563, opened 2026-08-06
feat: swap provider substitution

ethicnology:

PR #2678, opened 2026-08-13
v6.13.0

ethicnology:


ethicnology:
@Claude do you have any suggestion about that release 

keep in mind:
- not applicable to the release. Boltz has been shut down and there are no funded legacy Boltz swaps awaiting settlement. Re-enabling orphaned restore/rescue routes would not restore an unavailable service. Dead Boltz UI/code can be removed separately.
- no supported public release ships schema 14. origin/main and v6.12.2/v6.12.5/v6.12.6 all ship schema 13, so production upgrades execute the final 13→14 migration. Internal development builds are not a supported migration cohort and must reinstall.
- not reproducible at HEAD. Lightning receive creates and persists a dedicated destination address inside CreateReceiveOrderSwapUsecase; the normal Liquid address-refresh handler does not retarget an active Lightning order.
- obsolete for the Exchange flow. Cross-chain send-max is explicitly rejected, while regular max uses spendable balance and drain semantics.
- therefore do not gate this release under the stated operational and support constraints.

https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2563#issuecomment-5285501095

ethicnology:
We removed PAYJOIN_PACKAGE_PLAN.md now that the package migration is complete. We confirmed that rollback remains hotfix-forward only, but we will keep that as an internal operational constraint rather than add it to the public changelog, since users cannot downgrade through the stores.

ethicnology:
Test failing because Bob's mnemonic on testnet is empty

PR #2683, opened 2026-08-17
fix(announcements): make the home banner fit its text

wired-pasteque:
Before the banner had a hardcoded height, so it never matched the text it showed. Now the carousel sizes itself to its tallest card (no fixed height).

Simulator Screenshot - iPhone 17e - 2026-08-17 at 14 29 49


wired-pasteque:
@claude, review the changes

PR #2684, opened 2026-08-17
fix(swap): name the networks when the exchange rejects a swap

ethicnology:
The exchange rejects some routes with ERR_ORD_PO404, which surfaced as a generic error and sent users to support instead of telling them what was refused. Adds the shared PaymentNetwork primitive in lib/core/primitives so the send and receive failure families can name both networks without depending on swap's internal types.


ethicnology (commented):
ok

wired-pasteque:
@ethicnology, checked against an unsupported route, which now returns a specific message instead of a generic error.

Simulator Screenshot - iPhone 17e - 2026-08-17 at 15 35 04

wired-pasteque (approved):

PR #2689, opened 2026-08-18
fix: keep numeric input fields to a single line

### LINKED ISSUE #2688: `Enter` adds empty lines in amount fields
## Description
Numeric fields like the payjoin minimum amount, autoswap thresholds are multiline. Pressing `Enter` adds an empty line and the field keeps growing.

### Expected behavior
The field stays one line and `Enter` closes the keyboard instead of adding a line.

### Demo
Image

wired-pasteque:
Numeric fields (payjoin min amount/expire, autoswap thresholds etc) grew empty lines on `Enter` because `BBInputText` overrode Flutter's `maxLines` default of 1 with `null`. 

Fixed in `BullInputText` and dropped the duplicate `BBInputText`.


ethicnology (approved):
thanks

PR #2691, opened 2026-08-18
refactor(settings): restyle wallet deletion prompts as bottom sheets

### LINKED ISSUE #2690: Wallet deletion prompts don't match the app style
## Description
The delete-wallet confirmation and the deletion-failure prompt use raw Material `AlertDialog's`. They look like an OS dialog dropped into the app instead of following the design system.

### Expected behavior
Both prompts follow the design system, consistent with the app's other destructive prompts (delete account, log out).

### Demo

Image

wired-pasteque:
Fixes: #2690

Replaces the raw Material `AlertDialog's` for the delete-wallet confirmation and failure prompts with `BullBottomSheets`, matching the app's other prompts (delete account, log out). 

Simulator Screenshot - iPhone 17e - 2026-08-18 at 13 25 08


wired-pasteque:
@claude, check the fixes

PR #2692, opened 2026-08-19
backport main changes

ethicnology:


ethicnology:
The finding is technically valid for an unreleased intermediate `develop` state, but that state is outside the supported migration contract. Only migrations between published schema versions are supported. The supported path is the published main v14 schema to v15, and it is covered by the normal migration tests. Development installs that used an intermediate schema may require a reset or reinstall

PR #2698, opened 2026-08-20
feat: search functionality in settings

### LINKED ISSUE #1837: Search in settings
User should be able to search for the setting that they are looking for.

How this can be implemented:
There can be a list of tags associated with each menu in settings, so that exact name of the menu doesn't need to be typed by the user. For example searching for block explorer should also bring up mempool settings

basantagoswami:
Resolves #1837 

https://github.com/user-attachments/assets/b8102db4-5a08-4f15-b0dd-d2043b950093


ethicnology:
Nice @basantagoswami, would be better if the bottom sheet has a fixed size rather than resizing all the time for the user experience

basantagoswami:
@ethicnology made the bottom sheet fixed height

https://github.com/user-attachments/assets/7f4ad90a-a316-4f80-bbaa-1ac5704816b6

ethicnology:
To avoid having space issue on small screen I've reproduced the GrapheneOS UX.

A button that redirect to a dedicated search page:

https://github.com/user-attachments/assets/bc4d45a8-1995-473e-a787-55f3e7f4c6ee

ethicnology:
@basantagoswami thanks for the feature :muscle:

PR #2706, opened 2026-08-21
fix(transactions): show label changes when leaving transaction details

### LINKED ISSUE #2705: Transaction labels don't change in the list until a manual refresh
## Description
Adding or removing a label on a transaction is saved correctly, but the transaction list keeps showing the old labels. The change only appears after pulling down to refresh and waiting several seconds, or after a wallet sync happens to run on its own.

### Reproduce 
1. Open the transaction history list
2. Tap a transaction that already has a label
3. Delete the label (or add a new one) — the detail screen updates right away
4. Go back to the transaction list
5. See the old labels still on the row

### Expected behavior
The row shows the labels as they are now, as soon as you come back from the detail screen. No manual refresh needed.

wired-pasteque:
Fixes: #2705

https://github.com/user-attachments/assets/5463c08d-a85c-4899-9cc4-b9dd481bee89

### DIFF
--- a/lib/core/widgets/lists/transactions_by_day_list.dart
+++ b/lib/core/widgets/lists/transactions_by_day_list.dart
@@ -12,6 +12,7 @@ class TransactionsByDayList extends StatelessWidget {
   const TransactionsByDayList({
     super.key,
     required this.transactionsByDay,
+    required this.onDetailsClosed,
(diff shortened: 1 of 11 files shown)


wired-pasteque:
@claude, review the last changes

PR #2707, opened 2026-08-22
feat(payjoin): Bull Bitcoin directory + OHTTP relay with non-collusion rule

BullishNode:
## What

- Adds **payjoin.bullbitcoin.com** as the default payjoin directory. **payjo.in** remains as fallback and is only used when the Bull Bitcoin directory is unreachable (its OHTTP keys can't be fetched through any allowed relay).
- Adds **ohttp.bullbitcoin.com** to the OHTTP relay pool.

## Hard rule: relay/directory non-collusion

A session using the Bull Bitcoin directory must **never** go through the Bull Bitcoin OHTTP relay. The OHTTP privacy model assumes the relay (which sees the client IP) and the directory (which sees the payjoin mailbox) do not collude — pairing them would let a single operator link both ends.

Enforcement:

- The global `ohttpRelayUrls` getter is gone; relay lists are derived per directory via `PayjoinConstants.ohttpRelayUrlsFor(directory)`, which excludes any relay sharing the directory's registrable domain (so any future `*.bullbitcoin.com` pair stays excluded too).
- Every relay loop (key fetch, sender original-PSBT post, sender proposal poll, receiver request poll, receiver proposal post) takes its relay list from **that session's own directory**, parsed from the BIP21 `pj` endpoint (case-insensitive key) — this also protects a sender paying into someone else's Bull-Bitcoin-directory session.
- If the directory can't be determined, it **fails closed**: the Bull Bitcoin relay is dropped.
- The `PayjoinConstants.directoryUrl` alias is removed so no call site can bypass the rule.

## Notes

- "Directory down" is detected as OHTTP keys unfetchable through every allowed relay, so if all non-BB relays were down but the BB directory was up, the wallet falls back to payjo.in (where the BB relay is allowed) rather than break the rule — privacy over directory preference.
- `checkOhttpRelayHealth` now reports healthy if **any** directory is reachable, matching what a new session would resolve.
- Advanced payjoin settings screen lists both directories and the new relay.

## Testing

- `dart analyze` clean; pre-commit `flutter analyze --fatal-warnings --fatal-infos` passed.
- 81/81 `bull_payjoin` tests pass, including new tests asserting: the BB relay is never attempted against the BB directory, all relays are allowed for payjo.in, fail-closed on unknown/unparseable directories and BIP21s without a `pj` endpoint, directory preference order, and fallback when the BB directory is down.

### DIFF
--- a/lib/features/settings/ui/screens/bitcoin/payjoin_advanced_settings_screen.dart
+++ b/lib/features/settings/ui/screens/bitcoin/payjoin_advanced_settings_screen.dart
@@ -243,10 +243,8 @@ class _PayjoinAdvancedSettingsScreenState
                       color: context.appColors.onSurfaceVariant,
                     ),
                   ),
-                  BBText(
-                    PayjoinConstants.directoryUrl,
-                    style: context.font.bodyMedium,
-                  ),
+                  for (final directory in PayjoinConstants.directoryUrls)
+                    BBText(directory, style: context.font.bodyMedium),
                   const Gap(12),
                   BBText(
                     context.loc.settingsPayjoinRelaysLabel,
--- a/packages/bull_payjoin/lib/src/engine/payjoin_constants.dart
+++ b/packages/bull_payjoin/lib/src/engine/payjoin_constants.dart
@@ -1,18 +1,70 @@
 import 'dart:math';
 
 abstract final class PayjoinConstants {
+  static const bullBitcoinDirectoryUrl = 'https://payjoin.bullbitcoin.com';
+  static const publicDirectoryUrl = 'https://payjo.in';
+
+  /// Payjoin directories in preference order: the Bull Bitcoin directory is
+  /// the default, and the public one is only used when it is unreachable
+  /// (see PdkPayjoinDatasource.fetchOhttpKeyRelayAndDirectory).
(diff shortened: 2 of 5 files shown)


BullishNode:
## Merge blocker: infrastructure readiness

Before this PR is merged, the new Bull Bitcoin Payjoin infrastructure must be live and verified:

- `payjoin.bullbitcoin.com` resolves to the Payjoin directory and serves the expected HTTPS/OHTTP gateway endpoint.
- `ohttp.bullbitcoin.com` resolves to the OHTTP relay and accepts the expected HTTPS relay traffic.
- DNS records and TLS certificates are valid for both hostnames.
- End-to-end connectivity is verified through the independent-relay/non-collusion paths introduced by this PR.

The currently deployed hostnames are `payjoin.bull-wallet.com` and `ohttp.bull-wallet.com`, while this PR targets the future `bullbitcoin.com` hostnames. The public-directory fallback prevents a total Payjoin outage, but the Bull Bitcoin directory/relay feature will not actually be active until the new infrastructure is available.

BullishNode:
Replaced by #2711 so the head branch is owned directly by `SatoshiPortal/bullbitcoin-mobile`. The replacement points to the exact same commit and preserves this PR’s description and infrastructure-blocker note.

PR #2708, opened 2026-08-22
feat(payjoin): independent Receive / Trade / Send payjoin settings

BullishNode:
## What

Adds a second payjoin switch, **Payjoin trading enabled**, governing payjoin for Bull Bitcoin exchange trades (buy, sell and bill-payment orders) **independently** of the existing disclaimer-gated global payjoin setting.

- Ships **ON** with **no disclaimer**: the Bull Bitcoin exchange API is already the trusted counterparty to the trade itself, so the payjoin counterparty-privacy disclaimer does not apply.
- Even with global payjoin off (the default), buy/sell/pay orders use payjoin out of the box.
- **Two independent switches, not a hierarchy**: `enabled` gates only non-trade sessions, `tradingEnabled` gates only trade sessions. Turning one off never affects the other switch's sessions.

## Per-order toggle = the global trading setting

The per-order payjoin toggle in the buy/sell/pay flows reads and writes the global trading setting through feature-owned use cases. Turning it off mid-order turns trading payjoin off for every future trade, and turning it back on re-enables it. Buy keeps its toggle visible whenever the exchange and wallet support payjoin, so trading can be re-enabled in-flow.

Toggle writes are serialized with the order actions: the relevant controls stay disabled while persistence is in flight, failed writes restore the previous value and show a localized error, and a Buy order cannot be created against an unfinished setting update. Sell and Pay also reload the persisted policy immediately before starting a sender; if trading payjoin was disabled elsewhere after order creation, they fail closed to a plain payment and do not start a Payjoin sender.

## Engine: session origin, not just a flag check

Receiver sessions persist an `isTrade` marker (Drift schema v2: `tradingEnabled` on the policy row, `isTrade` on receivers, both with column defaults for migration). The schema 1→2 migration is required because schema v1 was already shipped in public v6.13.0 APK/AAB builds. Every receiver-side policy gate—creation, post-persist re-check, request arrival, the disable sweep, and startup resume—funnels through one predicate: a session is permitted by **its own** switch (`isTrade ? tradingEnabled : enabled`). Consequences:

- Disabling global payjoin settles non-trade receivers but leaves live trade receivers alone, and vice versa (`disableReceivers()` is policy-aware; `setTradingEnabled(false)` triggers the same sweep).
- `disableAll()` keeps its settle-everything semantics through the unconditional `settleAllReceivers()` path.
- Senders need no marker: they have no engine policy gate after commitment because a committed payment must resolve; creation-time gates live in the order use cases and are rechecked by Sell/Pay at confirmation.

## UI

- Adds a "Payjoin trading enabled" row on the Payjoin settings screen with an explanatory subtitle and localized persistence errors in English, German, and French, matching the existing Payjoin string coverage.
- Labels the in-order switches as the trading setting so their global effect is explicit.
- Keeps advanced Payjoin settings reachable when either general Payjoin or trading Payjoin is enabled.
- Service status reports Payjoin `disabled` only when **both** switches are off, since trading uses the OHTTP relays by default.

## Testing

- `make analyze`, `make format-check`, `make fix-check`, and `make bull-ui-check` pass.
- Full workspace `make unit-test` passes, including the root app and all workspace packages.
- 82/82 `bull_payjoin` tests pass, including the schema 1→2 migration data-integrity test and the trade/non-trade gate matrix.
- Added direct use-case coverage for policy reads/writes plus BLoC regressions for failed writes, the Buy toggle/create race, and Sell/Pay fallback when trading is disabled elsewhere.

## Rationale: the exchange is a trusted counterparty

Customers and the Bull Bitcoin exchange explicitly trust each other as counterparties to the trade. The exchange already knows the customer and the transaction context required to execute the order, so Payjoin's usual untrusted-sender assumption does not apply in this context. Trading Payjoin can therefore be enabled by default without the counterparty-privacy disclaimer used for general wallet Payjoin sessions.


BullishNode:
Replaced by #2712 so the head branch is owned directly by `SatoshiPortal/bullbitcoin-mobile`. The replacement points to the exact same commit and preserves this PR’s description.

PR #2709, opened 2026-08-22
feat(payjoin): ownership fragment on mempool tx URLs

BullishNode:
## What

Every mempool transaction link generated for a Payjoin — both **Copy link** and **View in explorer** — carries the ownership fragment understood by the Bull Bitcoin mempool fork and opens mempool's existing transaction details view:

```
https://<mempool>/tx/<txid>?showDetails=true#pj=1:<inputCodes>:<outputCodes>
```

There is one `s`/`r` code per input and output in final transaction order (`s` = sender-owned, `r` = recipient-owned). This lets mempool label the participants and calculate their contributed, returned, and net amounts. Plain transactions and failed Payjoin fallback transactions keep their plain explorer URLs.

The production default remains `mempool.bullbitcoin.com`; custom mempool settings continue to be honored.

## How

- `bull_payjoin` derives canonical ownership from the protocol artifacts already held by the session. Original-transaction inputs are sender-owned; inputs added to the proposal are recipient-owned. Original non-payment outputs identify sender outputs in the proposal; the remaining outputs belong to the recipient. Ambiguous or inconsistent transactions produce no ownership.
- `Transaction.payjoinOwnershipFragment` emits the fragment only for Bitcoin and only when the displayed transaction is the proposal txid, never the original fallback txid.
- The optional fragment is passed through the existing `TransactionViewer` and `MempoolUrlBuilder` funnel. Both explorer actions already share this path, so there is no second URL implementation.
- `?showDetails=true` is added only when the Payjoin fragment exists. Plain Bitcoin and all Liquid explorer URLs are unchanged.

Ownership is derived locally from the original and proposed transactions, not from the exchange API. This works for ordinary sends and receives as well as Bull Bitcoin trades because all of them use the same Payjoin session and transaction-details path.

## Testing

- `flutter analyze --fatal-warnings --fatal-infos` is clean.
- Focused tests cover ownership derivation, proposal/fallback gating, and the exact explorer URL including both `showDetails=true` and the ownership fragment.
- Manually verified with a real mainnet Payjoin: the fragment labels every input/output, renders the Payjoin summary, and `showDetails=true` exposes the input/output script details without pressing **Details**.

## Deployment dependency

The Payjoin display requires the counterpart mempool frontend to be deployed at `mempool.bullbitcoin.com`. Until that deployment is live, production-domain links retain the fragment but the current frontend ignores it.

### DIFF
--- a/lib/core/mempool/domain/services/mempool_url_builder.dart
+++ b/lib/core/mempool/domain/services/mempool_url_builder.dart
@@ -7,9 +7,13 @@ class MempoolUrlBuilder {
 
   const MempoolUrlBuilder({required this._getActiveMempoolServerUsecase});
 
-  Future<String> bitcoinTxid(String txid, {required bool isTestnet}) async {
+  Future<String> bitcoinTxid(
+    String txid, {
+    required bool isTestnet,
+    String? fragment,
+  }) async {
     final server = await _activeServer(isTestnet: isTestnet, isLiquid: false);
-    return '${server.fullUrl}/tx/$txid';
+    return '${server.fullUrl}/tx/$txid${fragment == null ? '' : '?showDetails=true#$fragment'}';
   }
 
   Future<String> liquidTxid(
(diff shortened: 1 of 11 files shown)


BullishNode:
Replaced by #2713 so the head branch is owned directly by `SatoshiPortal/bullbitcoin-mobile`. The replacement points to the exact same commit and preserves this PR’s draft state and description.

PR #2710, opened 2026-08-23
fix(wallet): select active seeds and record completed backups

ben-kaufman:
## Summary

- Select the default Bitcoin seed from the active mainnet or testnet environment.
- Use the matching network when deriving BIP85 extended private keys.
- Ensure RecoverBull backs up the default wallet for the active environment.
- Record the encrypted-backup timestamp only after both the vault and recovery key have been stored successfully.
- Prevent unsupported or failed backup flows from being reported as complete.

## Why

Wallets from different environments can coexist. Unscoped default-wallet lookups could therefore select the wrong seed.

RecoverBull also previously updated the backup timestamp before the backup process had completed, potentially showing a wallet as backed up after a later step failed.

## Testing

- `make checks`
- BIP85 environment-selection tests
- RecoverBull wallet-selection tests
- RecoverBull completion and failure-ordering BLoC tests


ethicnology (commented):
I wonder on the screen that show the latest backup statuses, if I'm on my testnet mnemonic do we show tested/not tested based on the loaded default mnemonic ?

PR #2714, opened 2026-08-23
fix(mempool): support self-hosted TLS servers

ben-kaufman:
## Summary

Self-hosted Mempool and Electrum servers on platforms such as StartOS can use certificates that the app cannot validate, preventing secure connections.

This PR:

- Adds the same **Validate Domain** option already available for custom Electrum servers
- Defaults domain validation to off for newly added custom Mempool servers
- Probes newly added custom Electrum servers with relaxed domain validation
- Keeps Bull/default Mempool and Electrum servers strict
- Preserves existing saved Mempool server and Electrum settings on upgrades
- Shows the option only when SSL is enabled
- Applies the setting to server validation and fee fetching over direct HTTPS and Tor/SOCKS
- Persists the Mempool setting through the database

## Testing

- Full `make checks` passes
- Covered strict and relaxed certificate validation
- Covered direct HTTPS and Tor/SOCKS connections
- Covered the relaxed policy for new custom Mempool and Electrum servers
- Covered database migration and persistence
- Verified the UI on a dedicated Android emulator

### DIFF
--- a/lib/core/electrum/application/usecases/add_custom_server_usecase.dart
+++ b/lib/core/electrum/application/usecases/add_custom_server_usecase.dart
@@ -54,9 +54,9 @@ class AddCustomServerUsecase {
         return const Err(ElectrumServerAlreadyExistsFailure());
       }
 
-      // Probe with the user's own validateDomain setting: accepting a
-      // certificate the sync would refuse saves a server that can never be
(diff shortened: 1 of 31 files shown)


BullishNode:
I think we should keep validation offbyu default 

I believe we also want to put electrum validation off by default too 

Since that is one of the most common user support answers

PR #2715, opened 2026-08-23
feat(linux): add webview support and unify cookie handling

basantagoswami:
## Summary

- Replace webview_cookie_manager with the cookie APIs from webview_flutter.
- Add a Linux implementation of webview_flutter so exchange authentication works in the Linux development app.
- Install the required native WebView dependency in the development container.
- Run Linux integration tests in the same pinned Debian Trixie environment as Containerfile.tools.

## Dependency note

webview_flutter_linux is sourced from [basantagoswami/webview_flutter_linux](https://github.com/basantagoswami/webview_flutter_linux), which is under my personal GitHub account, and is pinned to commit 0cae2b66f68ebf426fba81c3fb43347d2c5f0345. It is not published to pub.dev. Linux remains a development-only target; Android and iOS continue to use the official platform implementations.

## Why

The exchange WebView did not have a Linux implementation. Adding one also lets us remove the separate cookie-manager dependency and use one WebView API across platforms. The integration job now uses the same Debian base as the development container so it has the required native packages.

### DIFF
--- a/.github/workflows/analyze_and_test.yml
+++ b/.github/workflows/analyze_and_test.yml
@@ -153,7 +153,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y \
             clang lld cmake ninja-build pkg-config \
-            libgtk-3-dev liblzma-dev libstdc++-12-dev \
+            libgtk-3-dev liblzma-dev libstdc++-12-dev libwpewebkit-1.0-dev \
             libsecret-1-dev libsecret-tools libcurl4-openssl-dev \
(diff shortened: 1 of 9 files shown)


basantagoswami:
Moving this to draft.
CI is failing because Ubuntu 24.04 ARM64 does not package the WPE WebKit development libraries required by `webview_flutter_linux`. The available WPE packages target other distributions/releases and cannot be safely installed on Noble. We'll need some pinned prebuilt ARM64 artifact or an alternative CI environment for the CI to pass for this PR

PR #2717, opened 2026-08-23
feat: add deterministic wallet recovery foundations

BullishNode:
## What this adds

This introduces the shared persistence and deterministic derivation foundations used by wallet recovery and data backup.

- Stores wallet recovery manifests, wallet bindings, Nostr key records, and backup state in schema 15.
- Defines the fixed BIP85 reservations used by wallet backup identities and deterministic wallets.
- Resolves the default wallet and its seed without loading wallet balances.
- Creates deterministic wallets idempotently and records enough metadata to reproduce them safely.
- Keeps private seed material out of logs and backup inventory views.

## Approach

The migration is a single cumulative 14 to 15 upgrade with populated-upgrade and fresh-database coverage. Deterministic wallet creation is orchestrated by a use case over the existing wallet and seed repositories. Fixed reservations are compile-time core constants, and all ambiguous or inconsistent default-wallet states fail explicitly.

## Verification

Migration, reservation, seed-resolution, deterministic-wallet, entropy-listing, and rollback behavior are covered by tests.


BullishNode:
@ethicnology Review fixes are in `87bfed566126166f0494ed83587edc94090526bb`, including the additional RecoverBull namespace reservation. The full local check suite is green; please re-review after the new CI run completes.

BullishNode:
Clarification for review: RecoverBull vault keys are currently derived directly and are never stored in `bip85Derivations`, so they cannot appear in the entropy screen today. The `1608'/0'/…` registry entry prevents future namespace collisions; prefix exclusion is defense-in-depth. The PR body now states this explicitly. No code or commit changed, and the existing re-review request remains active.

PR #2718, opened 2026-08-24
feat: add purpose-scoped Nostr backup identity

BullishNode:
## What this adds

This adds a purpose-scoped Nostr identity used to authenticate wallet backup operations.

- Derives a stable backup public key from the default wallet seed at a fixed BIP85 path.
- Signs only validated 32-byte hashes for the permitted backup purpose.
- Exposes no raw private key, reusable key handle, or generic signing service.
- Uses a shared Nostr bech32 encoder for frozen public-key vectors.

## Approach

Only the facade is registered in the application container. Key resolution and signing remain private implementation details composed inside the facade. Seed lookup uses the balance-free default-wallet path and validates both stored and derived root fingerprints.

## Verification

Tests pin the derivation and signature vectors and cover missing or ambiguous defaults, locked seeds, environment selection, fingerprint mismatches, invalid hashes, sanitized logging, and programmer-error propagation.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -23,6 +23,7 @@ graph TB
     BTC_PRICE[Bitcoin Price]
     NETWORK[Network]
     BIP85[BIP85]
+    NOSTR_IDENTITY[Nostr Identity]
     FEES[Fees]
     WALLETS[Wallets]
     EXCHANGE[Exchange]
--- a/integration_test/bip85_derivation_test.dart
+++ b/integration_test/bip85_derivation_test.dart
@@ -3,13 +3,12 @@ import 'dart:typed_data';
 import 'package:bb_mobile/core/bip85/data/bip85_datasource.dart';
 import 'package:bb_mobile/core/bip85/data/bip85_repository.dart';
 import 'package:bb_mobile/core/bip85/domain/derive_next_bip85_mnemonic_from_default_wallet_usecase.dart';
-import 'package:bb_mobile/core/seed/data/repository/seed_repository.dart';
-import 'package:bb_mobile/core/settings/data/settings_repository.dart';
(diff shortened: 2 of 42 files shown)


BullishNode:
Review follow-up applied in cf73dbc46:

- validate the exact canonical Nostr reservation path, identity, account, owner, and purpose before derivation
- expose typed Nostr reservations from the registry facade instead of using string IDs
- register only `NostrIdentityFacade`; resolver and signing internals remain outside the global container
- normalize and validate wallet, seed, and derived-root fingerprints through the shared primitive
- reuse the derived x-only public point and hoist digest validation
- cover configuration and wallet-lookup failures, the container boundary, and an exact NIP-05 key vector
- retain the network check as a fail-closed repository postcondition and rename its test accordingly

`make checks` passes: analysis, fix/format checks, 1,390 application tests, and all package tests.

BullishNode:
Upstream validation for the current draft head `b34af9b8e5969b51e34cb302045b0605ef737bfb` was started manually because the Analyze and Test workflow auto-triggers only for PRs whose base is `main` or `develop`; this stacked PR targets `simplify/01-storage-foundations`.

Run: https://github.com/SatoshiPortal/bullbitcoin-mobile/actions/runs/32771748981

The same head has completed `make checks` locally: analysis, formatting, 1,393 app tests, migrations/security audits, and all workspace package suites passed. The frozen public-key, npub, and BIP-340 signature vectors are byte-identical.

BullishNode:
@ethicnology Review requested on the current draft head 79a9b8699. Nostr npub and nsec encoding now use one length-checked core codec, while the facade remains purpose-scoped and exposes no secret or generic signer. The frozen public-key, npub, and signature vectors are unchanged. Focused tests and whole-project analysis pass; the stacked CI run is https://github.com/SatoshiPortal/bullbitcoin-mobile/actions/runs/32793853987.

PR #2719, opened 2026-08-24
fix(tor): harden external SOCKS proxy routing

ethicnology:
- Validate a user-managed local SOCKS5 proxy before activation, persist the setting atomically, and refresh its status when the app resumes.
- Resolve the configured external route centrally and fail closed across Bitcoin Electrum, fee requests, wallet warnings, service status, and RecoverBull instead of silently falling back to a direct connection.
- Avoid starting embedded Tor when a verified external route is selected, and reuse the RecoverBull route across retries without taking ownership of the external proxy.
- Present the advanced proxy as provider-agnostic, expose its port and actual SOCKS5 reachability state, provide a retry action, and keep Orbot only as an optional documentation link.
- Preserve destination hostnames through SOCKS5 (`ATYP DOMAINNAME`) and verify that a rejected proxy request never falls back to a direct socket.

## Privacy and behavior

- Embedded Tor remains the default.
- The SOCKS5 handshake proves only that the configured proxy is reachable; it does not claim that the proxy is Tor.
- The app does not inspect installed applications, package identities, or Orbot installation state.
- Bitcoin Electrum and encrypted-backup traffic use the selected external route. Liquid traffic remains direct.
- External-route failures remain visible and fail closed; there is no silent embedded or clearnet fallback.

## Architecture

- Centralize configured external-route resolution behind the Tor use case.
- Restore the Settings and RecoverBull repository domain contracts.
- Keep route ownership explicit so borrowed external proxies are never stopped by the app.

## Review order

1. External-proxy activation and failure routing.
2. Consumer migration to centralized route resolution.
3. Lifecycle, startup, and RecoverBull retry behavior.
4. Repository-boundary refactors.
5. Provider-agnostic UX and SOCKS routing regression tests.

## Verification

- Whole-project Flutter analysis passed with fatal warnings and infos.
- `fvm dart fix --dry-run` reported `Nothing to fix!`.
- Formatting and ARB validation passed.
- The full workspace unit-test suite passed.
- Focused Tor, Electrum, Wallet, Settings, Status, Fees, and RecoverBull tests passed.
- A production-flavor debug APK was built for `android-arm64`, installed on a Pixel 5 running Android 14, and smoke-tested through successful app startup and the Tor settings screen.

## Out of scope

- No database schema change.
- No automatic detection or control of third-party Tor applications.
- No claim that an arbitrary reachable SOCKS5 endpoint provides Tor anonymity.


ethicnology:
@Claude I expect your review

ethicnology:
@Claude please give another round of review

PR #2723, opened 2026-08-24
feat(send): persist resumable bitcoin transactions

ben-kaufman:
## Summary

- Add schema 16 storage for Bitcoin drafts and partially signed transactions.
- Preserve selected policy paths, signer evidence, hash-preimage evidence, fees, and transaction metadata.
- Use revision checks to prevent stale saves or deletes from overwriting newer state.
- Validate stored transactions, inputs, outputs, ownership, and signer results before restoring or merging them.
- Keep valid pending transactions available when another stored record is invalid.

## Database

This adds the pending Bitcoin transaction tables to the existing schema 16 migration. It does not add another schema version.

### DIFF
--- a/lib/core/storage/migrations/schema_15_to_16.dart
+++ b/lib/core/storage/migrations/schema_15_to_16.dart
@@ -135,11 +135,17 @@ class Schema15To16 {
         );
       }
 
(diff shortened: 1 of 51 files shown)


ben-kaufman:

PR #2724, opened 2026-08-24
feat(send): add resumable descriptor signing

ben-kaufman:
## Summary

- Add a dedicated signing screen for multisig and Miniscript transactions.
- Keep spending-option selection separate from signer selection.
- Show every eligible signer and its current PSBT-derived status.
- Allow signatures to be added in any order.
- Show how many more signers are needed for threshold policies.
- Stop requesting signatures once the selected policy can be satisfied.
- Include Bull's on-device signer as an explicit signing option.
- Support sharing a PSBT and importing the signed result from an external signer.
- Let the user finish signing later and resume from the wallet.
- Show pending drafts and signing sessions with the normal wallet activity.
- Preserve the existing single-signature mobile and hardware-wallet flows.

Collected signatures are retained when the underlying transaction remains unchanged. If editing requires rebuilding the transaction, the user is warned that signing must restart.


ben-kaufman:

PR #2728, opened 2026-08-24
ci(ios): publish versioned store and simulator artifacts

ethicnology:


ethicnology:
Findings 1 and 2 are resolved by the semantics of the exact pinned upload-artifact SHA. With `archive: false`, the action uploads the single file as-is, uses its filename as the artifact name, and explicitly ignores the `name` input:

https://raw.githubusercontent.com/actions/upload-artifact/043fb46d1a93c77aae656e7c1c64a875d1fc6a0a/action.yml

The simulator steps are intentionally on the critical path because the simulator package is now a required release deliverable. Running them before the App Store upload also preserves the simulator artifact when Apple rejects an already-used build number.

PR #2731, opened 2026-08-24
feat: add wallet recovery manifest and Nostr key inventory

BullishNode:
## What this adds

This adds a recovery manifest for seed-derived identities and deterministic wallet bindings.

- Records each recoverable Nostr key with its derivation path, parent fingerprint, description, and timestamps.
- Records deterministic wallet bindings before remote registration or publication.
- Supports named user Nostr identities and the reserved system identities.
- Provides create and edit screens using the shared input controls.
- Allows warning-gated, point-of-use nsec reveal for user and system identities.
- Produces and parses one canonical JSON representation for backup and recovery.

## Approach

The repository is the single write and change-notification boundary. Local writes notify backup consumers; recovered inventory does not immediately republish itself. Parsing, validation, canonical ordering, and merge behavior use the same entity rules so malformed data fails before persistence.

Private keys are re-derived only when the user confirms a reveal. The reveal verifies the active seed fingerprint and stored public key, enables screen-capture protection before derivation, and clears the secret when copied, closed, or backgrounded.

## Verification

Tests pin canonical manifest bytes and frozen Nostr derivation vectors, cover local versus recovery writes, validate additive import behavior, exercise user-key allocation alongside non-BIP85 wallet entries, and verify protected secret reveal for both user and system identities.

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -24,6 +24,7 @@ graph TB
     NETWORK[Network]
     BIP85[BIP85]
     NOSTR_IDENTITY[Nostr Identity]
+    KEYCHAIN_MANIFEST[Keychain Manifest<br/>---<br/>Wallet inventory, Nostr keys,<br/>passphrase wallet records]
     FEES[Fees]
     WALLETS[Wallets]
     EXCHANGE[Exchange]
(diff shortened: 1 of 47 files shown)


BullishNode:
@ethicnology Review requested on the current draft head 6b37e6810. Local manifest commits now publish from the repository transaction boundary regardless of whether they originate from the facade or Settings use cases; recovery writes remain silent. Key creation derives only the public key, nsec encoding occurs only during reveal, Nostr encoding and validation have one implementation, and unused pre-release compatibility paths are removed. Frozen identity and wire-format vectors are unchanged. Full make checks pass with 1,433 app tests and all workspace package suites; stacked CI is https://github.com/SatoshiPortal/bullbitcoin-mobile/actions/runs/32793856516.

PR #2732, opened 2026-08-25
feat(wallet_backup): add standalone remote backup protocol

BullishNode:
## Summary

Adds the authenticated remote protocol used to store encrypted wallet metadata backups independently of payment services.

The client implements the BULL-metadata-backup v0.1.0 contract pinned at `152648008882c73f7a81fd590ef7819f3b1cc64d`:

- purpose-scoped Nostr authentication for fetch, store, and delete
- generation and ETag compare-and-swap semantics
- strict response, receipt, hash, byte-count, and tombstone validation
- canonical base64 ciphertext capped at 1 MiB
- bounded HTTP timeouts and local `Retry-After` enforcement
- HTTPS server origins, with loopback HTTP allowed only in debug builds

The default service is `https://backup.bull-wallet.com`. Builds can override it with `BULL_METADATA_BACKUP_BASE_URL`; later settings code can provide a user-selected origin through the same provider boundary.

## Protocol verification

The server release fixture is copied unchanged into the client tests. The tests pin every signing-message and ETag vector from v0.1.0, the three routes and request fields, response integrity checks, receipt validation, server error mapping, origin validation, and the 1 MiB limit.

## Validation

- `make analyze bull-ui-check fix-check format-check`
- `make unit-test`
- focused wallet-backup protocol tests

This PR is intentionally draft as part of an open stacked series.


BullishNode:
Replaced by #2757.

PR #2736, opened 2026-08-25
fix(wallet): correct RBF fees and improve settings connectivity

ethicnology:
- Fix RBF fee-rate prefill so the app-proposed value clears BDK's replacement threshold after weight/vsize conversion and the two-decimal input round-trip.
- Reorganize Settings into clearer App, Wallet, Bitcoin, Exchange, Map, Terms of Service, Service Status, and Logs sections.
- Open wallet details directly, preserve deletion for non-default wallets, and move backup and RecoverBull actions into their contextual settings.
- Publish the shared Tor configuration through RecoverBull and contextual Electrum entry points.
- Make custom onion Electrum servers use the selected Tor route reliably, with route-aware probes and onion-appropriate timeouts.
- Remove Bitcoin's redundant preliminary socket probe while preserving Liquid's socket and protocol checks.
- Publish Electrum server and Service Status results progressively instead of waiting for the slowest probe.
- Prevent stale asynchronous results from overwriting newer refreshes or Electrum mutations.

## RBF replacement fee

The previous prefill derived the original rate from `feeSat / vsize`, while BDK validates replacement fees against the transaction's actual weight. Because `vsize` is `ceil(weight / 4)`, the app could propose a rate slightly below BDK's threshold and then reject its own suggestion.

The original regression fixture demonstrated the mismatch:

- Previous prefill: 607 sat/kwu
- Minimum safe rate: 609 sat/kwu

The replacement rate now uses integer arithmetic and a conservative weight bound. It is rounded upward to remain valid after being displayed and reparsed as a two-decimal sat/vByte value.

BDK's exact required rate is logged at the failure boundary without exposing it to users.

## Settings hierarchy

The Settings root order is now:

1. App
2. Wallet
3. Bitcoin
4. Exchange
5. Map
6. Terms of Service
7. Service Status
8. Logs

Wallet actions are ordered as:

1. Start Backup
2. Test Backup, when a backup exists
3. RecoverBull
4. Import wallet
5. Labels
6. Transaction History

Additional navigation changes:

- Rename Wallet Backup to Wallet.
- Remove the redundant wallet-list screen.
- Open wallet details directly.
- Retain deletion in the top bar for non-default wallets.
- Place Addresses at the bottom of wallet details.
- Move Testnet Mode below Dev Mode and group developer controls at the bottom of App settings.
- Shorten Bitcoin list labels to Auto Transfer, Electrum Server, and Mempool Server without changing their screen titles.
- Move View Vault Key into RecoverBull advanced settings while preserving its existing warning and authentication flow.

## Tor ownership

Tor configuration remains owned by the Tor feature and is published through its public surface.

- RecoverBull advanced settings provide access to the shared Tor configuration.
- Adding a custom Bitcoin onion server exposes the same Tor configuration contextually.
- Onion inputs automatically disable SSL and show the Tor-specific guidance.
- Configured external proxies remain authoritative and fail closed when unavailable.
- Embedded routes are owned and closed by the app; external proxy routes are borrowed and never stopped by it.
- The standalone Tor route remains available for contextual error and retry flows.
- `FEATURES.md` records the Electrum Settings → Tor dependency.

## Electrum connectivity

Custom server validation now follows the same settings and route that production synchronization will use.

- Bitcoin validation uses the BDK Electrum client, including protocol and chain-data checks.
- Bitcoin no longer performs a redundant socket probe before the BDK probe.
- Liquid retains its preliminary socket check followed by its protocol check.
- The user's domain-validation, timeout, and retry settings are forwarded to the probe.
- Onion probes use a minimum 30-second timeout.
- Custom onion servers are probed through the resolved embedded or external Tor route.
- External Tor failures return a specific configuration failure instead of silently falling back to clearnet.
- Tor routes are closed after server validation.
- Duplicate, unreachable, malformed, and protocol-incompatible servers are rejected before persistence.

## Progressive status checks

Electrum Settings and Service Status now display results as each probe completes.

### Electrum Settings

- Servers appear immediately with an `Unknown` status.
- Each server independently transitions to `Online` or `Offline`.
- `Unknown` is no longer rendered as `Offline`.
- Generation guards prevent obsolete loads from overwriting newer loads.
- Adding, deleting, reordering, or changing advanced options invalidates in-flight load callbacks.

### Service Status

- Each completed service check is published immediately.
- A slow Bitcoin or Tor probe no longer blocks every other result.
- An exception from one probe marks only that service as `Unknown`.
- Successfully completed service results remain visible.
- Programmer `Error`s are allowed to propagate instead of being converted into connectivity failures.
- Obsolete refresh callbacks cannot overwrite a newer refresh.
- Payjoin remains `Disabled`, rather than `Offline`, when the feature is disabled.
- `Last checked` is finalized only after the complete refresh.

## Security and privacy

This PR moves the View Vault Key entry but does not change the reveal or authentication flow. The existing warning remains in place, and the entry is only displayed when the current wallet has an encrypted backup.

No mnemonic, seed, private key, vault key, PIN, or other secret material is added to UI state or logs.

External Tor failures remain visible and fail closed. No onion Electrum request silently falls back to a direct clearnet connection.

## Review order

1. RBF calculation and regression tests.
2. Settings and wallet navigation reorganization.
3. RecoverBull View Vault Key relocation.
4. Shared and contextual Tor settings.
5. Onion Electrum routing, timeout, retry, and validation behavior.
6. Removal of the redundant Bitcoin socket probe.
7. Progressive Electrum server statuses and stale-load protection.
8. Progressive Service Status updates and exception isolation.


ethicnology:
@claude review

ethicnology:
@Claude re-audit

PR #2741, opened 2026-08-26
fix(send): report insufficient funds when the fee doesn't fit

### LINKED ISSUE #2739: Insufficient-funds error shows as "Build Failed" when sending near-max amount
## Description
Sending an amount just below the full spendable balance fails with a generic "Build Failed" error. The actual cause is that the remaining sats can't cover the network fee, but the message gives the user no way to work that out - it reads like an app crash rather than a funding problem.

The MAX button works correctly, since it accounts for the fee automatically.

### Reproduce 
1. Open a Liquid wallet with a balance
2. Go to Send and paste any Liquid address
3. Manually enter an amount 2 sats below the spendable balance (don't use MAX)
4. Tap Confirm
5. See "Build Failed"

### Expected behavior
A clear message explaining the balance can't cover the amount plus the network fee — e.g. "Insufficient funds to cover the network fee".

### Demo

Image

wired-pasteque:
Fixes: #2739

Simulator Screenshot - iPhone 17e - 2026-08-26 at 11 32 02


wired-pasteque:
@claude, review the last changes

wired-pasteque:
@claude, review the last fixes

PR #2754, opened 2026-08-27
feat: show wallet backup status and protected data

BullishNode:
## Summary

Redesigns Backup Settings around the user’s recovery status and adds read-only views of the wallet data covered by backup.

The screen now presents three clear areas:

- Bull backup status and controls;
- the protected wallet data inventory;
- physical and encrypted recovery methods, including their latest test dates.

The existing seed backup, RecoverBull, label export, transaction export, remote-backup deletion, and recovery-retry actions remain available.

## Wallet recovery manifest

The new wallet recovery manifest explains how every wallet in the local catalogue can be recovered without displaying descriptors, fingerprints, xpubs, keys, or backup payloads.

For each wallet it shows the applicable recovery facts:

- default wallets recover from the main seed;
- BIP85 product wallets recover deterministically from the main seed;
- imported mnemonic wallets require their original mnemonic;
- watch-only wallets use their included public wallet definition;
- external-signer wallets use their public definition and named signing device;
- seed-passphrase use and scan birthday are shown when known;
- Bitcoin public definitions are marked included;
- Liquid public definitions are explicitly marked as not included yet.

The inventory reads wallet metadata directly and does not start BDK/LWK wallet engines or fetch balances.

## Wallet metadata view

The metadata view reports record counts from the same local contributors used to compose a backup:

- BIP329 labels;
- wallet preferences;
- frozen-coin state.

It also states that balances and transaction history are rebuilt from the network and are not stored in the backup. The existing Nostr identities screen is linked from the same protected-data section.

## Privacy and behavior

The backup feature publishes a sanitized contents model containing only recovery classification, network, optional display label, signer-device type, birthday, passphrase-use state, and metadata counts. Public wallet identifiers and payload bytes never cross into the settings UI.

Opening these views performs no remote request and does not change the backup format, database schema, encryption, recovery coordinator, or publication behavior. A contents-read failure is isolated to this section and can be retried without disabling the rest of Backup Settings.

## Testing

- whole-project analyzer with fatal warnings and infos
- complete root unit suite: 1,616 tests
- bull_payjoin, bull_ui, bull_ui_catalogue, and primitives package suites
- all five wallet provenance categories and Liquid exclusion status
- metadata counts from the production backup contributors
- sanitized recovery and metadata screen widget coverage
- local contents failure and retry behavior

This pull request remains a draft while the feature stack is under review.


BullishNode:
Replaced by #2758.

PR #2745, opened 2026-08-27
feat(coins): sweep selected bitcoin utxos

ben-kaufman:
Closes #329.

## Summary

Adds a single-destination **Sweep** action to the Bitcoin UTXO management flow.

Users can select one or more spendable coins, enter a Bitcoin destination, choose a fee, and send the entire selected value minus the network fee.

## Behavior

- Shows **Sweep** when all selected coins are spendable.
- Supports sweeping one or multiple selected UTXOs.
- Sends the selected balance minus the network fee to one destination.
- Recalculates the final amount when the fee or RBF setting changes.
- Uses only explicitly selected inputs.
- Stops the sweep if a selected input becomes spent, frozen, unavailable, or Payjoin-reserved.
- Disables Payjoin for selected-input sweeps.
- Verifies externally signed transactions against every confirmed input and output.

## Implementation

- Adds exact-selected-input mode through the Bitcoin send port, repository, and BDK datasource.
- Uses BDK manually selected inputs without automatic input supplementation.
- Returns selected outpoints to the Wallet flow, which opens Send through its public facade.
- Validates the sweep destination and confirms the selected coins remain available before preparing the transaction.
- Migrates Coins errors to typed failures.
- Moves signed-transaction verification into the shared wallet layer.

PR #2756, opened 2026-08-28
feat(send): support multiple bitcoin recipients

ben-kaufman:
Depends on #2745.

## What changed

- Adds and removes recipients from the Bitcoin on-chain address step.
- Supports multiple fixed outputs with normal wallet change.
- Allows one recipient to receive the remaining balance after network fees.
- Recalculates the remaining amount when fees, wallets, or selected inputs change.
- Keeps the existing single-recipient amount and MAX experience unchanged.
- Shows every recipient, final amount, and network fee on confirmation.
- Keeps selected-UTXO sweeps limited to the selected inputs.
- Disables Payjoin for multi-recipient and manually selected-input transactions.


BullishNode:
Concept Ack. Would you mind uploading a video recording of the whole flow and cases?

PR #2766, opened 2026-08-28
fix(send): stop coin control from spending unselected coins

wired-pasteque:
Hand-picking coins that fell short of amount plus fee let BDK silently top the transaction up from a coin the user had excluded; the selection is now exclusive and the `Done` button waits for a real fee.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to Bull Bitcoin Mobile will be documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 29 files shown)


wired-pasteque:
@claude, check this PR

ethicnology:
@Claude should we generalize bdk manuallySelected() explicitly rather than when no drain ?

We really want to make sure the user spend only the selected coins, if the transaction can't be build because of the selection, we should fail explicitly

ethicnology:
@claude last chance to review / audit

PR #2726, opened 2026-08-24
feat(wallet): import and sign taproot descriptors

ben-kaufman:
## Summary

- Import and analyze Taproot descriptor wallets.
- Support Taproot key-path signing.
- Support Tapscript policies using `multi_a`, thresholds, timelocks and hashlocks.
- Track script-path signatures using their exact Tapleaf hashes.
- Build PSBTs containing only the Tapleaf information needed for the selected spending option.
- Verify Taproot signatures before accepting them.
- Support Taproot wallet-policy registration, address verification and signing on compatible hardware wallets.
- Keep unsupported device actions hidden or clearly unavailable.

## Descriptor examples

Key-path wallet:

```text
tr([FINGERPRINT/86'/0'/0']XPUB/<0;1>/*)
```

Script-path multisig using the standard BIP341 NUMS internal key:

```text
tr(50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0,multi_a(2,KEY_A,KEY_B,KEY_C))
```

The recognized BIP341 NUMS internal key remains part of the descriptor and device wallet policy but is not presented as a signer.

## Scope

- Taproot key paths and supported Tapscript policies are included.
- MuSig2 is not supported.
- Device actions are offered only for policy and transport combinations supported by that device.

PR #2774, opened 2026-08-31
feat(wallet): reserve BIP48 signer accounts

ben-kaufman:
## Summary

- persist exact BIP48 account reservations per Bull seed fingerprint and Bitcoin coin type
- suggest the lowest unused account for Bull signer exports while preserving explicit account selection
- reserve manual exports only after explicit confirmation and reserve wallet accounts after verified creation or import
- cryptographically verify Bull-owned account xpubs before recording descriptor usage
- keep reservations after wallet deletion so exposed account keys are never silently reused

## Validation

- root Flutter test suite
- focused wallet, signing-key export, and watch-only import tests

## Stack

Draft stacked on #2726.

### DIFF
--- a/lib/core/utils/bip48_derivation.dart
+++ b/lib/core/utils/bip48_derivation.dart
@@ -1,0 +1,33 @@
+abstract final class Bip48Derivation {
+  static const maxAccount = 0x7fffffff;
+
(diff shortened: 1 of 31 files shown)


ben-kaufman:

PR #2775, opened 2026-08-31
feat(backup): verify a specific seed backup

ben-kaufman:
## Summary

- verify physical mnemonic backup status for an exact Bull seed fingerprint instead of the default wallet
- persist successful mnemonic verification against every local wallet using that seed
- expose fingerprint-scoped backup verification through the test-wallet-backup facade
- expose tested RecoverBull backup status for the same seed and return completion through its public flow
- preserve existing backup and recovery screens while allowing BullVault to reuse them safely

## Validation

- root Flutter test suite
- focused physical-backup and RecoverBull tests

## Stack

Draft stacked on #2774.


ben-kaufman:

PR #2776, opened 2026-08-31
feat(bullvault): add vault onboarding

ben-kaufman:
## Summary

- add BullVault creation with a phone key, one or two cold hardware keys, and optional inheritance
- build canonical Taproot Miniscript policies with immediate multisignature spending and adjustable absolute recovery schedules
- allocate compact receive/change branch pairs per repeated account-xpub occurrence without reusing exposed derivations
- acquire signer keys through the existing Ledger, BitBox, and QR-device flows and require successful policy registration during setup
- cross-check phone time against Bitcoin median time before committing recovery timestamps
- persist resumable setup state and activate the wallet only after its recovery package is saved
- restore from a portable recovery package or a recognized BullVault descriptor while identifying the Bull mobile key cryptographically
- reuse verified physical mnemonic and RecoverBull backup status while always requiring the separate BullVault recovery package
- add BullVault creation and restoration under Settings and keep it out of the generic wallet-import menu

## Validation

- whole-project analysis, formatting, fix, and boundary checks
- root Flutter test suite
- policy golden vectors for every protection and inheritance profile
- focused creation, restoration, persistence, onboarding, hardware, and UI tests
- release build installed and launched on a physical iPhone

## Stack

Draft stacked on #2775.

PR #2777, opened 2026-08-31
feat(bullvault): renew vault policies

ben-kaufman:
## Summary

- renew an active BullVault with fresh absolute recovery dates while retaining its verified signer configuration
- allocate only the next unused compact derivation pairs for each account xpub
- require the replacement recovery package and hardware policy registrations before activation
- activate the replacement atomically while preserving lineage and monitoring the retired vault
- allow prepared renewals to be cancelled without reusing exposed generations or changing the active vault
- surface cancelled replacements only if they later receive funds
- resume interrupted renewal setup and reconcile wallet visibility after restart
- add renewal status, recovery dates, setup completion, and cancellation to BullVault wallet settings
- protect active and monitored BullVault wallets from unsafe deletion
- route spending to the active generation and remind users to verify BullVault receive addresses on their hardware wallet

## Validation

- root Flutter test suite
- focused renewal, activation, cancellation, migration, wallet-visibility, Send, Receive, and UI tests

## Stack

Draft stacked on #2776.


ben-kaufman:

PR #2782, opened 2026-09-01
fix(wallet): handle unavailable coins before broadcast

ethicnology:
`ValidateBitcoinSelectionUsecase` rethrows `NoSpendableUtxoException` when a selected coin is no longer available, but three send and swap broadcast/rebuild paths still handled `InsufficientFundsException` instead. The transactions remained fail-closed, but users received generic errors and signed transaction or fee-preview state was not consistently invalidated.

## Changes

- Handle unavailable selected coins before Send broadcasts, invalidate the staged transaction, and surface `SendSelectedCoinsUnavailableFailure`.
- Map unavailable selected coins to `selectedCoinsUnavailableCode` in the cached Swap rebuild path.
- Clear Swap fee previews and surface the unavailable-selection error when confirmation validation fails.
- Add one behavioral regression test for each path, including assertions that no broadcast occurs.

## Non-goals

- No change to coin-selection rules or normal insufficient-funds handling.
- No change to transaction construction or broadcast behavior after successful validation.

## Risk and security

The affected paths remain fail-closed: validation completes before any broadcast call, and the regression tests verify that unavailable selected coins cannot reach broadcast. This change corrects error mapping and stale-state cleanup only. It does not touch key material, signing logic, wallet secrets, or privacy-sensitive data.

### DIFF
--- a/lib/features/send/presentation/bloc/send_cubit.dart
+++ b/lib/features/send/presentation/bloc/send_cubit.dart
@@ -2104,6 +2104,8 @@ class SendCubit extends Cubit<SendState>
         selectedInputs: state.selectedUtxos,
       );
       return true;
+    } on NoSpendableUtxoException {
+      _invalidateSignedTransaction();
     } on InsufficientFundsException {
       _invalidateSignedTransaction();
     } on ValidateBitcoinSelectionException {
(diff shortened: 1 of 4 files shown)


ethicnology:
@Claude can you add the commits based on your review https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/2782#issuecomment-5486590918 and make another round of audit?

PR #2805, opened 2026-09-08
feat(swaps): self-contained swap engine with full lifecycle drivers

i5hi:
Extracts the Boltz swap engine into a self-contained melos package (`packages/swaps`) and restores the full swap lifecycle that 6.13 lost, on top of the v6.13.1 base plus the stuck-swap rescue fixes proven in v6.13.2–v6.13.4. Version 6.14.0+220.

## Architecture (deliberately simple)

- `SwapRepository` — one abstract interface: reads, creates, and high-level resolve operations (`claim` / `refund` / `coopSign` handle addresses, fees, cooperative-vs-script path and electrum fallback internally). `BoltzSwapRepository` is the trustless implementation; a trusted (Bull exchange) implementation fits the same surface later.
- `SwapWatcher` — the only driver in the system: verify recorded completions, sweep every ongoing swap by status, subscribe + reconcile against Boltz, heartbeat, retry backoff. Blocs never move a swap.
- Thin usecases for blocs. No datasource layer, no ports file; the app injects the storage backends and three callbacks (new address, fastest fee, electrum runner) at construction.

## What this fixes

- Restores claim drivers (chain + reverse) and coop-sign — every previously orphaned swap state has a driver again.
- Swap restore/rescue works again: the repository self-derives the swap master key. All 6.13 builds shipped with the key never bound, so the restore screen threw unconditionally.
- `transaction.claim.pending` is no longer mis-stored as `claimable` for submarine swaps (a state the type can never legally hold).
- Startup no longer blocks on wallet syncs: completion verification is cache-first and only syncs on suspicion.
- Refund fees floored at relay minimum and capped at half the swap amount; submarine refunds get the electrum fallback too; the rescue rebuild uses the active electrum servers instead of hardcoded ones.
- User-configurable swap server: Settings → Bitcoin → Swap server (applies on next launch).
- Startup census trimmed: full txid detail only for unresolved swaps.
- Removes the unregistered `BoltzAutoswapProvider` dead code (it created chain swaps nothing could claim).

## Touches key material

The engine reads the default wallet seed only at the point of deriving the swap master key (unchanged semantics, moved behind a callback) and stores bull_sdk swap blobs in secure storage behind a dumb `SecretStore` contract. Storage key prefixes and the sqlite swaps table are unchanged — existing installs keep their swaps and secrets with no migration.

## Testing

- `make checks` green: whole-project analyze with zero issues, format, fix, and the full unit suite including the security-audit tests.
- 12 new engine tests in `packages/swaps/test`: refund settle, change-spend rejection (fail-closed outspend verification), non-final timelock skip, idempotency on recorded txids, cache-first verify, retract-and-reopen of unproven completions, watcher drive-per-status, backoff behaviour.
- The refund path is the one proven in production by the D5gdAL9UI29W recovery: a 24-sat script-path refund broadcast via electrum with the Boltz API unreachable.


i5hi:
@claude please review

PR #2808, opened 2026-09-09
perf(core): reduce wizard completion latency

ethicnology-agent:
## Linked issue

None — measured startup performance correction requesting the `no-issue-needed` exemption.

## Problem

Completing the first-run wizard blocked the UI while startup diagnostics and Android secure-storage initialization ran. On a Samsung S10e profile build, the median delay from the final tap to the wallet onboarding screen was 1.22 s on `develop`, with a 2.56 s cold run.

## Changes

- Create the log file on the startup path but defer runtime diagnostic collection until after the first application frame.
- Prewarm FSS10 during the wizard only when no storage-library flag, SQLite database, or legacy Hive box identifies a supported prior install.
- Keep `migrateOnAlgorithmChange` disabled and preserve the authoritative FSS10/FSS9 fallback for upgrades.
- Add regression coverage for deferred diagnostics, fresh-install prewarming, prior-install gates, and fail-closed behavior.

## Non-goals

- No change to wallet creation, synchronization, database schema, dependencies, or generated files.
- No change to the supported FSS9/FSS10 upgrade selection behavior.

## Risk and security

This PR touches secure-storage initialization used by seed storage and requires security/migration review. It does not log, return, or cache mnemonic or seed values. The optimization is skipped for every supported prior-install marker, failures skip the optimization, Android backup is disabled, and the authoritative startup probe remains unchanged. A partial out-of-model OEM transfer that restores secure preferences while omitting every supported app marker remains a residual platform risk.

## Performance

Samsung S10e, production profile arm64 APK, fresh app data for each run, measured from the visible final-button touch to the wallet onboarding scene:

| Build | Runs | Median |
| --- | --- | --- |
| 6.13.1 | 1.24 s, 1.39 s, 1.56 s | 1.39 s |
| `develop` at `eb7c4fa89` | 2.56 s, 0.90 s, 1.22 s | 1.22 s |
| This PR | 0.170 s, 0.235 s, 0.266 s | 0.235 s |

The median improves by about 81% from `develop` and 83% from 6.13.1. All final runs reached the expected Create New Wallet / Recover Wallet screen, and Android logs showed the cold FSS initialization completed before the first wizard interaction.

## Verification

- `make checks` — passed with Flutter 3.44.9 / Dart 3.12.2 via FVM 4.1.2.
- CI-equivalent Linux integration command under Xvfb, D-Bus, and GNOME Keyring — 21 passed, 10 expected skips; funded testnet fixtures were intentionally absent for this `develop` PR.
- Production profile arm64 APK — built, installed, and measured on a Samsung S10e.
- Independent secure-storage review — no confirmed regression for supported install and upgrade states.

## Review order

Single commit: `af1a2e5d4`.


ethicnology:
@Claude any opinion on that PR?

ethicnology-agent:
Thanks for the detailed review. I checked both native-side questions against the exact locked fork, `SatoshiPortal/flutter_secure_storage@0a0869d0ff0ba2b3cb3576c4a2350428dec30f77`.

For the possible prewarm/probe overlap: the Dart futures can overlap, but the Android calls cannot race under Bull's current options. `FlutterSecureStoragePlugin.onMethodCall` posts every call to one `HandlerThread` (`FlutterSecureStoragePlugin.java:64-68`). Bull uses the standard `AndroidOptions` constructor with the default RSA-OAEP key cipher and `enforceBiometrics:false`; this follows the synchronous branch in `FlutterSecureStorage.initializeStorageCipher`, assigning `storageCipher` and invoking the callback before the current `MethodRunner` returns (`FlutterSecureStorage.java:264-291`). A queued `readAll` therefore cannot overtake the prewarm `containsKey`. The asynchronous window you identified is real for the biometric AES path (`FlutterSecureStorage.java:294-312`), but that path is not configured here. Joining the Dart future would serialize additional fresh-install work without closing a race in the current configuration, so I do not plan to add that startup wait.

For the failed FSS10 initialization invariant: the fork does fail before writing custom-cipher state. When ESP data exists with `migrateOnAlgorithmChange:false` and `encryptedSharedPreferences:false`, initialization invokes the error callback and returns at `FlutterSecureStorage.java:184-199`. It never reaches `initializeStorageCipher` at line 260 or constructs `StorageCipherFactory` at line 266. The algorithm-marker write is in the factory constructor (`StorageCipherFactory.java:60-67`), so no custom cipher/current marker is persisted on that failure path. ESP manager initialization may open the existing Tink/Keystore material, but it cannot commit the custom marker that the fallback safety argument depends on.

I also verified the smaller observations. The prewarm catch is intentionally silent and fail-closed; the authoritative startup probe repeats initialization and reports any consequential failure. Diagnostic context is intentionally best-effort until the first frame, including the error-screen path, in exchange for removing it from the critical transition. `defaultTargetPlatform` is used specifically so the prewarm gates remain testable through `debugDefaultTargetPlatformOverride`; production Android still agrees with `Platform.isAndroid`.

Based on the exact native call order, I consider the two merge questions resolved without an additional code change.

Issue #2486, opened 2026-07-21
Add Autobuy button on the wallet/exchange screen


thibistaken:
Add an Autobuy (recurring buy / DCA) entry-point button on the wallet / exchange screen so users can start an Autobuy directly from there, matching the web exchange. Today the recurring-buy flow is reachable but there's no prominent button on the wallet/exchange view — this is a feature-parity gap with the web app.

## User story
As a user on the wallet / exchange screen, when I want to set up a recurring buy, I see an Autobuy button right there and can start the flow without hunting for it.

## Acceptance criteria
- [ ] Autobuy button is present on the wallet / exchange screen
- [ ] Tapping it opens the existing recurring-buy (DCA) flow
- [ ] Placement and labeling are consistent with the web exchange and the app's other buy entry points
- [ ] Button respects existing KYC / funding gating (no dead-ends)

Backend API: Not required (reuses the existing recurring-buy flow).

thibistaken:
### Test cases

**Happy path**
- On the Exchange home screen, verify the AutoBuy card is shown below the recurring-buy card.
- With AutoBuy inactive, tap **Activate AutoBuy** → the AutoBuy introduction opens.
- Continue to destination-wallet setup, configure a valid Bitcoin, Lightning, or Liquid wallet, and proceed to confirmation.
- Verify the confirmation screen shows the configured wallet details, then enable AutoBuy.
- Return to the Exchange home screen → the AutoBuy card shows the active state.
- Fund the Bull Bitcoin account → the deposited amount is converted to bitcoin at the current exchange rate and sent to the configured wallet.

**Existing destination wallet**
- With a default Bitcoin, Lightning, or Liquid wallet already configured, open AutoBuy → the existing details are shown and can be used without re-entering them.

**Validation and error handling**
- With no destination wallet configured, verify AutoBuy cannot be enabled until at least one valid wallet is provided.
- Enter invalid wallet details → validation prevents them from being saved or used.
- Simulate a failure while enabling AutoBuy → an error is shown and the UI does not report AutoBuy as active.
- For an ineligible user, verify the applicable restriction is shown and AutoBuy cannot be activated.

**Disable flow**
- With AutoBuy active, choose to disable it → a confirmation prompt is shown.
- Cancel the prompt → AutoBuy remains active.
- Confirm the prompt → AutoBuy is disabled and the Exchange home card returns to the inactive state.

basantagoswami:
[...]

PR #2835, opened 2026-09-22
refactor(default_wallets): expose reusable feature boundary

basantagoswami:
## Summary

- extract default-wallet presentation, UI, routing, and dependency wiring into a dedicated feature
- expose a scoped public facade that can be reused by the upcoming AutoBuy flow
- preserve the existing Exchange Settings flow and current core exchange storage behavior

This is part 1 of the implementation for #2486. It establishes the feature boundary needed for AutoBuy without adding AutoBuy behavior itself. Part 2 will be raised as a stacked PR targeting this branch.


basantagoswami:
@claude I'll defer both findings for now and address them later if the stacked AutoBuy implementation requires the embedded self-sizing mode or exposes the close-during-load race. The route name change is intentional and will remain.

basantagoswami:
Adding the `no-issue-needed` label so that merging this alone doesn't close the issue

